PR #9, opened 2015-04-07
adjustments in import / export packages in pom.xml for osgi support

ingoschaal:
Signed-off-by: ingo schaal <ingo.schaal@bosch-si.com>
I fixed a few import / export packages instructions in the pom.xml of
leshan-core
leshan-server-cf
leshan-server-core
in order to use this artefacts as OSGi-bundles in a OSGi container.

### DIFF
--- a/leshan-core/pom.xml
+++ b/leshan-core/pom.xml
@@ -67,11 +67,13 @@ Contributors:
                 <configuration>
                     <instructions>
                         <Export-Package>
-                            org.eclipse.leshan,        
+                            org.eclipse.leshan,
+                            org.eclipse.leshan.core.model,
+                            org.eclipse.leshan.core.model.json,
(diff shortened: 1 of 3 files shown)


ingoschaal:
Hello Kai,
yes, no problem.
In general in this commit, it's about that each bundle has to export the packages which are used in other bundles. And bundles which are using packages from other bundles has to import these.
This is done using the import / export instructions of the maven bundle plugin in pom.xml

jvermillard:
@sophokles73 when merging, please try to avoid merge commits by merging with --ff-only option in the command line and maybe "git rebase master" on the PR branch

sophokles73:
@jvermillard Thanks for pointing this out, Julien. I'll keep that in mind on the next commit ...

PR #12, opened 2015-04-23
Allow for interaction between a program using the leshan-client and the ...

jschloman:
...object instances it creates.  This is demonstrated in the example by allowing the a user to set the error code via the console.

Signed-off-by: John Schloman <jschloman@zebra.com>

### DIFF
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/ObjectsInitializer.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/ObjectsInitializer.java
@@ -20,14 +20,17 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.eclipse.leshan.core.model.LwM2mModel;
 import org.eclipse.leshan.core.model.ObjectLoader;
 import org.eclipse.leshan.core.model.ObjectModel;
+import org.eclipse.leshan.util.Validate;
 
 public class ObjectsInitializer {
 
     protected Map<Integer, Class<? extends LwM2mInstanceEnabler>> classes = new HashMap<Integer, Class<? extends LwM2mInstanceEnabler>>();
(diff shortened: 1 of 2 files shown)


jschloman:
Simon,

I've finally pushed through the simple as possible approach.  I put in some runtime exceptions so that the user doesn't inadvertently setClassForObject() and setInstanceForObject() on the same object.  I then added a simplistic Location object instance in the example to demonstrate how to use this.  Hopefully this should give the users the flexibility they need.

-JF

sbernard31:
John,
 I'm in vacation for 2 weeks, I will look at this when I will be back.

sbernard31:
It seems ok for the `setInstanceForObject` :) !
But now we can set an instance at initialization, we don't need the listener in the `ObjectsInitializer` anymore ?

(Not linked to this issue but I see my code is bugged  in `createNodeEnabler` ...  we should not  always create an `ObjectEnabler` with the `SimpleInstanceEnabler.class` as parameter, we should use the class in the `classes` map or the class of the instance we  just created.)

jschloman:
I'm wondering if it would still be useful to keep the listener:  in the case of a resource that allows multiple instances, it might be preferable for the device using the library to keep track of all instances created by the server?

Also there is the issue on the leshan-dev mailing list about parameters not being passed for ExecuteRequests.  I responded with where I think the issue is in the code... maybe it would be worthwhile to sneak that into this PR as well?

(Sorry for the late reply.  I'm on vacation right now and have intermittent internet access).

-J

sbernard31:
I think we probably need listener for instance creation but this should be managed at `ObjectEnabler` level.
(not in `ObjectsInitializer`, even if we can imagine that `ObjectsInitializer` could add listener on `ObjectEnabler`)

For the issue in the mailing list, you get right I fixed it and add the UI in standalone to be able to send execute request with parameters. I preferred having this changes in a separated commit.

jschloman:
OK so I've removed all of the listener work and now just allow for adding of the instance to use.

sbernard31:
I integrated this in master (I remove some unused import)
I fix the bug I talked about previously in another commit.
This should be more usable now :D
(thx john)

PR #16, opened 2015-04-29
introduced additional isRegisterAllowed() callback

alex-ellwein:
With the current implementation of ``RegistrationHandler`` there is an issue that ``ClientRegistry#registerClient()`` is called **before** the actual registration acknowledge (201 CREATED or an error) is being sent by back the server. So if you are, for instance, implementing a LwM2M read request directly in ``ClientRegistry#registerClient()`` or in the same thread there after, then the LwM2M client will process the particular read request before sending the actual registration response. Some LwM2M clients we have seen, do even refuse to process a read request *before* the registration response, other will go happily without it.

With the proposed PR the semantic of the ``ClientRegistry`` is changed a little bit:
the previously used ``ClientRegistry#registerClient()`` has been renamed to ``ClientRegistry#isRegisterAllowed()`` (as the old register() also used to return a boolean), and is called in the same place (before responding) to signal if the client will be registered. After the response has been sent successfully to the client, the ``ClientRegistry#registerClient()`` - the new one - is called.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/ClientRegistryImpl.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/ClientRegistryImpl.java
@@ -12,49 +12,30 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
+ *     Bosch Software Innovations GmbH - externalize registry listeners
  *******************************************************************************/
 package org.eclipse.leshan.server.impl;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
 
-import org.eclipse.leshan.server.Startable;
-import org.eclipse.leshan.server.Stoppable;
 import org.eclipse.leshan.server.client.Client;
 import org.eclipse.leshan.server.client.ClientRegistry;
-import org.eclipse.leshan.server.client.ClientRegistryListener;
 import org.eclipse.leshan.server.client.ClientUpdate;
 import org.eclipse.leshan.util.Validate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * In memory client registry
+ * In-memory client registry.
  */
-public class ClientRegistryImpl implements ClientRegistry, Startable, Stoppable {
+public class ClientRegistryImpl implements ClientRegistry {
 
     private static final Logger LOG = LoggerFactory.getLogger(ClientRegistryImpl.class);
 
     private final Map<String /* end-point */, Client> clientsByEp = new ConcurrentHashMap<>();
-
-    private final List<ClientRegistryListener> listeners = new CopyOnWriteArrayList<>();
-
-    @Override
-    public void addListener(ClientRegistryListener listener) {
-        listeners.add(listener);
-    }
-
-    @Override
-    public void removeListener(ClientRegistryListener listener) {
-        listeners.remove(listener);
-    }
 
     @Override
     public Collection<Client> allClients() {
@@ -67,28 +48,15 @@ public class ClientRegistryImpl implements ClientRegistry, Startable, Stoppable 
     }
 
     @Override
-    public boolean registerClient(Client client) {
+    public void registerClient(Client client) {
         Validate.notNull(client);
-
         LOG.debug("Registering new client: {}", client);
-
-        Client previous = clientsByEp.put(client.getEndpoint(), client);
-        if (previous != null) {
-            for (ClientRegistryListener l : listeners) {
-                l.unregistered(previous);
-            }
-        }
-        for (ClientRegistryListener l : listeners) {
-            l.registered(client);
-        }
-
-        return true;
+        clientsByEp.put(client.getEndpoint(), client);
     }
 
     @Override
     public Client updateClient(ClientUpdate update) {
         Validate.notNull(update);
-
         LOG.debug("Updating registration for client: {}", update);
         Client client = findByRegistrationId(update.getRegistrationId());
         if (client == null) {
@@ -97,10 +65,6 @@ public class ClientRegistryImpl implements ClientRegistry, Startable, Stoppable 
             Client clientUpdated = update.updateClient(client);
             clientsByEp.put(clientUpdated.getEndpoint(), clientUpdated);
 
-            // notify listener
-            for (ClientRegistryListener l : listeners) {
-                l.updated(clientUpdated);
-            }
             return clientUpdated;
         }
     }
@@ -108,20 +72,9 @@ public class ClientRegistryImpl implements ClientRegistry, Startable, Stoppable 
     @Override
     public Client deregisterClient(String registrationId) {
         Validate.notNull(registrationId);
-
         LOG.debug("Deregistering client with registrationId: {}", registrationId);
-
         Client toBeUnregistered = findByRegistrationId(registrationId);
-        if (toBeUnregistered == null) {
-            return null;
-        } else {
-            Client unregistered = clientsByEp.remove(toBeUnregistered.getEndpoint());
-            for (ClientRegistryListener l : listeners) {
-                l.unregistered(unregistered);
-            }
-            LOG.debug("Deregistered client: {}", unregistered);
-            return unregistered;
-        }
+        return toBeUnregistered == null ? null : clientsByEp.remove(toBeUnregistered.getEndpoint());
     }
 
     private Client findByRegistrationId(String id) {
@@ -135,45 +88,5 @@ public class ClientRegistryImpl implements ClientRegistry, Startable, Stoppable 
             }
         }
         return result;
-    }
-
-    /**
-     * start the registration manager, will start regular cleanup of dead registrations.
-     */
-    @Override
-    public void start() {
-        // every 2 seconds clean the registration list
-        // TODO re-consider clean-up interval: wouldn't 5 minutes do as well?
-        schedExecutor.scheduleAtFixedRate(new Cleaner(), 2, 2, TimeUnit.SECONDS);
-    }
-
-    /**
-     * Stop the underlying cleanup of the registrations.
-     */
-    @Override
-    public void stop() {
-        schedExecutor.shutdownNow();
-        try {
-            schedExecutor.awaitTermination(5, TimeUnit.SECONDS);
-        } catch (InterruptedException e) {
-            LOG.warn("Clean up registration thread was interrupted.", e);
-        }
-    }
-
-    private final ScheduledExecutorService schedExecutor = Executors.newScheduledThreadPool(1);
-
-    private class Cleaner implements Runnable {
-
-        @Override
-        public void run() {
-            for (Client client : clientsByEp.values()) {
-                synchronized (client) {
-                    if (!client.isAlive()) {
-                        // force de-registration
-                        deregisterClient(client.getRegistrationId());
-                    }
-                }
-            }
-        }
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ bin/
 *.war
 *.ear
 leshan-core/ddffiles
+.idea/
+*.iml
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
@@ -12,6 +12,7 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
+ *     Bosch Software Innovations GmbH - adapted for registration acceptor
  *******************************************************************************/
 package org.eclipse.leshan.server.californium;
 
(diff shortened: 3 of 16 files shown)


sbernard31:
Thx Alex, 

I think you point a real problem. We surely need a way to execute code just after we receive a registration. Currently we can do that by implementing a client registry or add a listener on it.
But as you say the code will be executed before we send the 201 CREATED to the client.
Currently, this is possible to do that by "asyncronizing"  this and add a crappy sleep but is not really elegant. (and not really sure)

About the solution you proposed, I'm not sure this is the best solution.
Maybe a better way to solve this issue is to use listener and so fire the clientRegistered event only after we sent the 201 CREATED. (and maybe even after we receive the ack)
To do that we surely need to separate the listener mechanism from clientRegistry to the LwM2mServer.
(I think we have exactly the same problem with the update and deregistration and we can solve this in a same way)

alex-ellwein:
Simon,
the reason why i implemented it this way is because we discussed in team and were not sure if the ``boolean`` response from previous ``registerClient()`` is needed by an application; maybe it is, then it actually would be right thing to change it to ``isRegisterAllowed()`` (it decides actually if the registration would proceed or not, 201 or error, could be useful to prevent the registration in some cases). So to keep the logic in ``registerClient()`` and to avoid API breaks (sure extending the app with one method always returning true by default *is* an API extension, but not a hard break), and for existing applications using ``ClientRegistry`` it's all fine.
We already tried the suggested asynchronous "crappy sleep" and also other workarounds, but they seem not to work with all tested clients as expected because of different processing times and network latencies. With this solution, i think we have a minimal intrusion in the API and the protocol flow is corrected.

sbernard31:
Add a `isRegisterAllowed()` method is maybe cleaner than the current API. I don't know. (This is not directly linked to this issue)
But I think this is not clean to send the 201 CREATED before we really create the registration.
And breaking the API is not really a problem, I mean we will not break the API just for fun, but if it's needed we must change it. The project is young, an the spec is not stable so this will not be the last break.

alex-ellwein:
IMHO if i get it right from the flow,  it is now a "the best effort" solution -- because the ``registerClient()`` is called after ``exchange.respond(ResponseCode.CREATED)``. The ACK response to client is a non-confirmable message, so we don't know exactly _if and when_ the client receives it (please consider [this issue](https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/7) for maybe the best flow proposal). But the spec is still not very clear and precise on this topic.
Since the registration state is pretty volatile - it can happen anytime for no reason, for instance, a device crashes or restarts even before any response from server arrives, i would prefer this solution as less time- and code-intensive.

jvermillard:
@alex-ellwein even with your fix I think you still need to fix your client. 
Because if the 2.01 CREATED answer from the server is lost, the server won't know it and will still try to send the device management command and your client won't like it because the registration answer was lost in the UDP space.

alex-ellwein:
@jvermillard that's what i wrote in my previous comment. You cannot guarantee the response has arrived because its a non-confirmable acknowledge, there is no response to it. Either we wait for change in the spec, but i imho would prefer a solution which fixes the callback order and works in the most cases right now.

jvermillard:
For reference the same issue on wakaama: https://github.com/eclipse/wakaama/pull/24

jvermillard:
Splitting the method in two call (isAllow and register) can lead to send 2.01 CREATED before doing the register in the registry.
For example you have an error during the registration (database error, backend 503, whatever) the device will think the registration was successful.

BTW we use isAllow in AirVantage for sending 400 error to device connecting with correct DTLS credentials but the device is over quota: e.g. the customer forgot to pay his bills :)

I think we should break the API and make something future proof so we can declare the API stable sooner :)

jvermillard:
So one solution could be to move the registrationListener on the Lwm2mServer and remove it from the ClientRegistry
So the ClientRegistry is not responsible of propagating the registration event to the user code, just to accept the registration (or not) and to store it.

WDYT?

alex-ellwein:
Julien, 
i think the problem is much deeper and i will try to explain what i think. 
I agree that this solution *may* lead to side effects, described below. But nevertheless anything what we do now to avoid this problem is mere workaround or best effort solution.
**The actual problem** is the untransactional behavior of the registration request, as described here:

https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/7

Let's consider it on examples.

**Current Leshan:**
The app using the subsequent GET after the registration, with the current Leshan may misbehave like this:
```
Client               Server          ClientRegistry            App
   | ->- CON POST ->   |                    |                   |
   |                   |  register()  -->   | ----(listener)--> |
   | <---  too early next request, possible, e.g. GET /3/0 ---  |
   | <-- ACK to POST - |                                             
```
Because the current impl calls ``registerClient()`` before sending ACK, you will agree that it is 
possible (i would say, most the time) to put a next request from the app (from actually synchronous callback) in between. 
With the "fix" you mentioned from David Navarro the client will drop the following GET before ACK from registration, so that the client would have to resend it, then in meantime the ACK for register is reached from server. But - still a hack and works maybe only with Wakaama, other client impl will not work.

**This solution:**
```
Client               Server            ClientRegistry            App
   | ->- CON POST ->   |                      |                   |
   |                   | isRegisterAllowed?-> |                   |
   |                   | <--- true -------    |                   |
   | <- ACK to POST -- |                      |                   |
   |                   | -registerClient()--->| --(listener)----> |
   | <------ possible GET /3/0 is now after ACK ----------------- |  
```
Possible side effects:
* ACK gets lost and is **not** received by the server:
  * What will happen: the client will retry the POST -> same flow is repeated -> ``registerClient()`` is   called again. Client is registered twice. I mean no problem here, since the client may crash or restarted after registration, it is usual situation with the current flow. Server has to deal with it anyways.
* GET comes before ACK:
  * using UDP there is no guarantee that the ACK arrives before the next GET. You can never know, 
    even with other scenarios. If client silently drops/ignores it, until ACK arrives (as with Wakaama), GET is then either respond by the client or retried by the server. 

##### My conclusions:
Right now there is no way to determine on server that client also accepted the register ACK. That's what i mean by being "transactional registration". The register flow must be extended as proposed in https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/7 or every LwM2M client *must* implement the same mechanism of ignoring any request before POST ACK, which is not specified. 
What will be our directions here?

jvermillard:
@alex-ellwein thanks for the details.
I think we agree we need to break the API compatibility here :) and provide two methods.
Since we are still targeting June for a stable server API, I think we should step back and think about the ClientRegistry API design and fix it in a sustainable way. I don't like the way the ClientRegistry which is basically an interface hiding a database or whatever, is also in charge of propagating events.

I known it's frustrating because you need the workaround now but I don't want to merge it and break compatibility again a week after. It would be even more frustrating for you :)

BTW I'm on vacation for 2 weeks.

sophokles73:
I tend to agree with Julien, we should explicitly assign the responsibility for notifying listeners about registration events to a component other than the ClientRegistry first.

leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/ClientRegistryImpl.java
```diff
@@ -67,22 +67,32 @@ public Client get(String endpoint) {
     }
 
     @Override
-    public boolean registerClient(Client client) {
-        Validate.notNull(client);
+    public boolean isRegisterAllowed(final Client client)
+    {
+       Validate.notNull(client);
+       LOG.debug("checking if Client {} is allowed to register", client);
+       clientsByEp.put(client.getEndpoint(), client);
```
jvermillard: if you do that now, I suppose when you call registerClient(..) the unregistration event won't work correctly (line 83)

alex-ellwein: You got it right, the nasty side effect here is, that the unregister() will always happen once before the register(), because 'previous' is set. And the listeners have to deal with that...

alex-ellwein:
Hello all together,
here is an other attempt to separate the client persistence logic done in
ClientRegistryImpl from notifying the ClientRegistryListeners.

The concerns related to the listeners, are now separated in
two main interfaces: ``ClientRegistryListenerManagement`` (for add/remove
operations) and ``ClientRegistryNotification`` (for the listener
notifications). These can be provided either in LeshanServer or via the
LeshanServerBuilder. The notifications are triggered only by the
RegistrationResource and the client registry is used only in a straight
"persistence" manner in RegistrationHandler.

I also externalized the functionality to allow/disallow
client registrations in a ``RegistrationAcceptor`` interface which can
also be passed to the LeshanServerBuilder and externalized the
RegistrationCleaner in a separate class.

bbrighttaer:
Hello, is it possible to get the javadoc of the current leshan?

sbernard31:
@bbrighttaer  is it linked with this current pull-request in any way ? 
If not, please, ask this on https://dev.eclipse.org/mailman/listinfo/leshan-dev

sbernard31:
Hi Alex,
Thx for the PR :), I just take a look at this and I see 2 problems : 
a) we lost the unregistered notification when a registration is cleaned.
b) we lost the unregistered notification when we re-register a device.

I also have some concerns about some news classes.
1) Does it necessary to separate `ClientRegistryListenerManagement` and `ClientRegistryNotification` ? 
It's seems that we need to know listeners to be able to fire notifications
2) Does it necessary to separate `ClientRegistry`, `RegistrationAcceptor`, and `RegistrationCleaner`  ? 
RegistrationCleaner implementation seems dependent of ClientRegistry implementation.
3) I don't understand the interest of `ClientAware` interface.
4) I'm not totally satisfied with the way we hold the `Client` from `RegistrationHandler` to `ClientRegistryNotification` but I can not find a better solution for now.

I will take some time to think more deeply about this problem.

sbernard31:
I start another [branch](https://github.com/eclipse/leshan/tree/client_registry_refactoring) to experiment and share some code about this issue.

In [this commit](https://github.com/eclipse/leshan/commit/776c8e48595e050569553a7b6c5b089a18485697), I don't try to split the client registry, I just focus on the 4) problem.
I introduce a callback in LwM2MResponse, this is more consistent in term of responsability.
`RegistrationHandler` handle lwm2m registration logic.
`RegistrationResource` is just responsible to translate CoAP request in LWM2M request and LWM2M response in CoAP response.

alex-ellwein:
Hi Simon,
thank you for reviewing the pull request.
Indeed the unregister notification is not passed if a device already registered. 
I am also feeling that you are right about the concerns, but I honestly didn't know how to separate ``RegistrationHandler`` from the ``RegistrationResource`` cleanly. 
I think the idea with the callback on ``Lwm2mResponse`` is not bad, however, ``RegistrationHandler`` contains IMHO still too much logic and should be simplified. What do you think?

sbernard31:
@alex-ellwein, about the `RegistrationHandler` maybe it could be simplified. Which logic would you like to extract from `RegistrationHandler` ?

alex-ellwein:
In particular, i don't like the following code block in ``register()``, it's rather long and hard to read. Maybe it would be good to extract it to a function?
Line 68 in your version.
```
            // if this is a secure end-point, we must check that the registering client is using the right identity.
            if (registerRequest.isSecure()) {
                PublicKey rpk = registerRequest.getSourcePublicKey();
                String pskIdentity = registerRequest.getPskIdentity();

                if (securityInfo == null) {
                    LOG.debug("A client {} without security info try to connect through the secure endpont",
                            registerRequest.getEndpointName());
                    return new RegisterResponse(ResponseCode.FORBIDDEN);
                } else if (pskIdentity != null) {
                    // Manage PSK authentication
                    // ----------------------------------------------------
                    LOG.debug("Registration request received using the secure endpoint {} with identity {}",
                            registrationEndpoint, pskIdentity);

                    if (pskIdentity == null || !pskIdentity.equals(securityInfo.getIdentity())) {
                        LOG.warn("Invalid identity for client {}: expected '{}' but was '{}'",
                                registerRequest.getEndpointName(), securityInfo.getIdentity(), pskIdentity);
                        return new RegisterResponse(ResponseCode.FORBIDDEN);
                    } else {
                        LOG.debug("authenticated client {} using DTLS PSK", registerRequest.getEndpointName());
                    }
                } else if (rpk != null) {
                    // Manage RPK authentication
                    // ----------------------------------------------------
                    LOG.debug("Registration request received using the secure endpoint {} with rpk {}",
                            registrationEndpoint, DatatypeConverter.printHexBinary(rpk.getEncoded()));

                    if (rpk == null || !rpk.equals(securityInfo.getRawPublicKey())) {
                        LOG.warn("Invalid rpk for client {}: expected \n'{}'\n but was \n'{}'",
                                registerRequest.getEndpointName(),
                                DatatypeConverter.printHexBinary(securityInfo.getRawPublicKey().getEncoded()),
                                DatatypeConverter.printHexBinary(rpk.getEncoded()));
                        return new RegisterResponse(ResponseCode.FORBIDDEN);
                    } else {
                        LOG.debug("authenticated client {} using DTLS RPK", registerRequest.getEndpointName());
                    }
                } else {
                    LOG.warn("Unable to authenticate client {}: unknown authentication mode.",
                            registerRequest.getEndpointName());
                    return new RegisterResponse(ResponseCode.FORBIDDEN);
                }
            } else {
                if (securityInfo != null) {
                    LOG.warn("client {} must connect using DTLS ", registerRequest.getEndpointName());
                    return new RegisterResponse(ResponseCode.BAD_REQUEST);
                }
            }
```

alex-ellwein:
@sbernard31, i tested with [your branch](https://github.com/eclipse/leshan/tree/client_registry_refactoring) and it behaves very well. Good work.

PR #22, opened 2015-06-16
Json support

jvermillard:


sbernard31:
1) There still some code about object link, we should remove this right ? 
2) `LwM2mNodeEncoder` and `LwM2mNodeDecoder` begin to be fat, we should split it.
3) Not sure the `lwM2mJson` class is really necessary.
4)`JsonArrayElement` and  `LwM2mJsonObject` class name is not really consistent.
Maybe, I propose : `JsonArrayEntry` / `JsonRootObject` or `SenMLArrayEntry` / `SenMLRootObject`

jvermillard:
@alseddnm what do you think?

alseddnm:
Hello @sbernard31 and @jvermillard 
1. OBJECTLINK is just a new type, and it needed for JSON impl.. We did follow the specs and used test cases have ov (object link) field.. That why we added it.
2. LwM2mNodeEncoder (456 lines ) and LwM2mNodeDecoder (460  lines).. Not very fat -:) .. But if you want split am OK with that. 
3. LwM2mJson.java yes it can be removed and used the google gson API's. However don't you think is better to have a wrapper class for JSON type?
4. JsonArrayElement and LwM2mJsonObject, yes we can rename them.. Both names you have proposed are fine .. I would think JsonArrayEntry / JsonRootObject are better names :).

alseddnm:
@jvermillard why am getting permissions denied while trying to push changes to this branch

git -c diff.mnemonicprefix=false -c core.quotepath=false push -v --tags origin json:json
remote: Permission to eclipse/leshan.git denied to alseddnm.
fatal: unable to access 'https://github.com/eclipse/leshan.git/': The requested URL returned error: 403

jvermillard:
nizar you can't, just create a pull-request targeting this branch

sbernard31:
I rebased the code on master.
I added some integration tests, fixed an issue in json decoder remove the objectlink code and split the encoder/decoder.

I think this is in a mergeable state.
But, it still miss some features :
**1) support basename attribute in json**
**2) add a way to choose content format in Read Request** 
_The request **MAY** specify an Accept option containing the preferred Content-Format to receive. When the  specified content format is not supported by the LWM2M Client, the request MUST be rejected._
**3) support of "ct" attribute at registration**
_If the LWM2M Client supports the JSON data format for all the Objects it **SHOULD** inform the LWM2M Server by including the content type in the root path link using the ct= link attribute. An example is as follows (note that the content type value 1543 is an example, the actual value will be assigned by IANA for the LWM2M JSON format).
</>;ct=1543, </1/0>, </1/1>, </2/0>, </2/1>, </2/2>, </3/0>, </4/0>, </5>_
**4) support of historical value for notification**
_The JSON format also includes **optional** time fields, which allows for multiple versions of representations to be sent in the same payload. The time fields MUST only be used when sending notifications. Historical version of notifications are typically generated when “Notification Storing When Disabled or Offline” resource of LWM2M Server Object is set to true (see Appendix Appendix Z.) and when the Device comes on line after having been disabled for a period of time._

sbernard31:
Integrated in master (commit : bcf48457d7b928857327c744771696ccfc64d675)

PR #23, opened 2015-06-26
Fix TLV encoding for negative integer

msangoi:
Use signed magnitude representation instead of two's complement


msangoi:
OK with Wakaama client.

sbernard31:
Integrated in master (commit : 0e0bd6e1fdd09ddbdabed1ca0841c483d56975d5)

msangoi:
Reverted in master with commit bbebd77bd5848449306a942cdb517fac77150bae.

Since the spec is ambiguous about the signed integer representation (Opened OMA issue: https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/55#issuecomment-121798495), we will wait for a clarification before merging the fix.

jvermillard:
Seems fixed in OMA-TS-LightweightM2M-V1_0-20150910-D

PR #24, opened 2015-06-26
Add configurable timeout on synchronous send() method

sbernard31:
There are now 2 ways to send synchronous request.
one with a timeout parameter in millisecond, and another one without  timeout parameter which will try to send the request with all the CoAP retransmission attempts.


leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumLwM2mClientRequestSender.java
```diff
@@ -224,13 +218,21 @@ public void onReject() {
 
         public T waitForResponse() {
             try {
-                final boolean latchTimeout = latch.await(timeout, TimeUnit.MILLISECONDS);
-                if (!latchTimeout || coapTimeout.get()) {
+                boolean timeEllapsed = false;
```
jschloman: should be 'timeElapsed'

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumLwM2mClientRequestSender.java
```diff
@@ -224,13 +218,21 @@ public void onReject() {
...
+                if (timeout != null) {
+                    timeEllapsed = !latch.await(timeout, TimeUnit.MILLISECONDS);
+                } else {
+                    latch.await();
```
jschloman: Should we find a way to warn the user that this will now wait indefinitely?  Should we possibly put a maximum timeout of some very large number (say 60 minutes?) or some way for the user to unblock the client?  Also if the stop() method is called is there a way to have all of these latches possibly released?

leshan-client-example/src/main/java/org/eclipse/leshan/client/example/LeshanClientExample.java
```diff
@@ -111,7 +111,7 @@ public LeshanClientExample(final String localHostName, final int localPort, fina
             public void run() {
                 if (registrationID != null) {
                     System.out.println("\tDevice: Deregistering Client '" + registrationID + "'");
-                    client.send(new DeregisterRequest(registrationID));
+                    client.send(new DeregisterRequest(registrationID), 1000);
```
jschloman: It might be worthwhile doing this too on the above RegisterRequest as that could possibly block forever if there is no server, correct?  Or at least have it printout a "Attempting Registration with no timeout".

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -261,14 +257,22 @@ public void onReject() {
 
         public T waitForResponse() {
             try {
-                final boolean latchTimeout = latch.await(timeout, TimeUnit.MILLISECONDS);
-                if (!latchTimeout || coapTimeout.get()) {
+                boolean timeEllapsed = false;
```
jschloman: 'timeElapsed'

sbernard31:
Integrated in master (8ef0d4d145b90d30a95c7d843fd4f27b11fdf383)

PR #25, opened 2015-06-26
Update to maven-bundle-plugin to match client refactoring.

dwoodard1:
Signed-off-by: David Woodard <david.woodard@eurotech.com>

The client bundles pom.xml files need to be updated to match the client refactoring. The change will allow these bundles to be used in an OSGi container.


sbernard31:
Thx :)

PR #28, opened 2015-07-01
Spec change : Use POST for registration Update instead of PUT

sbernard31:
The spec has changed since V1_0-20150615-D version, the registration update must be a CoAP POST instead of CoAP PUT.

We now accept PUT and POST to be compatible with older version of the spec.


msangoi:
:+1:

PR #31, opened 2015-07-13
Implement Queue Mode feature

alex-ellwein:
Intro
-----

A "queue mode" is an operation mode of LWM2M protocol (since the early versions

* A Client has to register using specified binding mode (e.g. using "UQ" for
  "UDP+Queue");
* A Server has to understand this binding mode (it is *mandatory*);
* A client is not expected to be always online - instead, a server must queue the
  requests and only allowed to send downlink requests one by one to the client
  when server received client's registration update
  (according to the spec, the client has to wait at least ACK_TIMEOUT seconds
  before it may go offline).

Basic Concepts
--------------

This implementation focuses on creating a new ``LwM2mRequestSender`` called
``QueueRequestSender``. This sender is meant to be a decoration for the already
existing ``CaliforniumLwM2mRequestSender`` and to provide only an asynchronous
mode sending. The queue sender is aware of client's binding mode and transparently
delegates the requests to the underlying ``CaliforniumRequestSender``, if a client
is not in queue mode. In case of queue mode operation, the ``QueueModeRequestSender``
maintains an in-memory request queue and the outstanding (Downlink-)request is
wrapped into ``QueueRequest`` object (with additional attributes, see below),
created by the ``QueueRequestFactory``, and is enqueued to the queue (with the
status "ENQUEUED", see below for the request state descriptions). The queue is
processed by the queue reactor, which schedules the queue tasks, responsible
for queue processing and request state transitions.

The queue requests are created as single requests by default. It is also
possible to create queue request sequences, i.e. a sequence of requests which
keeps the sequential order inside the sequence and allows to priorize the
request sequences up and down across the queue (see "Queue Management").

Queue Persistence
-----------------

In the current implementation, the request queue (or queues, as they are
organized per client) are persisted only in memory, however, one can easily
replace a ``RequestQueue`` implementation to provide, for instance, a DB-based
queue persistence. The ``QueueRequestSender`` supports, along with the usual
request sending timeout, two configurable intervals. These are:

* send expiration interval - how long should the queue sender try to send a request
  before it gets expired or no longer needed;

* keep expiration interval - how long should the queue keep the requests (even
  elapsed ones) archived in the queue for the history purpose.

Queue Request States
--------------------

1. A fresh created queue request which is sent via ``QueueRequestSender`` is
put in the queue with the state ``ENQUEUED``. A request remains in this state,
until the destination client sends registration update or a notify. (exception:
client has not yet updated for its first time, so the queue will try to send
the request as best effort as soon as possible).

2. A client's update or notify triggers a queue request to be put in
``PROCESSING`` state. This indicates that the queue request is currently being
sent. For every client, there can be only one request in a processing state,
because all of the requests are processed strictly sequential (per client).

3a. If a processing request is timed out, it returns to the queue with the
state ``DEFERRED``. This means, that after next client's update/notify this
request will be sent as first (unless the send expiration interval is
already elapsed, in this case it will be put into ``TTL_ELAPSED`` state as
described in 4a.)

3b. If a processing request could be sent, then it is put into ``EXECUTED``
state. The request remains archived in the queue with this state, until
a keep expiration interval is elapsed.

4a. A ``DEFERRED`` request, which could not be sent at all, is put into
``TTL_ELAPSED`` state. It is kept in the queue, until the keep expiration
interval elapses.

Queue Management
----------------

The additional ``QueueManagement`` interface provides basic functionality
for managing requests in the queue. The management includes following
functions:

* move a single request up/down the queue;
* move a single request to the top/bottom of the queue;
* move a request sequence up/down the queue;
* move a request sequence to the top/bottom of the queue.
* drop a single request or request sequence.

Limitations
-----------

As said before, this particular implementation of the queue mode allows
only an in-memory persistence of the request queue. The queue management
also lacks a frontend/UI or REST API and is also limited to programmatic
management only. However, there are possibilities for extension of the
functionality upon the defined interfaces:

* ``RequestQueue``
* ``QueueManagement``
* ``QueueRequest``
* ``QueueRequestFactory``

and so on.

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/request/BindingMode.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/request/BindingMode.java
@@ -12,8 +12,12 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
+ *     Alexander Ellwein (Bosch Software Innovations GmbH) 
+ *                     - extended for using with queue mode
  *******************************************************************************/
 package org.eclipse.leshan.core.request;
+
+import java.util.EnumSet;
 
 /**
  * Transport binding and Queue Mode
@@ -36,5 +40,14 @@ public enum BindingMode {
     US,
 
     /** UDP with Queue Mode and SMS */
-    UQS
+    UQS;
+
+    private static final EnumSet<BindingMode> QUEUE_MODES = EnumSet.of(UQ, SQ, UQS);
+
+    /**
+     * @return true, if a binding mode is a queue mode, otherwise false.
+     */
+    public boolean isQueueMode() {
+        return QUEUE_MODES.contains(this);
+    }
 }
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
@@ -1,0 +1,142 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *     Alexander Ellwein (Bosch Software Innovations GmbH)
+ *                     - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.leshan.integration.tests;
+
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.HashSet;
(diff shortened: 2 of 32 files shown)


sophokles73:
Hi Alex,

thanks for submitting this PR. I have added some comments to the code which I kindly ask you to review.

Could you also please rebase your PR to leshan's current master's HEAD?

leshan-core/src/main/java/org/eclipse/leshan/core/request/BindingMode.java
```diff
@@ -36,5 +37,12 @@
     US,
```
Areontar: I know this is out of scope of this PR (maybe even irrelevant), it would be nice to try and use EnumSet instead of specific enum per combination. As more Transport layer are added (in the future), this will become hard to manage.

alex-ellwein:
Kai,
thank you for reviewing. I will try to integrate the improvements soon as well as rebase it to the latest master.
@Areontar thanks for the hint. It is definitely a good point to use EnumSets.

alex-ellwein:
Hi,
i've rebased the pull request onto latest leshan's master and also integrated the suggestions from @sophokles73 and @Areontar.

alex-ellwein:
Hi all,
after discussion with my team we decided to withdraw this pull request and to provide the contributions for queue mode step-by-step in smaller portions. This way we want to involve other developers in discussing and reviewing the code in order to have a broader feedback for a review. More pull requests will follow. Thank you.

sbernard31:
Thx Alex, This will be easier for us to integrate this :)

PR #37, opened 2015-10-06
Fix TLV identifier serialization and COAPs support

lulf:
* Current head build does not work at all with setting COAPIFACES, as the value is not propagated to the correct setter
* TLV identifiers in the range 8-16 bit are not serialized according to spec.

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/tlv/TlvEncoder.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/tlv/TlvEncoder.java
@@ -27,6 +27,10 @@ public class TlvEncoder {
 
     private static final Logger LOG = LoggerFactory.getLogger(TlvEncoder.class);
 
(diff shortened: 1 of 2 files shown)


sbernard31:
Thx  luif :)
 Could you please [sign the eclipse CLA](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Foundation_Contributor_License_Agreement) and [sign your commits](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#The_Commit_Record) too.

lulf:
Right, I'll have to open a new pull request, since I see that i committed with wrong email account.

sbernard31:
In this case you can change your commit (e.g. with interactive rebase), then do a push force in your banch, (but close the pull-request and open a new one is ok too)

lulf:
See https://github.com/eclipse/leshan/pull/38

PR #38, opened 2015-10-07
Fix TLV identifier serialization and COAPs support

lulf:
* Current head build does not work at all with setting COAPIFACES, as the value is not propagated to the correct setter
* TLV identifiers in the range 8-16 bit are not serialized according to spec


sbernard31:
I add a unit test for 8-16bit identifier issue.
This is now integrated in master (commit 5b4628561967fce178dbbe88e2863550497d66d6 and 3405093137bb1a838ed39b98ed2d577da1c2142d )

Thx :)

PR #39, opened 2015-10-14
Refactoring lwm2m response

sbernard31:
1) I add the possibility to add a message error on `LwM2mResponse`. (following the [CoAP spec](https://tools.ietf.org/html/rfc7252#section-5.5.2))
2) I add syntactic sugar to simplify the `Lw2mResponse` construction.
3) I fix the issue about the new `ObserveResponse` (casting is no more needed)
4) I enhance the way we handle the `observe response` and the `notify`.

(ok ok I should have done 4 commits not 2 :p)


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LwM2mResponseBuilder.java
```diff
@@ -237,53 +247,49 @@ public void visit(final DeleteRequest request) {
...
         case CONTENT:
-            if (coapResponse.getOptions().hasObserve()) {
-                // observe request succeed so we can add and observation to registry
-                final CaliforniumObservation observation = new CaliforniumObservation(coapRequest, client,
-                        request.getPath(), model);
-                coapRequest.addMessageObserver(observation);
-                observationRegistry.addObservation(observation);
-                // add the observation to an ObserveResponse instance
-                lwM2mresponse = buildObserveResponse(request.getPath(), coapResponse, observation);
+            // just handle first response as observe request (the next one will be handle as NOTIFY)
```
msangoi: typo in comment: request/response

msangoi:
Ok for me. 
Cleaner API :+1:

sophokles73:
I particularly like the _syntactic sugar_ :-) I have added some comments as well...

jschloman:
I like this syntactic improvement.  It makes the response interface very similar to that in javax.ws.rs.  A couple of general comments:
1.  With all of the provided static constructor methods, is there any need for the true constructors to be public?  Couldn't they all be protected/private?  This would go a long way to enforce the style in future revisions of Leshan.
2.  For a successful response, should an exception be thrown if getErrorMessage is called (say the way Optional in Java 8 throws NoSuchElementException if .get() is called and there is no member)?
3.  In a similar way private static subclasses (e.g., SuccessfulResponse, FailureResponse) could be introduced to reduce conditional logic on the Response classes (i.e., SuccessfulResponse could throw a NSE exception on getErrorMessage and not have a constructor that takes an ErrorMessage while FailureResponse would toString() with the ErrorMessage provided, etc).  This could also prevent cases like a user able to getLocation() from an unsuccessful CreateResponse.

I like this change quite a bit :)

sbernard31:
@jschloman  
1. We discussed about that with Manu and Julien, but we are worried this could be too restrictive/directive.
2. 3. I don't know. At first sight, this seems to me a little overkill. But I will think about that and get the Manu and Julien feeling.

Thx all for your feedback. :D

sbernard31:
This is now integrated in master (commit  7db865a7b8016a128d7b9fcfdd538c4e74deaef9 and f4cf26865cc4420a8f3504424d5bfdd329799ec3)

PR #40, opened 2015-10-16
Now the bootstrap server send end-of-boostrap message

jvermillard:
Following last spec revision, when the bootstrap is done we send a POST /bs to the device


sbernard31:
integrated in master (commits e353e9dd8f6333e2f893a118bb22cf3b0636a535 and e1bca39b313719c7d260fd15a8c9e9623c93c693)

PR #42, opened 2015-10-22
add factory for object instance creation at client side

sbernard31:
After [discussion](https://github.com/eclipse/leshan/commit/7ad1b7c9611c3ed086e1a511b33e597c855a76d1#commitcomment-13922970), I add a factory to create object instance.

### DIFF
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/BaseInstanceEnabler.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/BaseInstanceEnabler.java
@@ -18,7 +18,6 @@ package org.eclipse.leshan.client.resource;
 import java.util.ArrayList;
 import java.util.List;
 
-import org.eclipse.leshan.core.model.ObjectModel;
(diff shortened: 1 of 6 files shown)


jschloman:
It might also be useful to add this into the leshan-client-example as I think this would be a useful feature to demonstrate to users.

sophokles73:
I think the classes in `org.eclipse.leshan.client.resource` could use some JavaDoc as well explaining the underlying concepts and some example code snippets ... :-) I will try to add some during the weekend ...

sbernard31:
Should I merge this one ?

sophokles73:
@sbernard31,

I like your latest changes which exactly reflect my thinking :-)
I have added some JavaDoc to `LwM2mInstanceEnabler` as a starting point for more to come.
However, I think you should go ahead and merge this to master now ... Maybe squashing the multiple commits into one or two before?

sbernard31:
I merged it in master (commit 9daf45b48aeea1ce3094ffe33eeb85920e4cafd1)
Thx for the feedbacks :)

PR #43, opened 2015-10-27
refactoring to add identifier at resource instance level

sbernard31:
We currently have an issue with resource instance, the data structure does not handle correctly that resource instances should have an Identifier, so with Manu we try to find a solution.

While I'm working on that, we also found that the `type` information of the resource should be handle at resource level, not at value level. (because all the instances of a multiple resource have the same type).
This led us to remove the `Value` class.

To simplify the code and avoid [some confusion](https://dev.eclipse.org/mhonarc/lists/leshan-dev/msg00252.html) about `DataType` vs `Type`. The `DataType` class was removed too.
(There is only one Type class in the Leshan code base now). That's mean that at LwM2mNode, all Integer resource are stored as long, and all float resource are stored as double.

Some javadoc is probably missing, but I would like to have feedback about that because to start finishing touch.


leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/json/LwM2mNodeJsonDecoder.java
```diff
@@ -94,72 +94,81 @@ private static LwM2mNode parseJSON(JsonRootObject jsonObject, LwM2mPath path, Lw
             Integer resourceId = Integer.valueOf(resourcePath[0]);
 
             if (!multiResourceMap.isEmpty() && multiResourceMap.get(resourceId) != null) {
```
msangoi: This "if" block could be part of the "if(resourcePath.length > 1)" block below. For readability.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
```diff
@@ -208,26 +209,26 @@ public void json_device_object_instance0() throws InvalidValueException {
 
         assertEquals(0, oInstance.getId());
 
-        assertEquals("Open Mobile Alliance", (String) oInstance.getResources().get(0).getValue().value);
-        assertEquals("Lightweight M2M Client", (String) oInstance.getResources().get(1).getValue().value);
-        assertEquals("345000123", (String) oInstance.getResources().get(2).getValue().value);
-        assertEquals("1.0", (String) oInstance.getResources().get(3).getValue().value);
+        assertEquals("Open Mobile Alliance", ((LwM2mSingleResource) oInstance.getResources().get(0)).getValue());
```
msangoi: Cast not needed here.

leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mMultipleResource.java
```diff
@@ -0,0 +1,218 @@
...
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+import org.eclipse.leshan.core.model.ResourceModel.Type;
+import org.eclipse.leshan.util.Validate;
+
+public class LwM2mMultipleResource implements LwM2mResource {
```
msangoi: It would be nice to rename this into MultipleResource. And remove the LwM2m prefix for all nodes objects. But we have an issue with LwM2mObject... 
Keeping the prefix only for Object is a bit weird. I can't find a nice solution for now.

msangoi:
:+1:

sbernard31:
No more feedback, I could go for adding javadoc :p ?

sophokles73:
Good for me :-) Go ahead ... I am at EclipseCon this week so please do not be too surprised if I am not overly responsive this week...

sbernard31:
integrated in master (commit 2543ab0b491ec83d5744d316d9fd7e0752645222)

PR #44, opened 2015-10-28
Replace usage of javax.xml.bind.DatatypeConverter

sophokles73:
Get rid of javax.xml.bind.DatatypeConverter usage because the javax.xml.bind package is not available on Android which I think should be a supported platform for the leshan client.

I also fixed a problem with parsing Base64 encoded OPAQUE data which was erroneously treated as hexBinary data ...

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/Lwm2mNodeEncoderUtil.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/Lwm2mNodeEncoderUtil.java
@@ -11,15 +11,23 @@
  *    http://www.eclipse.org/org/documents/edl-v10.html.
  * 
  * Contributors:
- *     Sierra Wireless - initial API and implementation
+ *     Sierra Wireless                               - initial API and implementation
+ *     Kai Hudalla (Bosch Software Innovations GmbH) - remove dependency on
+ *                                                     javax.xml.bind.DatatypeConverter
+ *                                                     which is not available on Android
  *******************************************************************************/
 package org.eclipse.leshan.core.node.codec;
 
 import java.util.Date;
 
+import javax.xml.datatype.DatatypeConfigurationException;
+import javax.xml.datatype.DatatypeFactory;
+import javax.xml.datatype.XMLGregorianCalendar;
+
 import org.eclipse.leshan.core.model.ResourceModel.Type;
 import org.eclipse.leshan.core.node.Value;
 import org.eclipse.leshan.core.node.Value.DataType;
+import org.eclipse.leshan.util.Hex;
 import org.eclipse.leshan.util.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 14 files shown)


jschloman:
LGTM.  Though there seem to be a few places where non-Base64 code had indentation/line-wrap applied that might not be the Leshan formatter.  Or maybe that's just how Github is displaying them...

sophokles73:
@jschloman,

you're right, in some places it looks strange but I am using the leshan code formatter and I am pretty sure that it is correctly formatted. Can you indicate one particular place where you think the formatting is wrong? I can then double check ...

jschloman:
@sophokles73 

I would say LwM2mNodeJsonDecoder, lines 82-83 for parseJsonPayLoadLwM2mResources, and lines 109-110 are some examples.  The lines look identical AFAIK but they seem to break differently now.  Again this could just be how Github is displaying them or some subtlety I am missing.

sbernard31:
The java formatter was rewritten in Mars. Since this, a same formatter configuration does format in a same way before or after Mars. Maybe this explain the formatting issue here.
(see http://eclipse-n-mati.blogspot.fr/2015/06/eclipse-mars-how-to-switch-back-to.html)

sophokles73:
@jschloman,

Thanks for the examples. I have checked and found out that my eclipse workspace actually didn't use the LeshanFormatter settings :-( 
I have tried to fix the formatting now. What do you think? Good to merge with master?

sbernard31:
There is no more formatting issue.
2 questions:
1) Could we integrate this in only one commit ? 
2) Why choosing to encode OPAQUE data in base64 in the LwM2mNodeSerializer which is used for the REST API ?  (I don't talk about the fix done in the LwM2mNodeJsonDecoder which should use the base64 as it's defined in the spec)

jschloman:
Looks good for formatting :+1:

sophokles73:
@sbernard31,

1) we could but since two issues were addressed (replace DatatypeConverter usage & fix erroneous handling of binary encoding), I thought it should be two commits. Is there a particular reason why you would prefer a single commit?

2) I wasn't sure about this myself and am therefore glad that you question it :-) I thought that for consistency reasons, we should also use Base64 encoding for this (which is also smaller in size than plain xsd:hexBinary). Again, is there a particular reason why we should use hexBinary encoding in this case while all OPAQUE values are encoded using Base64 when transferred in a LWM2M request/response?

sbernard31:
1)Not so important, but I'm a bit disturbed by the fact that in the first [commit you replace a base64 conversion in hexBinary](https://github.com/eclipse/leshan/commit/b5444a415c0953841f9aa28df0ec0ca2943fc140#diff-ec2db012607f64da3db55a4e98193754L140), then in the next commit you do the reverse change.
2) In fact, we choose hexBinary as it is more readable than base64. As Standalone is mainly used for testing, I think this make senses. (This value is directly print in the standalone UI and could also be read using browser inspector)

sophokles73:
@sbernard31,

1) you are totally right, I didn't realize that `LwM2mNodeJsonEncoder` actually used Base64 encoding already :-) I will merge the two commits into one ...
2) As I said, I am unbiased here. If it makes more sense to use hexBinary for display purposes we should keep it that way. I will change it back to hexBinary ...

sbernard31:
2)In absolute terms, I agree with you. base64 is much more effective than hexBinary to transport binary data as a string and we can display the data in the angular UI as hexBinary. But in this case we lost the readability when we use directly the REST API.

Anyway, if we do that we need to : 
**a)** also change the format [here](https://github.com/eclipse/leshan/commit/b5444a415c0953841f9aa28df0ec0ca2943fc140#diff-a65751d7c541eca763b1437d58716184L113) in leshan.core. (There is a risk to disturb leshan.core users as this will be a silent change, the ideal would be to handle the 2 formats in Lwm2mNodeEncoderUtil)
**b)** have a way to convert base64 to hexBinary and hexBinary to base64 in javascript to keep hexBinary in the UI.

In we think that's worth it, we should probably do that in another commit/branch/pullrequest as this is not directly linked to the DatatypeConverter replacement. Personally, I think the current solution is enough for our current needs.

sophokles73:
I have incorporated the latest changes discussed. Hope this works now for everyone ;-)

PR #47, opened 2015-11-04
Queue Mode: base interfaces for ongoing implementation

alex-ellwein:
Intro
=====

A "queue mode" is an operation mode of LWM2M protocol (since the early versions

* A Client has to register using specified binding mode (e.g. using "UQ" for
  "UDP+Queue");
* A Server has to understand this binding mode (it is *mandatory*);
* A client is not expected to be always online - instead, a server must queue
  the requests and only allowed to send downlink requests one by one to the client
  when server received client's registration update or notification.

- **org.eclipse.leshan.server.queue** - contains base interfaces to operate
  with queue mode, i.e. public API for sending downlink requests and managing
  the queue;
- **org.eclipse.leshan.server.queue.persistence** - contains interfaces to
  implement own queue persistence (we will provide only a basic in-memory
  persistence).

org.eclipse.leshan.server.queue - Queue Mode Public API
=======================================================

QueueRequestSender
------------------

``QueueRequestSender`` is an entry point for queue mode related sending.
This is basically an extension of ``LwM2mRequestSender``, which is capable
to maintain a queue of requests for every client registered with queue
binding mode and provides a basic queue management by means of
``QueueManagement`` interface.

QueueManagement
---------------

Queue Management is a public interface for interacting with the queue. It
can be used, e.g. to show QueueRequests for a particular client, drop a
single request or even a whole client's queue.

QueueRequest
------------

A single element of a queue of ``QueueRequestSender`` is represented by a
``QueueRequest``. Amongst the downlink request and client's endpoint, it
contains an information about request's state (``QueueRequestState``),
ID of the request and also an expiration date of the queue request, i.e.
the time point when the queue request should not be considered/send anymore.

Note: this is the user's view on a request, for persistence's view see ``QueueRequestEntity`` below.

QueueRequestState
-----------------

Every ``QueueRequest`` is passed through different states during its life
cycle. ``QueueRequestState`` enumeration describes following states:

1. A fresh created queue request which is sent via ``QueueRequestSender`` is
   put in the queue with the state ``ENQUEUED``. An enqueued request remains in
   this state, waiting to be send.

2. A client's registration update or notify triggers a queue request to be put
   in ``PROCESSING`` state. This indicates that the queue request is currently being
   sent. For every client, there can be only one request in a processing state,
   because all of the requests are processed strictly sequential (per client).

3. a) If a ``PROCESSING`` request is timed out while sending, it returns to the
   queue with the state ``DEFERRED``. This means, that after next client's
   update/notify this request will be sent as first (unless the send
   expiration interval is already elapsed, in this case it will be put into ``TTL_ELAPSED``
   state as described in 4)

  b) If a ``PROCESSING`` request is sent, then it is put into ``EXECUTED``
   state. The request will be then put into ``EXECUTED`` state.

 c) If a ``PROCESSING`` request has been deleted by the user while
      sending, it is put into ``PROCESSING_DELETED`` state. That is,
      the request will still be processed until the successful or
      unsucessful result of the sending operation, but will either
      be promoted to ``EXECUTED`` state and then deleted or deleted
      immediately upon a timeout of the send operation.
4. A ``DEFERRED`` queue request, which could not be sent within its expiration
   time, will be put in ``TTL_ELAPSED`` state.
5. Queue requests which are in states ``EXECUTED`` or ``TTL_ELAPSED`` may be
   removed from the queue (or archived for some period of time, depending on
   the queue's concrete implementation).

QueueRequestDeletedException
----------------------------

A ``QueueRequestDeletedException`` is needed to notify the sender on an explicit
removal of a ``QueueRequest``. It is thrown in the ``ErrorCallback`` registered
in the ``QueueRequestSender``.

org.eclipse.leshan.server.queue.persistence - Persistence Interfaces
====================================================================

QueuePersistence
----------------

``QueuePersistence`` provides basic operations for adding, retrieving, updating
and deleting ``QueueRequestEntity`` elements.

QueueRequestEntity
------------------

``QueueRequestEntity`` represents a persistence object for ``QueueRequest``.

QueueRequestEntityFactory
-------------------------

An interface for a factory, able to create ``QueueRequestEntities``, based
on given client's endpoint, downlink request and given Time-To-Live for the
request.

ExpirationChecker
-----------------

An ``ExpirationChecker`` allows to check an expiration of a particular
queue request and calculates the absolute time stamp for a queue request
expiration, note that the implementation of ``ExpirationChecker`` may be
dependent or specific to the ``QueuePersistence`` implementation.

Also-by: Daniel Maier <daniel.maier@bosch-si.com>


sophokles73:
This is a good starting point for the discussion, guys :-) Thanks for submitting this PR.

Apart from the naming issues, I would like to discuss a general thing:
Do we really need separate concepts for queued requests? We already have support for sending `LwM2mRequests` asynchronously, thus my feeling is that we should be able to bring `LwM2mRequest` and `QueueRequest` together somehow. My concern is that we will end up with hundreds of switch statements where we need to distinguish between a `QueueRequest` and a plain `LwM2mRequest` ...
From my point of view, a request being sent to client that is online should just be a special case of the more general case that a request is queued before delivery. We could even make queuing outbound requests the standard case and make support for *Queued Mode* simply a matter of TTL or queue capacity ...

Anyways, I think we should aim for a _tighter_ integration with the existing interfaces and concepts.
What do you think?

sbernard31:
I'm ok with the naming proposed by @sophokles73.

What is the real difference between `QueueRequest` and  `QueueRequestEntity` ?
Why should we define a `ExpirationChecker` interface ?
Maybe this are silly questions, but it's hard to have a good vision without code.

sophokles73:
@sbernard31,

I have to take responsibility for the lack of implementations of the interfaces. I asked our team to split up their implementation into multiple smaller chunks that are easier to digest than thousands of LOC at once. I think we should first try to agree on the core concepts we want/need for supporting the queued mode before we do the full implementation.
That said,I think you are asking exactly the right questions and I would like to encourage my colleagues from Bosch to provide the answers to them :-)

sbernard31:
Ok no problem, I think this is a good thing to contribute step by step. 
Do not hesitate to contribute code regulary even if the code in not perfectly written. We can do finishing touch later. Code is a good base of discussion.

DanielMaier-BSI:
Thanks for your comments so far.

I'm also OK with the naming proposed by @sophokles73.

@sophokles73 some thoughts about your other suggestions:

- Bring ``LwM2mRequest`` and ``QueueRequest`` together: From my point of view these two interfaces accomplish two different aspects. A ``LwM2mRequest`` is a request created by the user 
to be sent. A ``QueueRequest`` is a request that was already scheduled to be sent and is only for management purposes. A ``QueueRequest`` is immutable and not intended to be created by the user 
itself. To send requests with queue mode the concepts are the same. In both cases, queue mode and non queue mode, a ``LwM2mRequest`` needs to be created by the user and needs to be sent
with ``QueueRequestSender``. So I think there will be no switch cases because one will use these two object types in totally different places.

- Non queue mode special case of queue mode: For a moment I agreed here with you. But if we do this, behavior of sending non queue mode requests will change. Specification states out that queue mode requests must be sent in strictly sequential order. I think for a device that is "not so constrained" and does not use queue mode it is OK to send requests in parallel. I did not see anything in the spec that prohibits this. Please correct me if I am wrong. 
Also it is not clear if resend behavior of queue mode and non queue mode should be the same (also resend queue mode requests until max CoAP retries are reached). See also https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/32.
But please also see next point regarding tighter integration.

- Tighter integration: Yes, I agree that a tighter integration would be beneficial. Queue mode is a mandatory feature of a LWM2M server. So there is no need to give users the possibility
to create a leshan server instance without queue mode feature. I think it would be nice to have the management capabilities (see currently sending requests, see some statistics regarding 
requests, perhaps canceling requests) also for non queued requests. But this would require a huge refactoring effort to our current implementation and also the interfaces we are discussing here at the moment. Perhaps we should contribute this tighter integration later on.

So, what do you think?

DanielMaier-BSI:
@sbernard31,

``QueueRequestEntity`` is intended to be used for persistence purposes and only used by the queue mode implementation, not by the user itself. ``QueueRequest`` is intended to be used by the user in the context of queue mode management operations. In other words, these two interfaces are to separate persistence from application layer. If someone wants to implement his own persistence he needs to implement ``QueueRequestEntity``. ``QueueRequest`` is not intended to be implemented by anyone else.

The ``ExpirationChecker`` interface emerged when we realized that checking of expiration is persistence dependent. Our current in memory persistence for example uses ``System.nanoTime()`` to determine expiration of queued requests. A SQL based persistence might use a timestamp derived from database to check if a request is expired. So we can not do timestamp comparison inside queue mode implementation, but need to do this in persistence layer.
Well, of course we could also implement this logic into ``QueueRequestEntity`` but we wanted to keep this a "stupid" entity class without logic.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/QueueManagement.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.List;
+
+/**
+ * Queue Management interface allows users of QueueRequestSender to monitor the requests being queued and to influence
+ * them in some way, e.g. drop single request or even a whole queue.
+ *
+ * @see QueueRequestSender
+ */
+public interface QueueManagement {
```
jvermillard: I don't think we need this feature for a first implementation

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/QueueRequest.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * @return the actual downlink request which is to be send
+     */
+    DownlinkRequest<LwM2mResponse> getDownlinkRequest();
+
+    /**
+     * @return calendar date, when the time-to-send for the request expires and the sender will not anymore try to send
+     *         it.
+     */
+    Calendar getSendExpiration();
```
jvermillard: expiration could not be part of a first implementation

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/QueueRequestDeletedException.java
```diff
@@ -0,0 +1,31 @@
...
+ *     Alexander Ellwein, Daniel Maier (Bosch Software Innovations GmbH)
+ *                                - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+/**
+ * This special Exception is used to signal a kind of recoverable situation to the request sender. In case an
+ * application has been send a request while an other thread deletes it, this exception is passed in the error callback.
+ */
+public class QueueRequestDeletedException extends Exception {
```
jvermillard: I think we can also remove that from first step: just don't make request deletable/cancelable by the end user

jvermillard:
I like the effort to reduce the LoC in the PR :)

Queued request will probably disrupt the current request seding API (blocking API call will probably need to disappear).

I don't think it's the good way to do, I'm never a fan of "Big Design Upfront", I think we should try to make a first viable implementation and then extend it (no queue management, no expiration, etc..)

WDYT?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/persistence/ExpirationChecker.java
```diff
@@ -0,0 +1,42 @@
...
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue.persistence;
+
+import java.util.Calendar;
+
+/**
+ * Expiration checker is responsible to check expiration of a particular queue request (entity). An implementation of an
+ * expiration checker may be specific to a particular QueuePersistence implementation.
+ */
+public interface ExpirationChecker {
```
msangoi: I agree with @jvermillard about expiration.
Since expiration is not mentioned in the specification, I'm not sure it is needed in the API. 
From what I understand of the Queue Mode mechanism, requests are not expired but I assume they must be cleaned on client de-registration.
I think that expiration should be moved to the implementation. Does it make sense?

sophokles73:
@jvermillard,

when you say
> Queued request will probably disrupt the current request seding API (blocking API call will probably need to disappear)

does that mean that you are open to actually -disrupt- the current API, i.e. remove the blocking call, or does that mean that we can/should not add queued mode using the same `LwM2MRequestSender` interface?

Anyways, I like the discussion that we started around this :-) My proposal based on the arguments so far would be to start with an implementation that does very basic things only and evolve from there.
The most basic functionality could be:
* Sending a message to an offline client (queued request). No need for message expiration, re-ordering, queue management.
* Client gets queued message one after one in FIFO order once it becomes online again.

The rest could be added piece by piece in future PRs.

WDYT?

jvermillard:
>does that mean that you are open to actually disrupt the current API, i.e. remove the blocking call, or does that mean that we can/should not add queued mode using the same LwM2MRequestSender interface?
yes I think we should change/extend the current API, I think the first steps you are prposing are great for exploring API changes.

jvermillard:
BTW a 15 year old email on why small incomplete/broken contributions are better than finished ones:
https://mail-archives.apache.org/mod_mbox/cocoon-dev/200010.mbox/%3C39FAD0B5.2D3F404C@apache.org%3E

DanielMaier-BSI:
The expiration was intended when a device does not respond to a specific request due to a bug. We wanted to prevent hanging of the queue forever. But well, this would be a valid point to omit for the first version.

What we do need at the moment is the deletion functionality of the queue management. We decided to delete the queued entries of a client on de-registration. But we wanted to leave this decision to the application (to delete or not to delete on de-registration), because it is not clear stated in the specification. We also asked this question in OMA issue tracking (https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/69). But not answer so far.

alex-ellwein:
As a first step, I removed the expiration concepts from the API and done the renamings suggested by Kai. As @DanielMaier-BSI  mentioned, we need the ``QueueManagement`` at least for queue request removal.

sbernard31:
About `QueueManagemnt`, I'm not sure we need to expose queue request removal in the first implementation. Leshan can do it internally at deregistration in a first time.

sophokles73:
So, I think we would all like to see the first (not to be perfect) implementation of the basic functionality described above then :-)
Can somebody from Bosch please update the PR with the implementation classes?

And, following the great advice given by Stefano Mazzocchi (as pointed out by @jvermillard earlier), the fewer LOC the better :-)

vgajjala:
Good to a nice discussion
 
I agree with Bernard on combining the LwM2MRequest and QueueRequest. If we have two classes what comes to my mind is the server can queue up the requests when it see that the client registration expired using QueueRequest and also try to send Commands immediately using LwM2MRequest 

Also, I do you think the server should try to send the request at least once (can be configurable, no of retries) to the client if it see that  the registration has expired before queuing it up.

And then should be able to send and SMS to wake up the device after it reaches the retry count or send the command using alternate transport say SMS

Also, Is it mandatory that the Queue mode should be  FIFO ? The spec says 
> queued requests to the LWM2M Client in a serial fashion
but there could be a possibility that some requests/commands need to be delivered in queue and some need not or to put it differently may be we should have a persistence mechanism for requests/commands and have queue and non queue requests

balsmn:
Created a new PullRequest for this feature.

https://github.com/eclipse/leshan/pull/126

PR #48, opened 2015-11-05
API Improvements

sbernard31:
This pull request will contains severals API improvements.
(Details below)

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/request/WriteRequest.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/request/WriteRequest.java
@@ -36,8 +36,20 @@ import org.eclipse.leshan.util.Validate;
  */
 public class WriteRequest extends AbstractDownlinkRequest<WriteResponse> {
 
+    /**
+     * Define the behavior of a write request.
+     */
     public enum Mode {
-        REPLACE, UPDATE
+        /**
+         * Replaces the Object Instance or the Resource(s) with the new value provided in the “Write” operation. (see
+         * section 5.3.3 of the LW M2M spec).
+         */
+        REPLACE,
+        /**
+         * Adds or updates Resources or Resource Instances provided in the new value and leaves other existing Resources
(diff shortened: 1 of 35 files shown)


leshan-core/src/main/java/org/eclipse/leshan/core/request/WriteRequest.java
```diff
@@ -29,23 +36,116 @@
...
+        REPLACE, UPDATE
+    }
+
     private final LwM2mNode node;
     private final ContentFormat contentFormat;
+    private final Mode mode;
+
+    // ***************** write object ****************** //
+
+    public WriteRequest(final Mode mode, final ContentFormat contentFormat, final int objectId,
```
jvermillard: I would remove the ContentFormat from the constructor and let the user to use a setter for forcing JSON or anything else after the construction.

This would remove ~5 signatures for the constructor but make the class mutable.

@sophokles73 WDYT?

sophokles73: Since there are so many constructors already, I'd rather have five more than sacrificing immutability.
Since Simon has introduced *static creator methods* inother places as well, we could instead think about turning these constructors into something like
```
public static new(final Mode mode, final int objectId ...) {
  return new WriteRequest(...);
}
...
public static new(....) {
  return ...
}
...
```
Another question is whether we actually need to know the content format at the time the request object is instantiated. Or can we postpone that decision to the time the object actually gets marshalled?

jvermillard: when I say "mutable", I mean "mutable just for the content type to be set before marshalling" :)

sbernard31: I'm ok with @sophokles73 about immutability.
About static constructor, this will not be homogeneous with other requests and I'm not sure to see the benefits in this case,
I will integrated this in master (I have other request blocked by this one), but the discussion is still opened.

sbernard31:
integrated in master.

PR #66, opened 2016-01-26
Added Optional Registration Parameter (as a Map)

phtrivier:
Client can now be passed a free-form map of registration query parameters.

Some context : 

I have a custom `ClientRegistry`, that will respond to `Register` requests with an extra, non-standard "imei" parameter. It needs to do some validation on this parameter.

Rather than adding yet another field to the `RegisterRequest`, I added a free-form map (String to String) with all register query parameters. (Sounds future-proof enough)

I initially though about passing the `RegisterRequest` object in `ClientRegistry.register`, but I suppose this would break backward compatibility. So instead, I attached the parameter map to the client itself. 

(An alternative would be to have two `register` method with different arities , and using defaults methods - but is leshan supposed to be compiled for java 8 compatibility ?)

What do you think ?

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
@@ -18,7 +18,9 @@ package org.eclipse.leshan.server.californium.impl;
 import java.net.InetSocketAddress;
 import java.security.Principal;
 import java.security.PublicKey;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -167,9 +169,20 @@ public class RegisterResource extends CoapResource {
         if (request.getPayload() != null) {
             objectLinks = LinkObject.parse(request.getPayload());
         }
+
+        Map<String, String> registerParams = new HashMap<String, String>();
+        for (String param : request.getOptions().getUriQuery()) {
+            if (param != null) {
+                String[] tokens = param.split("\\=");
+                if (tokens != null && tokens.length == 2) {
+                    registerParams.put(tokens[0], tokens[1]);
+                }
+            }
+        }
+
         // Create request
         RegisterRequest registerRequest = new RegisterRequest(endpoint, lifetime, lwVersion, binding, smsNumber,
-                objectLinks);
+                objectLinks, registerParams);
 
         // Handle request
         // -------------------------------
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
@@ -88,6 +88,7 @@ public class RegisterResource extends CoapResource {
         try {
             super.handleRequest(exchange);
         } catch (Exception e) {
+            e.printStackTrace();
             LOG.error("Exception while handling a request on the /rd resource", e);
             exchange.sendResponse(new Response(ResponseCode.INTERNAL_SERVER_ERROR));
         }
@@ -151,6 +152,9 @@ public class RegisterResource extends CoapResource {
         String lwVersion = null;
         BindingMode binding = null;
         LinkObject[] objectLinks = null;
+
+        Map<String, String> registerParams = new HashMap<String, String>();
+
         // Get parameters
         for (String param : request.getOptions().getUriQuery()) {
             if (param.startsWith(QUERY_PARAM_ENDPOINT)) {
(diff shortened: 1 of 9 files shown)


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
```diff
@@ -167,9 +169,20 @@ private void handleRegister(CoapExchange exchange, Request request) {
         if (request.getPayload() != null) {
             objectLinks = LinkObject.parse(request.getPayload());
         }
+
```
jvermillard: should be done is the upper loop with the other param?

sbernard31:
It's pretty strange to add query params to RegisterRequest and Client.
I mean why just this CoAP option ? and why just on Register Request ? 

I think if we want to expose CoAP information in Leshan API, we must have a more global thought about that.
Currently, I see 2 possible solutions : 
  1. add a `Object getCoapRequest()` on lwM2mRequest and `Object getCoapResponse()` on lwM2mResponse, return type wil depends of the CoAP implementation used by Leshan.
  2.  expose our own minimalist CoAP message abstraction. A CoAP message is mainly a type, a code,  a payload and a set of option.

jvermillard:
Having access to CoAP messages is a good idea, but I don't think it's the good level in the stack.
LWM2M is modeled on top of REST and is going to be ported to HTTP/2 & MQTT, I  would prefer to expose common REST concepts like URL & parameters.

sophokles73:
I like the idea of being able to process not only standard query params but also have access to all query params submitted by a client.
@sbernard31, I do not see a general need to have access to basic CoAP concepts like codes and options which in my opinion should not be exposed to *user space code*. I do see value though in exposing *application level* information contained in URL and query params as @jvermillard already pointed out. We could extend `LwM2mRequest` with methods to get access to  query parameters and URL for that purpose. Consequently, we should then also provide access to this information in upstream listeners like the `ClientRegistryListener`.

phtrivier:
@sophokles73 : after some discussions with @sbernard31 and @jvermillard , here is another proposition : it's basically the same except we only add the "other" parameters in a free-form map. This way : 
* it's still open enough to support extra params
* this is not as far reaching a adding info to all requests (which is a bit of an overkill ; and we're not sure which abstraction we would expose that is not too tied to a particular coap implementation)

As the `Client` constructor was getting too big, I took the liberty to add a Builder pattern - that can be a separate topic if you prefer.

sophokles73:
I am ok with the approach of adding the additional registration attributes.
I also love the introduction of a `Client.Builder` :-) I don't think that we need a separate PR for it but maybe have a separate commit?

phtrivier:
@sophokles73 I renamed the map, and made it unmodifiable. 
I'd rather keep everything in a single commit ( my git-fu is not strong enough to separate the past commits without breaking or having to rewrite stuff, and @sbernard31  will be squashing things anyway :) )

sophokles73:
I see, no sweat.
@sbernard31: +1 for merging then :-)

leshan-server-core/src/main/java/org/eclipse/leshan/server/client/Client.java
```diff
@@ -62,27 +65,21 @@
 
     private final LinkObject[] objectLinks;
 
+    private Map<String, String> additionalRegistrationAttributes = new HashMap<String, String>();
+
```
sbernard31: should be final

leshan-server-core/src/main/java/org/eclipse/leshan/server/client/Client.java
```diff
@@ -275,6 +275,14 @@ public boolean isAlive() {
         return lastUpdate.getTime() + lifeTimeInSec * 1000 > System.currentTimeMillis();
     }
 
+    public Map<String, String> getAdditionalRegistrationParams() {
+        return additionalRegistrationAttributes;
+    }
+
+    public void setAdditionalRegistrationParams(Map<String, String> additionalRegistrationParams) {
```
sbernard31: no setter (client is immutable)

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationHandler.java
```diff
@@ -62,10 +63,16 @@ public RegisterResponse register(Identity sender, RegisterRequest registerReques
...
 
-        Client client = new Client(RegistrationHandler.createRegistrationId(), registerRequest.getEndpointName(),
-                sender.getPeerAddress().getAddress(), sender.getPeerAddress().getPort(),
-                registerRequest.getLwVersion(), registerRequest.getLifetime(), registerRequest.getSmsNumber(),
-                registerRequest.getBindingMode(), registerRequest.getObjectLinks(), serverEndpoint);
+        Client.Builder builder = new Client.Builder(RegistrationHandler.createRegistrationId(),
+                registerRequest.getEndpointName(), sender.getPeerAddress().getAddress(), sender.getPeerAddress()
+                        .getPort(), serverEndpoint);
+
+        builder = builder.lwm2mVersion(registerRequest.getLwVersion()).lifeTimeInSec(registerRequest.getLifetime())
```
sbernard31: builder = ...  is not needed
(our formatter does not really like inline syntax :p)

leshan-server-core/src/main/java/org/eclipse/leshan/server/client/Client.java
```diff
@@ -62,27 +65,21 @@
...
     private final String rootPath;
 
     private final Date lastUpdate;
 
-    public Client(String registrationId, String endpoint, InetAddress address, int port,
-            InetSocketAddress registrationEndpointAddress) {
-        this(registrationId, endpoint, address, port, null, null, null, null, null, registrationEndpointAddress);
-    }
-
+    @Deprecated
```
sbernard31: As we break the API (by adding parameters), I think we can remove the deprecated. 
If we want to force user to use Builder, we can make the constructor protected.

sbernard31:
Integrated in master (commit a3eb59dd782da8d6fe6deb04155e89c15acd9d3d)

PR #68, opened 2016-02-04
Enhance Demo

sbernard31:
I have done some modification to make our demos CLI more intuitive.


sbernard31:
Last change I would like to do, rename :
 * `leshan-standalone` in `leshan-server-demo`
 * `leshan-client-example` in `leshan-client-demo`
 * `leshan-bs-server` in `leshan-bsserver-demo`
What do you think ?

sophokles73:
Hi @sbernard31,

can you just shortly explain why so many methods now explicitly throw `InterruptedException`? The reason for this is not obvious to me from flying over the code and I want to understand what made this necessary because `InterruptedException` is an exception I do not see explicitly thrown that often ...

sbernard31:
Ok :)
There is not so many changes, I mainly change the way interrupted exception is handled when we send request in synchronous way. Most throwing exception you see are in the tests, the real change is the API  on LwM2mServer.java)

Before this change, an `InterruptedException` was handled as a timeout (I mean the `send` method return `null`), but we see in practice this is not really a good idea. You don't really want to do the same thing if the thread is interrupted or if the request timeout.

So we choose to throw `InterruptedException` as it is the common way to handle thread interruption on blocking method in Java.
(If you use the asynchronous way to send request, you don't have `InterruptedException` but in our tests  we mainly use the synchronous way)

sophokles73:
I see, thanks for the explanation, @sbernard31 :-)
+1 for renaming (this should help making clear that these implementations are not intended for production deployment)
+1 for merging

msangoi:
:+1: for merging

PR #71, opened 2016-02-16
Fixed Security check for non-secure case

phtrivier:
Should fix NPE.
Also added integration test for secure cases.

Signed-off-by: Pierre-Henri Trivier <phtrivier@sierrawireless.com>


leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityCheck.java
```diff
@@ -38,26 +38,26 @@
...
+                return false;
+            } else {
+                for (SecurityInfo securityInfo : securityInfos) {
+                    if (checkSecurityInfo(endpoint, clientIdentity, securityInfo)) {
+                        return true;
+                    }
+                }
+                return false;
+            }
+        } else {
```
jvermillard: } else if() { ?

PR #76, opened 2016-02-18
SecurityRegistryImpl file persistance refactoring

oflisback:
This is a refactorization suggestion for SecurityRegistryImpl file persistance which looked quite wonky. If the patch has potential, please suggest approach for testing it. Currently the file persistance part of SecurityRegistryImpl isn't covered by the integration tests.

Specifically the previous implementation writes to the file while in a loop reading objects from it (loadFromFile did add which did saveToFile), it also had some funky tripple-nested exception handling.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/SecurityRegistryImpl.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/SecurityRegistryImpl.java
@@ -18,6 +18,7 @@ package org.eclipse.leshan.server.impl;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
+import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.security.PrivateKey;
@@ -26,8 +27,8 @@ import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.leshan.server.security.NonUniqueSecurityInfoException;
 import org.eclipse.leshan.server.security.SecurityInfo;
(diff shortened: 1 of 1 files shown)


jvermillard:
Agree on the wonkyness, now why keep "concurrent hashmap" if we use synchronize ?
I would prefer two "normal" maps and a ReadWrite lock. WDYT?

This class is not tested because it's used for running leshan-demo, it's not mean to be used in production (at least we are not for AirVantage), but the location (server-core) and the doc is not explicit.

So yes, let's fix it.

sbernard31:
I agree if we use synchronize we don't need concurrent hashmap anymore.

I'm ok that load/save method are a bit strange, I rewrote it for the [BootstrapStoreImpl](https://github.com/eclipse/leshan/blob/master/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapStoreImpl.java#L87), but I don't take time to apply this change to `SecurityRegistryImpl`.

Adding `SecurityInfo[]` and `filename` to load and save method seems not really necessary? what is the purpose ?
You missed to close the `FileInputStream` in `loadFromFile`.

oflisback:
Thanks a lot for your feedback, I have pushed an update based on your comments. I reverted the changes to the saveToFile/loadFromFile method signatures too, I prefer input parameters but I agree it wasn't a necessary change.

sbernard31:
Currently, the file creation is done on load. I think we should do it on save, like we do [here](https://github.com/eclipse/leshan/blob/master/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapStoreImpl.java#L87). We can copy/paste a large part of the save/load method.

oflisback:
Agreed, actually this is how it's done in the current patch. :)

sbernard31:
you're right ^^ !
I integrated you patch in master (commit 4a00f36ec47ef7d5acdb3a110b836cfea680aa0e).
I just change a log.info to log.debug.
And merge all the catching exception in one clause.
Thx @oflisback

oflisback:
Thanks guys!

Issue #75, opened 2016-02-18
Client: cannot create an object instance without instance Id


msangoi:
Following commit d16c6b3dfba0e84faedd3413593a344b3fbe1041, a regression was introduced in the client side. 
When an instance is created for a multi-instance object (e.g. Software Update, ACL), the TLV payload failed to be decoded because the instance Id is not provided.

I will provide a fix for this.

jvermillard:
![](https://media.giphy.com/media/ZOhNKz99I4TwA/giphy.gif)

msangoi:
Actually, the instance ID is not anymore a parameter of the Create operation (spec change).
We need to change the Create Request API (impacting both server and client).

sbernard31:
Fixed in master (commit a4bf1b3fcb5429d514c0ddf5c425aae5ae790903)

PR #83, opened 2016-02-23
Remove instance Id from Create Request path

msangoi:
Following a specification change, the instance Id is not part the the Create Request path anymore.
The request payload contains an object instance or list of resources (if the client is in charge of picking the new instance Id).

Encoders/decoders were updated to be able to encode/decode an object instance as an instance or as a array of resources. Some work is stil to be done for JSON format.

This PR fixes the related issue on the client-side (#75).


sbernard31:
sounds good :+1:

sbernard31:
Integrated in master (commit  841538266dfe3f0196d02b8b337831b6245513cb  and a4bf1b3fcb5429d514c0ddf5c425aae5ae790903)

PR #84, opened 2016-02-23
use read/write lock for less read contention

jvermillard:
@oflisback I changed your last PR for using readwritelock in place of synchronized map.
It make read able to be processed in parallel and not serialized


oflisback:
looking good!

PR #89, opened 2016-03-03
Added Redis based stores (security & clients)

jvermillard:
CQ under review for jedis + common-pool2
http://dev.eclipse.org/ipzilla/show_bug.cgi?id=11027
http://dev.eclipse.org/ipzilla/show_bug.cgi?id=11029
so don't merge before we get the go :)

However I would appreciate feedback

### DIFF
--- a/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/cluster/RedisClientRegistry.java
+++ b/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/cluster/RedisClientRegistry.java
@@ -1,0 +1,232 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Sierra Wireless and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *     Sierra Wireless - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.leshan.server.demo.cluster;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutput;
+import java.io.ObjectOutputStream;
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import org.eclipse.californium.scandium.util.ByteArrayUtils;
+import org.eclipse.leshan.server.client.Client;
+import org.eclipse.leshan.server.client.ClientRegistry;
+import org.eclipse.leshan.server.client.ClientRegistryListener;
+import org.eclipse.leshan.server.client.ClientUpdate;
+import org.eclipse.leshan.util.Validate;
+
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.ScanParams;
+import redis.clients.jedis.ScanResult;
+import redis.clients.util.Pool;
+
+/**
+ * A client registry storing registration in Redis.
+ * 
+ * the main key is the client end-point and the registration key is used as a secondary index
+ * 
+ * TODO
+ * 
+ * - receive key expiration event (see http://redis.io/topics/notifications)
+ */
+public class RedisClientRegistry implements ClientRegistry {
+
+    private static final String EP_CLIENT = "EP#CLIENT#";
+
+    private static final String REG_EP = "REG#EP#";
+
+    private final Pool<Jedis> pool;
+
+    private final List<ClientRegistryListener> listeners = new CopyOnWriteArrayList<>();
+
+    public RedisClientRegistry(Pool<Jedis> pool) {
+        this.pool = pool;
+    }
+
+    @Override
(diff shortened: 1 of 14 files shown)


leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/cluster/RedisClientRegistry.java
```diff
@@ -0,0 +1,232 @@
...
+            byte[] k = (EP_CLIENT + client.getEndpoint()).getBytes();
+            byte[] old = j.getSet(k, data);
+
+            j.expire(k, client.getLifeTimeInSec().intValue());
+
+            // secondary index
+            byte[] idx = (REG_EP + client.getRegistrationId()).getBytes();
+            j.set(idx, client.getEndpoint().getBytes());
+
+            j.expire(idx, client.getLifeTimeInSec().intValue());
```
msangoi: No call to ClientRegistryListener.unregistered() when the lifetime is expired.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/cluster/RedisClientRegistry.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+    }
+
+    @Override
+    public Client deregisterClient(String registrationId) {
+        try (Jedis j = pool.getResource()) {
+            // first rename for atomicity
+            if (!"OK".equals(j.rename((REG_EP + registrationId).getBytes(),
+                    ("TODELETE#" + REG_EP + registrationId).getBytes()))) {
+                System.err.println("del meh");
```
msangoi: Syserr meh

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/cluster/RedisClientRegistry.java
```diff
@@ -0,0 +1,232 @@
...
+            byte[] key = j.get(("TODELETE#" + REG_EP + registrationId).getBytes());
+            if (key == null) {
+                return null;
+            }
+            byte[] data = j.get(ByteArrayUtils.concatenate(EP_CLIENT.getBytes(), key));
+
+            Client c = deserialize(data);
+
+            // delete everything
+            j.del(("TODELETE#" + REG_EP + registrationId).getBytes());
```
msangoi: `("TODELETE#" + REG_EP + registrationId).getBytes() ` -> This key is built 3 times in this method.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/cluster/RedisClientRegistry.java
```diff
@@ -0,0 +1,232 @@
...
+            j.del(ByteArrayUtils.concatenate(EP_CLIENT.getBytes(), key));
```
msangoi: This key construction is duplicated too.

sophokles73:
This looks good, @jvermillard :-)
My only concern would be using standard Java serializaiton for marshalling/unmarshalling the objects. The Serializable contract is not easy to implement (and keep for that matter) as e.g. discussed in chapter 11 of Joshua Bloch's *Effective Java*. The arguments Joshua brings forward may not be applicable here (mainly due to the short-lived-ness of the cached objects) but I think it would still be wise to reflect on this a little more and maybe use another (custom) serialization format or use e.g. Jackson to marshall to JSON instead.

jvermillard:
yes java serialization is really for the PoC.
My next move on tha would be to use https://capnproto.org/ 
or to split the objects in redis HSET fields for having the capability to patch them on updates

sophokles73:
The latter sounds reasonable to me. Do you think it is worth the effort of defining a schema and compiling type classes using something like capnproto for the few times we actually serialize, deserialize the objects?

jvermillard:
Now with json serialisation

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/cluster/RedisClientRegistry.java
jvermillard: see the TODO is javadoc comment, the plan is to receive expiration from redis

jvermillard:
waiting for CQ then it can be merged (if no other comments)

ingoschaal:
I try to use the new Redis based stores with a redis cloud service for example RedisCloud of pivotal. This service requires authentication.
There are no options, to enter credentials (password). 
Thereby I get this exception redis.clients.jedis.exceptions.JedisDataException: NOAUTH Authentication required.
The Jedis Object provides a mehthode auth(Stirng password) with which it works
Is it provided that it comes with authentication?

sophokles73:
@ingoschaal 

looks like a good opportunity for you to create a PR adding support for authentication ;-) But you should wait until Julien has merged this PR ...

jvermillard:
you can PR on this branch, we are waiting to have observation working for merging, so we wait for https://github.com/eclipse/californium/pull/16 to be done and released

ingoschaal:
Ok, good suggestion!
Another question:
Is it provided that the Redis based implementation of the Clientregistry remains in the demo server or should this be included in the core?
In my opinion, the core is the better place. Maybee in future ?

jvermillard:
IMO the future is to make demo a real deployable server, and keep core as a pure LWM2M protocol library but we can discuss that :) my plan was to open this topic once leshan-demo is working as a cluster and so more than a simple demo server but something nearly deployable out of the box

sbernard31:
We decide to merge #93  and #89 in #113.

Fakhruddin-Kararawala:
@jvermillard 
how can we observe a resource with lt,gt,pmin,pmax parameter as threshold.
what we are trying is that the device can send us notification only when these threshold values exceeds.

PR #92, opened 2016-03-07
Client registry listener API change

msangoi:
Add the ClientUpdate parameter when a client registration is updated


sbernard31:
Integrated in master (commit 5f7d1491597dd244d39e90950e3b875d1aab568f)

PR #94, opened 2016-03-09
Implement POST "/oid/iid" as UPDATE instance

boaks:
Implement POST "/oid/iid" as UPDATE instance according
OMA-DM-LightweightM2M-2015-0054R02-CR_Commands_Definition_Conflicts.

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/ObjectResource.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/ObjectResource.java
@@ -14,6 +14,8 @@
  *     Sierra Wireless - initial API and implementation
  *     Achim Kraus (Bosch Software Innovations GmbH) - use ObserveRelationFilter
  *     Achim Kraus (Bosch Software Innovations GmbH) - use ServerIdentity
+ *     Achim Kraus (Bosch Software Innovations GmbH) - implement POST "/oid/iid" 
+ *                                                     as UPDATE instance
  *******************************************************************************/
 package org.eclipse.leshan.client.californium.impl;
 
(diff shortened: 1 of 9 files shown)


sbernard31:
Sounds good. 
We could also add some tests in `leshan-integration-tests`.
if we want to handle this correctly we should also add the code which manage `UPDATE` and `REPLACE` in [`ObjectEnabler`](https://github.com/eclipse/leshan/blob/master/leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/ObjectEnabler.java#L145)

boaks:
OK, I will add test (and remove whitespace :-) .

boaks:
An other alternative for removing not longer contained resources in REPLACE may be the introduction of a "onInitialize" method in `BaseObjectEnabler`.

sbernard31:
Tests is ok for me.

About `REPLACE`/`UPDATE`, I should have chosen the `DELETE`/`CREATE`.
I mean intuitively, It seems that `REPLACE` means delete the instance and replace it by this new one. (so this seems to be a creation)
With your solution we keep the previous instance.

The spec is really don't clear about this:
*" - Replace: replaces the Object Instance or the Resource(s) with the new value provided in the “Write” operation."*

But maybe you're right, your solution avoid some troubles:
- e.g. On `Server` Object a `delete`/`create` means that we will lost the shortID
- e.g. With mandatory single instance a `delete`/`create` will be problematic as a delete is forbidden (For this one we can choose to use the SYSTEM identity ...)

**If someone else has an idea about that ?**

In all case, if we chose your solution, I prefer we avoid the `write(null)` stuff and use the `DeleteRequest` on the resource.

boaks:
I have to confess, that currently my "Objects Instances" are more "wired" and it would be very hard to change that right now to `delete` / `create` (I know, this means also no bootstrap :-) ).

I will go for a `Delete` on a resource

boaks:
(Forget this ...

Why? Delete ("/o/i/r") works, may be a more radical then expected :-).

boaks:
Using DELETE on resource level only for the SYSTEM and not for DM SERVER makes a dependecy to the PR #95.
It would be cleaner to use the newly introduce ExtendedIdentity.

boaks:
After reanalysing, I choose the name `reset()` for cleanup the none mandatory writable resources in case of write / UPDATE.

For a more "far solution" I started thinking about something with
- `startWrite(update/replace)`
- `finishWrite(success/rollback)`
- `onDelete()`
on `InstanceEnabler` base.
This introduce the possibility to support 
- "atomic instance writes" (with rollback)
- collect changes for multi resources changes caused by instance writes
- may protect reads for changing instances.
- release resource, when the instance is deleted.

The issue for such a "far solution" would be, that it should have a defined
"Threading Model" (see Californium `Resource.getExecutor()` ). Otherwise
I would guess, that it may get to complicated.

sbernard31:
Integrated in master (commit 37fb24f499269fcff4a7744407f5e7d4d343fd97)
Thx again @boaks.

boaks:
WARNING:
The REPLACE is not proper working for Bootstrap!
It doesn't write the server ID in the server instance ...
I'm going to fix this.

jvermillard:
:scream:

boaks:
Yes, but usually a BS will first delete all security and server object instances.
So the REPLACE is a rare case for BS.

PR #95, opened 2016-03-09
Protect security object from invalid access.

boaks:
Protect security object from access of DM server.
Protection is implemented in CoAP method mapping to allow the internal use of read/write operations as in ServersInfoExtractor.

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/BootstrapResource.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/BootstrapResource.java
@@ -12,15 +12,16 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
+ *     Achim Kraus (Bosch Software Innovations GmbH) - use ExtendedIdentity
  *******************************************************************************/
 package org.eclipse.leshan.client.californium.impl;
 
 import org.eclipse.californium.core.CoapResource;
 import org.eclipse.californium.core.server.resources.CoapExchange;
 import org.eclipse.californium.core.server.resources.Resource;
+import org.eclipse.leshan.client.request.ExtendedIdentity;
 import org.eclipse.leshan.client.servers.BootstrapHandler;
 import org.eclipse.leshan.core.request.BootstrapFinishRequest;
-import org.eclipse.leshan.core.request.Identity;
 import org.eclipse.leshan.core.response.BootstrapFinishResponse;
 
 /**
@@ -37,7 +38,7 @@ public class BootstrapResource extends CoapResource {
 
     @Override
     public void handlePOST(CoapExchange exchange) {
-        Identity identity = ResourceUtil.extractIdentity(exchange);
+        ExtendedIdentity identity = ResourceUtil.extractIdentity(exchange, bootstrapHandler);
         BootstrapFinishResponse response = bootstrapHandler.finished(identity, new BootstrapFinishRequest());
         exchange.respond(ResourceUtil.fromLwM2mCode(response.getCode()), response.getErrorMessage());
     }
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/BootstrapResource.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/BootstrapResource.java
@@ -12,14 +12,14 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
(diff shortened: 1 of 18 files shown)


sbernard31:
Ideally, the responsibility of `ObjectResource` should be to translate CoAP request in LWM2M request. 
The idea is to keep a minimum of LWM2M logic in `leshan-server-cf` which is dependent of `californium`.

But to do that we should move "identity checks"  in `BaseObjectEnabler`. This means that it needs access to `BootstrapHandler`  :-/

About return codes, maybe `4.01 Unauthorized` is more adapted, what do you think ?

boaks:
The issue for me was:
- read is never allowed from outside
- but used from inside (`ServersInfoExtractor`).

So solving it with "identity check" would require to introduce a "SYSTEM" identity to allow such internal operations. And that seems to me a little "overdone".
Therefore I chose the `ObjectResource` also as point of "external use", leave the `BaseObjectEnabler` for easy internal use.

sbernard31:
We already face this kind of problem, currently we consider [null identity as a "SYSTEM" identity](https://github.com/eclipse/leshan/blob/master/leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/ObjectEnabler.java#L133).

boaks:
About return codes, maybe 4.01 Unauthorized is more adapted, what do you think ? 

I chose 4.04 Not Found, because, if the object is not reported in the registration, its not there :-).

boaks:
We already face this kind of problem, currently we consider null identity as a "SYSTEM" identity.

OK, this makes it easier ...

sbernard31:
Ok for 4.04 Not Found.

boaks:
We already face this kind of problem, currently we consider null identity as a "SYSTEM" identity.

Ok, but

    public static Identity extractIdentity(CoapExchange exchange) {
        InetSocketAddress peerAddress = new InetSocketAddress(exchange.getSourceAddress(), exchange.getSourcePort());

        Principal senderIdentity = exchange.advanced().getRequest().getSenderIdentity();
        if (senderIdentity != null) {
            if (senderIdentity instanceof PreSharedKeyIdentity) {
                return Identity.psk(peerAddress, senderIdentity.getName());
            } else if (senderIdentity instanceof RawPublicKeyIdentity) {
            } else if (senderIdentity instanceof X500Principal) {
                // Extract common name
                Matcher endpointMatcher = Pattern.compile("CN=.*?,").matcher(senderIdentity.getName());
                if (endpointMatcher.find()) {
                    String x509CommonName = endpointMatcher.group().substring(3, endpointMatcher.group().length() - 1);
                } else {
                    return null;
                }
            }
        }
        return Identity.unsecure(peerAddress);
    }

returns null ...

Even fixing the above, null may be accidently introduced by someone somewhere. 
So, I think, we should Introduce a SYSTEM Identity ...

sbernard31:
We can do that, this is probably cleaner.
You think about something like that ?
```java
public class Identity {
    public static final Identity SYSTEM = new Identity();
    ....
```

boaks:
I introduced a ExtendedIdentity for easier distinguish between `SYSTEM,` `BOOTSTRAP_SERVER` and `DM_SERVER.`
I'm not sure, if the server works well with the return null Identity. But though I don't wan't to change this also, I chose the client side extension of the Identity.

boaks:
I will reformat it ...

boaks:
So, no I hope, the format is proper ...

(By the way, I'm back on tuesday :-)  )

boaks:
Sorry, I killed you comments ;-).

Yes, a warning may be added (I hope, this comment was at the "missing common name in x509")

I chose `ExtendedIdentity,` because `SYSTEM` is no "Server"

boaks:
Warning added, but now its late for me.
So tuesday ...

sbernard31:
No problem ;)
(I continue to comment the PR)

sbernard31:
Why do you want to rebase ?

boaks:
Because the master was updated?
Or do you just wait, until conflicts are reported?

sbernard31:
I propose to not rebase, when we will be ok about the PR. We will resolve conflicts later.

sbernard31:
Sounds Good (Except ExtendedIdentity name  ^^ )

boaks:
I already renamed it.
I will push soon (this afternoon).

boaks:
If it's OK, I would rebase?

sbernard31:
I did it. 
This is integrated in master (commit c5f03e19ba3760d7a5480cf9c2f12daa8c78ddc7)
Thx @boaks  ;)

Issue #91, opened 2016-03-05
Query: Security object not deleted on bootstrap


kiranpradeep:
Is this [`DeleteRequest`](https://github.com/eclipse/leshan/blob/8066e9bc1c2b0eac1980eaf9a28c70d2304c8ed3/leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/BootstrapHandler.java#L87) on `SERVER` object intentional ? Feels it should have been on `SECURITY` object. Shall I go with a pull request ?

jvermillard:
looks like a copy/paste error to me. :+1: for the PR ;)

sbernard31:
Fixed in master (commit 1056533fa8f4d9d7048dd8f65e395844cf2947a1)

PR #99, opened 2016-03-13
Delete security object on bootstrap completion

kiranpradeep:
Fix for issue #91. Security object was not getting deleted on bootstrap completion.
Signed-off-by: Kiran Pradeep <kiran.happy@gmail.com>

### DIFF
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/BootstrapHandler.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/BootstrapHandler.java
@@ -83,7 +83,7 @@ public class BootstrapHandler {
             // Delete all security instance (except bootstrap one)
             // TODO do not delete boostrap server (see 5.2.5.2 Bootstrap Delete)
             LwM2mObjectEnabler securityObject = objects.get(SECURITY);
-            for (Integer instanceId : serverObject.getAvailableInstanceIds()) {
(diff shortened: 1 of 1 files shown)


kiranpradeep:
1. About the `//TODO: don't delete bootstrap server` line in `BootstrapHandler::delete`, shall I do a ReadRequest on the retrieved securityObject and skip deletion if it is a bootstrap server as in `SecurityObjectPskStore::getKey` ? Missed any other point ?
2. Why `BaseObjectEnabler::doDelete` returns internalServerError instead of a real deletion?

sbernard31:
1. The bootstrap server should not be deleted while a bootstrap session, but should be done at "boostrap finished".  See the other [`TODO`](https://github.com/eclipse/leshan/blob/master/leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/BootstrapHandler.java#L57). Both `TODO` should be done in the same PR ;).
2. The   `BaseObjectEnable.doDelete()` returns `internalServerError("not implemented")` because this is an "abstract" class and it is not able to do the real deletion. This is the responsibility of concrete classes. (like [ObjectEnabler](https://github.com/eclipse/leshan/blob/master/leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/ObjectEnabler.java#L218))

sbernard31:
Integrated in master (commit 1056533fa8f4d9d7048dd8f65e395844cf2947a1)

Issue #77, opened 2016-02-19
Race condition with observation on client update


vvalchev:
We have the following situation:
1. there is a client with endpoint 'endpoint' and registration ID "reg1".
2. we've provided an ObservationRegistryListener
3. we send observe request to a certain instance
4. simultaneously the client sends 'update' request to our server changing it's registration id to 'reg2'.

From that point, the ObservationListener.newValue() receives new values but the observation is with registration id 'reg1', while it should be 'reg2'.

Because of the invalid registration ID I'm also unable to find the client it is coming from. Therefore I suggest, that the Observation should contain the endpoint ID. Or, at least Leshan should update the existing Observations with the new registration ID, on ClientRegistryListener.updated(Client) event is received.

sbernard31:
If I well understand, you say that an `UPDATE` change the registration ID ?
I think it's not allowed by the spec..

vvalchev:
Hmm, good point! I should check the specification, and if you are right - then the problem is in the device I use. Thank you.

p-koev:
I would just like to clarify, that this occurs when the device is restarted, not on update. It sends a new registration request with the same endpoint name and address, so it is assigned a new registration id. However the observation remains with the old registration id, and it notifies the ObservationListeners for value changes using that expired id.

sbernard31:
Ok, this use case is no so clear in the spec. An [issue](https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/71) is open at the OMA
As I understand the spec, the observation lifetime is linked to a registration. 

When you said the registration remain you mean at client or server side ?
Which client implementation did you use ? Leshan client ?

p-koev:
Real device connected to Leshan server. The problem arises because the observer request is triggered after the device goes offline, but before the lifetime interval has expired. So it is still present in the client registry. Then when the device quickly comes back online it receives a new registration Id, and also the observation request. Thus an observation is created both on the client and server, but the one in Leshan server still has the old registration Id.

sbernard31:
When Leshan receive a new registration request, it cancels all observations of the previous registration.
An observation is linked to a registration so, it's normal that the old observation has the old registration id. But it should have a new observation linked to the new registration (if you resent the request)

Did you use Leshan server demo (previously named standalone) or did you use Leshan as a lib in your server ?
In this case, did you have custom implementation ? like a custom ClientRegistry for example ?

p-koev:
[...]

PR #100, opened 2016-03-15
Cancel pending request on de-registration

sbernard31:
(This should fix the #77 )

### DIFF
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
@@ -20,7 +20,9 @@ import static org.eclipse.leshan.integration.tests.IntegrationTestHelper.ENDPOIN
 import static org.eclipse.leshan.integration.tests.IntegrationTestHelper.LIFETIME;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 import java.net.InetAddress;
@@ -36,7 +38,9 @@ import org.eclipse.leshan.client.californium.impl.CaliforniumLwM2mClientRequestS
 import org.eclipse.leshan.client.resource.LwM2mInstanceEnabler;
 import org.eclipse.leshan.client.resource.LwM2mObjectEnabler;
 import org.eclipse.leshan.client.resource.ObjectEnabler;
+import org.eclipse.leshan.core.request.ReadRequest;
 import org.eclipse.leshan.core.request.RegisterRequest;
+import org.eclipse.leshan.core.response.ReadResponse;
 import org.eclipse.leshan.server.client.Client;
 import org.eclipse.leshan.server.security.NonUniqueSecurityInfoException;
 import org.junit.After;
@@ -83,6 +87,38 @@ public class RegistrationTest {
         helper.client.stop(true);
         helper.waitForDeregistration(1);
         assertTrue(helper.server.getClientRegistry().allClients().isEmpty());
+    }
+
+    @Test
+    public void deregister_cancel_pending_request() throws InterruptedException {
+        // Check registration
+        assertTrue(helper.server.getClientRegistry().allClients().isEmpty());
+
+        helper.client.start();
+        helper.waitForRegistration(1);
+
+        assertEquals(1, helper.server.getClientRegistry().allClients().size());
+        Client client = helper.server.getClientRegistry().get(ENDPOINT_IDENTIFIER);
+        assertNotNull(client);
+        assertArrayEquals(LinkObject.parse("</>;rt=\"oma.lwm2m\",</1/0>,</2>,</3/0>".getBytes()),
+                client.getObjectLinks());
+
+        // Stop client with out de-registration
+        helper.client.stop(false);
+
+        // Send a read which should be retransmitted.
+        Callback<ReadResponse> callback = new Callback<ReadResponse>();
+        helper.server.send(client, new ReadRequest(3, 0, 1), callback, callback);
+
+        // Restart client (de-registration/re-registration)
+        helper.client.start();
+
+        // Check the request was cancelled.
+        boolean timedout = !callback.waitForResponse(1000);
+        assertFalse("Response or Error expected", timedout);
+        assertTrue("Response or Error expected", callback.isCalled().get());
+        assertNull("No response expected", callback.getResponse());
+        assertNotNull("Exception expected", callback.getException());
     }
 
     @Test
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
@@ -29,6 +29,7 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.eclipse.californium.core.network.Endpoint;
@@ -119,6 +120,47 @@ public class RegistrationTest {
         assertTrue("Response or Error expected", callback.isCalled().get());
         assertNull("No response expected", callback.getResponse());
         assertNotNull("Exception expected", callback.getException());
+    }
+
+    @Test
+    public void deregister_cancel_multiple_pending_request() throws InterruptedException {
+        // Check registration
+        assertTrue(helper.server.getClientRegistry().allClients().isEmpty());
+
+        helper.client.start();
+        helper.waitForRegistration(1);
+
+        assertEquals(1, helper.server.getClientRegistry().allClients().size());
+        Client client = helper.server.getClientRegistry().get(ENDPOINT_IDENTIFIER);
+        assertNotNull(client);
+        assertArrayEquals(LinkObject.parse("</>;rt=\"oma.lwm2m\",</1/0>,</2>,</3/0>".getBytes()),
+                client.getObjectLinks());
+
+        // Stop client with out de-registration
+        helper.client.stop(false);
+
+        // Send multiple reads which should be retransmitted.
+        List<Callback<ReadResponse>> callbacks = new ArrayList<Callback<ReadResponse>>();
+
+        for (int index = 0; index < 4; ++index) {
+            Callback<ReadResponse> callback = new Callback<ReadResponse>();
+            helper.server.send(client, new ReadRequest(3, 0, 1), callback, callback);
+            callbacks.add(callback);
+        }
+
+        // Restart client (de-registration/re-registration)
+        helper.client.start();
+
+        // Check the request was cancelled.
+        int index = 0;
+        for (Callback<ReadResponse> callback : callbacks) {
+            boolean timedout = !callback.waitForResponse(1000);
+            assertFalse("Response or Error expected, no timeout, call " + index, timedout);
+            assertTrue("Response or Error expected, call " + index, callback.isCalled().get());
+            assertNull("No response expected, call " + index, callback.getResponse());
+            assertNotNull("Exception expected, call " + index, callback.getException());
+            ++index;
+        }
     }
 
     @Test
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
@@ -17,11 +17,15 @@ package org.eclipse.leshan.server.californium.impl;
 
 import java.net.InetSocketAddress;
 import java.util.Set;
+import java.util.SortedMap;
+import java.util.concurrent.ConcurrentNavigableMap;
+import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.eclipse.californium.core.coap.MessageObserver;
 import org.eclipse.californium.core.coap.MessageObserverAdapter;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
@@ -47,6 +51,9 @@ public class CaliforniumLwM2mRequestSender implements LwM2mRequestSender {
     private final Set<Endpoint> endpoints;
     private final ObservationRegistry observationRegistry;
     private final LwM2mModelProvider modelProvider;
+    // A map which contains all pending CoAP requests
+    // This is mainly used to cancel request and avoid retransmission on de-registration
+    private final ConcurrentNavigableMap<String/* registrationId#requestId */, Request /* pending coap Request */> pendingRequests = new ConcurrentSkipListMap<>();
 
     /**
      * @param endpoints the CoAP endpoints to use for sending requests
@@ -76,7 +83,8 @@ public class CaliforniumLwM2mRequestSender implements LwM2mRequestSender {
         final Request coapRequest = coapRequestBuilder.getRequest();
 
         // Send CoAP request synchronously
-        final SyncRequestObserver<T> syncMessageObserver = new SyncRequestObserver<T>(coapRequest, destination, timeout) {
+        final SyncRequestObserver<T> syncMessageObserver = new SyncRequestObserver<T>(coapRequest, destination,
+                timeout) {
             @Override
             public T buildResponse(final Response coapResponse) {
                 // Build LwM2m response
@@ -87,6 +95,9 @@ public class CaliforniumLwM2mRequestSender implements LwM2mRequestSender {
             }
         };
         coapRequest.addMessageObserver(syncMessageObserver);
+
+        // Store pending request to cancel it on deregistration
+        addPendingRequest(destination.getRegistrationId(), coapRequest);
 
         // Send CoAP request asynchronously
         final Endpoint endpoint = getEndpointForClient(destination);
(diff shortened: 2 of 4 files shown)


boaks:
Would it be possible to put this change on an "experimental" branch?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -108,23 +117,102 @@ public T buildResponse(final Response coapResponse) {
...
         final Endpoint endpoint = getEndpointForClient(destination);
         endpoint.sendRequest(coapRequest);
     }
 
+    public synchronized void cancelPendingRequests(String registrationId) {
+        ArrayList<Request> requests = pendingRequests.get(registrationId);
+        if (requests != null) {
+            for (Request coapRequest : requests) {
+                coapRequest.cancel();
+                removePendingRequest(registrationId, coapRequest);
```
boaks: Did you check, if it's OK to remove something from a collection while "iterating" over it?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -108,23 +117,102 @@ public T buildResponse(final Response coapResponse) {
...
         // Send CoAP request asynchronously
         final Endpoint endpoint = getEndpointForClient(destination);
         endpoint.sendRequest(coapRequest);
     }
 
+    public synchronized void cancelPendingRequests(String registrationId) {
+        ArrayList<Request> requests = pendingRequests.get(registrationId);
+        if (requests != null) {
+            for (Request coapRequest : requests) {
+                coapRequest.cancel();
```
boaks: Is it really intended to call "thirdparty" stuff within a "synchronized" section?
Even, if at least logicially, this "calls back" into the same synchronized section?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -108,23 +117,102 @@ public T buildResponse(final Response coapResponse) {
...
+
+        // Store pending request to cancel it on deregistration
+        addPendingRequest(destination.getRegistrationId(), coapRequest);
 
         // Send CoAP request asynchronously
         final Endpoint endpoint = getEndpointForClient(destination);
         endpoint.sendRequest(coapRequest);
     }
 
+    public synchronized void cancelPendingRequests(String registrationId) {
```
boaks: Does it scale to introduce a "server global synchronized" section?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -47,6 +51,7 @@
     private final Set<Endpoint> endpoints;
     private final ObservationRegistry observationRegistry;
     private final LwM2mModelProvider modelProvider;
+    private final Map<String/* registration id */, ArrayList<Request> /* pending coap Request */> pendingRequests = new HashMap<>();
```
boaks: I'm not sure, if an additional map with key <reg id> is really required.
Ok, it "isolates the concerns" but extending the "Client" may be a alternative, which is also not too bad.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -108,23 +117,102 @@ public T buildResponse(final Response coapResponse) {
...
+            // TODO remove this test when we will integrate the new observation registry implementation.
+            if (coapRequest.getOptions() == null
+                    || !(coapRequest.getOptions().hasObserve() && coapRequest.getOptions().getObserve() == 0)) {
+                removePendingRequest(registrationId, coapRequest);
+            }
+        }
+
+        @Override
+        public void onAcknowledgement() {
+            removePendingRequest(registrationId, coapRequest);
```
boaks: This will "fix/bind" the usagee of piggybacked reponses.
Though "separate responses" have also advantages, I'm not sure, if binding to hard to piggybacked is a good idea.

boaks:
Is "cancel request on deregister" only wanted for californium?
If not, shouldn't we introduce a "cancel" into the core?
(OK, I already have a experimental extension for the client side. I don't want to have to many PRs open, but just for a "impression" I will create one and comment it with "don't merge".)

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
```diff
@@ -86,6 +90,38 @@ public void register_update_deregister() {
...
+        helper.client.start();
+
+        // Check the request was cancelled.
+        boolean timedout = !callback.waitForResponse(1000);
+        assertFalse("Response or Error expected", timedout);
+        assertTrue("Response or Error expected", callback.isCalled().get());
+        assertNull("No response expected", callback.getResponse());
+        assertNotNull("Exception expected", callback.getException());
+    }
+
```
boaks: May be add a test with multiple reads will show, if the "remove while iterate" is working.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
sbernard31: The test I write is ok, but this is clearly a mistake. (I will fix this)

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
sbernard31: It should be fixed too.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
sbernard31: The map is shared between the server.
 What do you propose ?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
sbernard31: Not sure to understand.  You want to extend the `org.eclipse.leshan.server.client.Client` ?
I'm not sure this is a good idea as Client should be an immutable bean which should be easily serializable.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
sbernard31: The main idea was to avoid the retransmission. That's why I firstly chose to remove pending request on acknowledgement, but maybe this is better to just wait the response.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
boaks: @Test
    public void deregister_cancel_multiple_pending_request() throws InterruptedException {
        // Check registration
        assertTrue(helper.server.getClientRegistry().allClients().isEmpty());

        helper.client.start();
        helper.waitForRegistration(1);

        assertEquals(1, helper.server.getClientRegistry().allClients().size());
        Client client = helper.server.getClientRegistry().get(ENDPOINT_IDENTIFIER);
        assertNotNull(client);
        assertArrayEquals(LinkObject.parse("</>;rt=\"oma.lwm2m\",</1/0>,</2>,</3/0>".getBytes()),
                client.getObjectLinks());

        // Stop client with out de-registration
        helper.client.stop(false);

        // Send multiple reads which should be retransmitted.
        List<Callback<ReadResponse>> callbacks = new ArrayList<Callback<ReadResponse>>();

        for (int index = 0; index < 4; ++index) {
            Callback<ReadResponse> callback = new Callback<ReadResponse>();
            helper.server.send(client, new ReadRequest(3, 0, 1), callback, callback);
            callbacks.add(callback);
        }

        // Restart client (de-registration/re-registration)
        helper.client.start();

        // Check the request was cancelled.
        int index = 0;
        for (Callback<ReadResponse> callback : callbacks) {
            boolean timedout = !callback.waitForResponse(1000);
            assertFalse("Response or Error expected, no timeout, call " + index, timedout);
            assertTrue("Response or Error expected, call " + index, callback.isCalled().get());
            assertNull("No response expected, call " + index, callback.getResponse());
            assertNotNull("Exception expected, call " + index, callback.getException());
            ++index;
        }
    }

Debugging this test shows:
- remove during iterate throws an exception
- that exception is "silently ignored"

sbernard31:
> Would it be possible to put this change on an "experimental" branch?

This is already in a branch [cancel_request](https://github.com/eclipse/leshan/tree/cancel_request).

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
boaks: If the "pendingRequest" should not be "serialized", I think it's possible to exclude them.

May idea about this was very simple:
- the server already has a map with the <reg-id> as key. And the access to such a map is not well scaling.
- so just reuse that "already made access" by storing the stuff in the objects maintained by the first map.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
sbernard31: At my side, this test works too. (and I can not find any silenty ignored exception, but maybe I don't find at the good place)
But in all case, I agree with you we must not remove pending iteration using foreach.

sbernard31:
> Is "cancel request on deregister" only wanted for californium?
For now that's was the idea, as we don't have cancel on request.
> If not, shouldn't we introduce a "cancel" into the core?
We could but this is a more impacting work.
> (OK, I already have a experimental extension for the client side. I don't want to have to many PRs open, but just for a "impression" I will create one and comment it with "don't merge".)
Sounds good. We will be able to look at this.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
sbernard31: > so just reuse that "already made access" by storing the stuff in the objects maintained by the first map.
Which map ? (could you give the name of attribute and class of map you are talking about)

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
boaks: I only analysed the deregister_cancel_multiple_pending_request() and that shows the exception.

But though you agree, I guess the fix will make this test workign as well.

boaks:
Sounds good. We will be able to look at this.

See PR #102. The principle is very primitive. So it should be also possible for the server side.
I'm not sure about the "synchronous" calls and how they should be handled.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
boaks: Sure, "ClientRegistryImpl.clientsByEp".

boaks: OK, this is "endpoint name" to client, not "register id".
So staying with the "reg-id", the entry in the map should be removed on "cancelPendingRequests", (otherwise this may be something like a memory leak).
Alternatively, we use the endpoint name, copy the request list on cancel and iterate over the copy. I will make a purposal.

boaks: PR #103 
Is a approach using concurrent collections instead of synchronization staying with the "reg-id" and is close to the original fix.

sbernard31: The `clientsByEp` is a map of the in memory `ClientRegistry` implementation. So you can not consider that this class will be always here. We have a [redis version](https://github.com/eclipse/leshan/blob/redis/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/cluster/RedisClientRegistry.java#L50) in development  where this map does not exists.
For me `ClientRegistry` should not handle pending request this is not its responsibility.

boaks: Sure, the `ClientRegistry` should not handle pending request, but it's the central map for `Client`, regardless of its implementation.
The `CaliforniumLwM2mRequestSender.send` therefore has already a `Client` as parameter, so in my idea there is not to use a "second map/lookup" and instead use that Client. If that `Client` is extended by a List of `RequestCanceler` (see PR #102) it would be easy to cancel that list. But it was just an idea. Though this PR ist a Bugfix, I think it's better to find a solution then to discuss possibilities over and over. Therefore my proposal (see PR #103) is much closer to this solution than my "idea".

sbernard31: Totally agree, we should  keep focus on this bugfix PR. 
(About the use of Client object, I see what you mean, but this will not work with a not in memory implementation because you could have 2 `Client` instances which will represent the same Client/Registration. Each get on ClientRegistry will return a new instance)

boaks: Each get on ClientRegistry will return a new instance

OK, I see.

sbernard31:
I hope I fixed the concurrency issue and also added the Achim's test..

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -137,34 +139,41 @@ public T buildResponse(final Response coapResponse) {
...
+    private String getCeilingKey(String registrationId) {
+        if (registrationId == null)
+            return null;
+        return registrationId + "#A";
+    }
+
+    private String getKey(String registrationId, Request coapRequest) {
+        if (registrationId == null || coapRequest == null)
+            return null;
+        return registrationId + '#' + coapRequest.hashCode();
```
boaks: I think, this relys on a unique `hashCode()` ...

sbernard31: The java doc say :
```java
    /**
     * ....
     * As much as is reasonably practical, the hashCode method defined by
     * class {@code Object} does return distinct integers for distinct
     * objects. (This is typically implemented by converting the internal
     * address of the object into an integer, but this implementation
     * technique is not required by the
     * Java<font size="-2"><sup>TM</sup></font> programming language.)
```
But maybe this is not the best approach... I thought about the CoAP messageID  but not sure this is better?
The CoAP spec say that:
"The same Message ID MUST NOT be reused (in communicating with the same endpoint) within the EXCHANGE_LIFETIME"_

jvermillard: Use  default hashcode or MID. I won't be :bike: shedding on this one :)

boaks: Though `addPendingRequest` always creates a `CleanerMessageObserver`, it should not be too hard to move the "key" (reg-id + unique id) to that class and then use that key in add and remove.
So there is no need to depend on `hashCode` of a third party library (even if we know, it's not really third party :-).

sbernard31:
I think all comments were taken into account.
So I rebased the code and cleaned the history to make this more readable for the next reviewers. (@msangoi ?)

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
```diff
@@ -86,6 +91,79 @@ public void register_update_deregister() {
...
+    @Test
+    public void deregister_cancel_multiple_pending_request() throws InterruptedException {
```
msangoi: With this test, the first unit test with a single request (deregister_cancel_pending_request) is probably not necessary anymore.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -108,23 +119,109 @@ public T buildResponse(final Response coapResponse) {
...
+        // Store pending request to cancel it on deregistration
+        addPendingRequest(destination.getRegistrationId(), coapRequest);
 
         // Send CoAP request asynchronously
         final Endpoint endpoint = getEndpointForClient(destination);
         endpoint.sendRequest(coapRequest);
     }
 
+    private String getFloorKey(String registrationId) {
+        if (registrationId == null)
```
msangoi: This check is not needed. If this method returns `null`, a NPE will be raised later (`concurrentSkipListMap.subMap()`).
The null-check can be moved to the `cancelPendingRequests()` method.
Same thing for `getCeilingKey()`.

sbernard31:
I added comment about keys generation and also changed the `null` test by a `Validate.notNull()`
This is now integrated in master (commit ead8c0323da676f8cf1b39937b73e83a78b05838, and 292dc374d5074b9a2e0eb1aff7e296ab4a5e2076)
Thx all for feedback.

PR #102, opened 2016-03-16
(NOT TO BE MERGED) Introduce cancel request for client side.

boaks:
PLEASE DO NOT MERGE!

This should give a first idea, how "cancel request" could be introduced into the "core" (here client side). This "idea" would make it possible, to move parts of "cancel on deregistration" into the core (if that is wanted).  


boaks:
closed with #100

PR #103, opened 2016-03-16
Use concurrent collections instead of  synchronized methods.

boaks:
Use concurrent collections instead of  synchronized methods.
Add test for cancel multiple requests.

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
@@ -16,12 +16,16 @@
 package org.eclipse.leshan.server.californium.impl;
 
 import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.eclipse.californium.core.coap.MessageObserver;
 import org.eclipse.californium.core.coap.MessageObserverAdapter;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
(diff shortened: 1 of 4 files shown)


boaks:
The downside may be, that a "request" for a Client, which is currently "deregistering", will make trouble :-).

To fix that, I would like to introduce a "pendingDeregistration" in the "org.eclipse.leshan.server.client.Client". 
This seems for me to be better than a "global synchronized". 
Any comments/opinions on that?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -137,22 +138,23 @@ public T buildResponse(final Response coapResponse) {
...
 
-    private synchronized void addPendingRequest(String registrationId, Request coapRequest) {
+    private void addPendingRequest(String registrationId, Request coapRequest) {
         // Get or create all pending requests for this registrationId
-        ArrayList<Request> requests = pendingRequests.get(registrationId);
+        List<Request> requests = pendingRequests.get(registrationId);
         if (requests == null) {
-            requests = new ArrayList<>();
-            pendingRequests.put(registrationId, requests);
+            requests = Collections.synchronizedList(new ArrayList<Request>());
```
sbernard31: This is not thread safe.  2 threads could check that requests == null and so add the new collections twice. (Only one request will be added.

boaks: Uups sorry, I thought

            requests = Collections.synchronizedList(new ArrayList<Request>());
            List<Request> availableRequests = pendingRequests.putIfAbsent(registrationId, requests);
            if (null != availableRequests)
                requests = availableRequests;

the `putIfAbsent` would fix that.

boaks:
From PR #100 

    Each get on ClientRegistry will return a new instance

This invalidates the proposal extending Client.

But the "race condition" is also valid for PR #100.
Any previous retrieved `Client` will cause problems when used in `send` after the deregister.
Even the synchronized `cancelPendingRequests`/`addPendingRequest` will not solve this, because you can't ensure, that someone who retrieves a `Client` finishs using it before `cancelPendingRequests` is called. And a `send` after `cancelPendingRequests` will always be handled wrong.
So we need a solution ...

sbernard31:
I think we can close this one, right ?

boaks:
sure.

PR #105, opened 2016-03-17
Move pending request key to CleanerMessageObserver.

boaks:
Remove dependency from "hashCode" of third party class.
 

sbernard31:
Integrated in #100 PR.

PR #106, opened 2016-03-17
Check BS authentication before computing BS config

phtrivier:
Currently, BS configuration to send to a device is
computed before checking that the BS credentials
sent by the device are valid.

When rotating BS credentials, a server might generate
new credentials as a result of checking credentials.

Changing the order or the calls makes it possible
to send newly generated BS credentials as soon as
they are generated.

Signed-off-by: Pierre-Henri Trivier <phtrivier@sierrawireless.com>


sbernard31:
Integrated in master (commit bcb9a5483dc4b6ba5fd69a070691151226c4bfca)

PR #107, opened 2016-03-18
Fix bootstrap write server id.

boaks:
Fix writing the `short server id` in the LWM2M Server.

Currently no junit test could be provided, because the used demo bootstrap server will first erase all instances and so create is instead of replace.


sbernard31:
Integrated in master (commit 03ab830010fcec874725cd8f14125c17d1878c2b)
(I just invert identity and request parameters for homogeneous reason)

PR #108, opened 2016-03-18
Add json configuration file format.

boaks:
Add json format for configuration file and parameter for that filename.

If filename ends with ".json" son format is used to store the configurations.
So, the content could be edited to adjust the values to your need not
limited to the possiblities of the web gui. There for special "local setup"
are possible.

### DIFF
--- a/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapStoreImpl.java
+++ b/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapStoreImpl.java
@@ -12,14 +12,18 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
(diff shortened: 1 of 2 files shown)


jvermillard:
nice idea
I think we can just replace the old format with json serialization.
Also I'm not fan of gson high level API and would like to migrate everything to [minimal-json](https://github.com/ralfstx/minimal-json/)

boaks:
Also I'm not fan of gson high level API and would like to migrate everything to minimal-json

I agree, that this may be not a "long term" stable solution :-), but it helps to quickly enable other to adjust the web gui created configuration according their needs.
Therefore my idea was just to add it as alternative and leave the "long term" stuff to someone else.

sbernard31:
I remove the old serialization and fix the environment variable name. 
_"All options could be passed using environment variables.(using long option name in uppercase)"_

For minimal-json we will do that later.
This is integrated in master (commit 8279220bd03c99fc39e275c3b8f4b1d97aaad0e8)

PR #111, opened 2016-03-29
Introduce server extensions.

boaks:
Though the "standalone"-server was promoted to a "demo"-server, it may be of general interest to add simple extension to it. Therefore I added a simple JSON configured extension functionality. I use this mainly for extend the server to setup clients and to setup the server in special test modes for testing clients.

Signed-off-by: Achim Kraus achim.kraus@bosch-si.com

### DIFF
--- a/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/extensions/MessageDelayerExtension.java
+++ b/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/extensions/MessageDelayerExtension.java
@@ -1,0 +1,198 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Achim Kraus (Bosch Software Innovations GmbH) - Initial contribution
+ ******************************************************************************/
+package org.eclipse.leshan.server.demo.extensions;
+
+import java.util.Random;
+
+import org.eclipse.californium.core.coap.EmptyMessage;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.coap.Response;
+import org.eclipse.californium.core.network.Endpoint;
+import org.eclipse.californium.core.network.interceptors.MessageInterceptor;
+import org.eclipse.leshan.server.californium.impl.LeshanServer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Message delayer extensions. Delays processing of messages to simulate processing under stress condition. Used for
+ * robustness test of clients. The extension toggles between delayed and undelayed processing to see, if the client is
+ * able to work under "normal" condition when it was "stressed before.
+ */
+public class MessageDelayerExtension implements MessageInterceptor, LeshanServerExtension {
+    /**
+     * Configuration key for maximum number of undelayed processed messages.
+     */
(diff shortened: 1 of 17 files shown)


jvermillard:
this is interesting you see renaming "standalone" to "demo" as a promotion :smile: 

In my mind extension would be done using the HTTP/JSON API.

My feeling is you want to create an automatic testing platform, I don't think it's in the scope of this "demo" server

boaks:
In my mind extension would be done using the HTTP/JSON API.

I would prefer a “in process solution”, not to be influenced by additional communication (delays, instability).

    My feeling is you want to create an automatic testing platform, I don't think it's in the scope of this "demo" server

Yes, your feeling is right!
 I’m not sure, for which purpose the demo server is used by others, but from my side it’s a perfect “close to real life test” for client implementations. And not to change the primary “scope” of the demo server, I added the “test stuff” as optional extension. The extensions provided in the PR are just a collection of extensions, I thought they could be interesting for others all so. With the “LeshanServerExtension“ it’s possible to implement specific extensions and use them, without changing the source of the demo server.
And the extensions are not just usable for “test automation”, they also provide more basic support e.g. the “FileResourceExtension” offers a COAP base file server, which may be used for firmware update. Or the "MessageTracerExtension" for analyzing the communication, if wireshark is not possible.

boaks:
I understand that, from the viewpoint of the leshan project, it's critical not to overload the demo server within the leshan project. 
But therefore a simple extension mechanism would enable others to do that on their own without "continuously patching" the server demo.
So, if I reduce the PR to `ExtensionConfig,` `LeshanServerExtension,` `LeshanServerExtensionManager,` `Tagging` and one extension implementation (e.g. `FileResourceExtension` for support (small) firmware updates over CoAP), would there be a chance to get it merged?

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/extensions/MessageDelayerExtension.java
```diff
@@ -0,0 +1,198 @@
...
+import org.eclipse.leshan.server.californium.impl.LeshanServer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Message delayer extensions. Delays processing of messages to simulate processing under stress condition. Used for
+ * robustness test of clients. The extension toggles between delayed and undelayed processing to see, if the client is
+ * able to work under "normal" condition when it was "stressed before.
+ */
+public class MessageDelayerExtension implements MessageInterceptor, LeshanServerExtension {
```
jvermillard: Why not use http://www.linuxfoundation.org/collaborate/workgroups/networking/netem in place of thus interceptor ?

boaks: Because:
- not everyone use linux
- not to read the documentation of an other tool

I'm not saying, that all extension currently contained in the PR are "usefull" for all.
But the principle of extension may be usefull. Therefore I proposed to shrink it, if it would then have chance.

(The "noise test stuff" will only be usefull together with the ToggleExtension. This temporary enables/disables the noise so you can check, if a device get connected again after the noise.)

sbernard31:
I also think this is a out of scope. I mean currently Leshan-server-demo is just a demo.
But you're not the first one who want to reuse leshan-server-demo. So I suppose we need to search a good way to aim this need. We plan to provide a server more production ready maybe this will fit this need. (This way you can reuse it to implement your test server)

boaks:
Hello again,
I still feel, that an "extension" mechanism is very useful!
With LWM2M TS using CoAP blockwise transfer for "Firmware Update", I would like to propose again to reduce the extension to logging and FileResource. With those two and a "fake firmware image", the sandbox leshan would allow clients to test the "Firmware Download".

msangoi:
A new opinion on this old/stuck PR :)
As stated by Simon, I also think that the purpose of the demo server should be restricted to be a simple demo. And it should not be used as a starting point to build a server. 
And I think that integrating an "extension mechanism" to the demo server will encourage people to build servers on top of it.
Apart from the "extension" idea, I like the idea to add features to the demo server so that people can use the sandbox to test more things (e.g. firmware download).

boaks:
> And I think that integrating an "extension mechanism" to the demo server will encourage people to build servers on top of it.

I'm not sure, if forcing people to "copy"/"continously patching" instead of "offer extending" the demo is the right approach. But sure, if not wanted, I could keep the stuff for internal use.

PR #93, opened 2016-03-08
New Observe implementation based on experimental Californium observe impl.

sbernard31:
This is a new implementation of ObservationRegistry based on the californium [PR #16](https://github.com/eclipse/californium/pull/16).

This should allow us to deploy [Leshan in a Cluster](https://github.com/eclipse/leshan/wiki/Cluster).

PR #113, opened 2016-04-01
Add clustering support (based on Redis)

sbernard31:
This PR aim to allow to deploy [Leshan in a Cluster](https://github.com/eclipse/leshan/wiki/Cluster).

The first step was to add  `ClientRegistry`, `SecurityRegistry` and `ObservationRegistry` implementation based on [Redis](http://redis.io/) to Leshan Server Demo. (Previously done in #89  and #93)

The Redis `ObservationRegistry` implementation is based on the californium [PR #16](https://github.com/eclipse/californium/pull/16). So you need to get it and build it locally to test it.

**This work is still in progress.**


ingoschaal:
Hi Julien
based on the discussion in PR#89 I have a suggestion:
I fully agree with you to keep leshan-core as a pure LWM2M protocol library, without clustering. 
But what about those who do not use the demo server , rather use their own server implementation.
Perhaps we could establish a new Cluster Layer which could used from demo serve and all other server implementations, but not have to.

ingoschaal:
Hello everybody
are there any news regarding this two impediments?

CQ under review for jedis + common-pool2
http://dev.eclipse.org/ipzilla/show_bug.cgi?id=11027
http://dev.eclipse.org/ipzilla/show_bug.cgi?id=11029

otherwise we would be pleased if this PR is merged into the master. 
We have had good experience and think that it is a good basis for proposed [Cluster](https://github.com/eclipse/leshan/wiki/Cluster)

jvermillard:
Thanks @ingoschaal for the feedback!
The CQ are validated. Now we are waiting for https://github.com/eclipse/californium/pull/16 merge & release in californium.

ingoschaal:
Another question:
The consistency concept of Redis is eventual consistency. This means that the write operation is not directly propagated to all replicas.
The KeyToken which is generated in Californium Matcher should be unique to the server. 
What happens when the same token is generated to be persisted on a cluster which is not synchronized at this thime? 
Is redis at this point the right choice?
Or concerns this Problem only [PR #16](https://github.com/eclipse/californium/pull/16)

mhemmeter:
Hi @jvermillard and others,

regarding the upcoming merge into the master branch: Currently all LWM2M Servers that are based on Eclipse Leshan but have their own server implementation would need to include the demo server in order to use the new clustering feature. That's something I would like to avoid in our product and therefore I would propose to introduce a new module _leshan-server-cluster_ (as an alternative one could also introduce a more generic module for all upcoming extensions of the core library, something like _leshan-server-extensions_).

What do you and others think? If everybody agrees we can provide a PR for that.

Thanks
Mathias

sbernard31:
@mhemmeter,
I'm glad to see that you want to avoid to depend on leshan-server-demo, We take a lot of time to explain that leshan-server-demo is not intended to be directly reused. It's just a demo/example :)

Currently, we put all redis implementation in leshan-server-demo because this is still an experimental works, but we aims to make this reusable and so this will not stay in leshan-server-demo for ever ;) !

sophokles73:
@sbernard31  I see you point and in general I would agree that the Redis stuff is still *experimental*. However, we could make it easier for people like @mhemmeter to test (end evolve) the code by putting it into a separate module that would not require them to depend on the demo server.

We can still put a disclaimer in there stating that this module is considered experimental ...

sbernard31:
@ingoschaal  about the token issue.
In some [rare cases](http://redis.io/topics/cluster-tutorial#redis-cluster-consistency-guarantees), we can lost a token and so an observation. This is a trade-off between performance and consistency.
We will not be able to provide an implementation which covers all needs that's why we define interfaces to let users free to use adapted technology.

@sophokles73 I'm not sure to see the difference between depends on `leshan-server-demo` or `leshen-server-experimental` ;)

sophokles73:
That's why I wouldn't call it `leshan-server-experimental` but put it in the right module from the beginning, e.g. `leshan-server-clustering` or whatever :-)

jvermillard:
IMO the leshan maven modules already reached an unbearable level of over engineering for me :)
20k SLOC  and 9 modules.

Before introducing more module I would like to explore if we can merge some.

Since the main proponents of having californium  replaceable are gone I think we could merge: 
 * leshan-server-core and leshan-server-cf
 * leshan-client-core and leshan-client-cf

sophokles73:
I have no problem at all with merging them.
I also do not think that we *need* a separate module for the clustering stuff. We can simply add it to a new package within *leshan-server-core* as well, I guess.

Just out of curiosity: who is the party that was proposing the split up and that is now *gone*? Did I miss something?

mhemmeter:
@sophokles73: In  PR #89 it has been discussed that the _leshan-server-core_ module is currently seen as a pure LWM2M protocol library and therefore the clustering feature was implemented as part of the demo server. As I can follow this argument I proposed to create a new module. But I would also be fine if it is included in the core in a separate package. My only wish is that it is not in the demo server :-) 

Background of the request is that we are planning to invest in making the clustering feature production ready in the next months. So I would like to start having it in the right module from the beginning. `leshan-server-clustering` sounds good to me if it becomes a separate module.

sbernard31:
I'm not sure the among of modules is a real problem.
The module separation is pretty simple : https://github.com/eclipse/leshan#modules
On the 9 module there is only 5 real modules (the others are just for tests and demos)

About merging cf and core, It would be a shame to lost the abstraction we introduce between LWM2M and transport layer. It could help us to implement LWM2M over other implementation or transport layer (CoAP/TCP).

About clustering, I'm not ok to create a leshan-server-clustering now, because I'm currently not able  to know if it's the right way to separate the code (for now we have only 2 classes to put on it ...). I suppose we can  move this 2 redis implementation from demo to core (with redis optional depency on server-core), but this will not make it less experimental. (sweeping dust under the carpet?)

jpradocueva:
Just for information.
OMA is developing OMA LwM2M v1.1 where one of the new features is CoAP over TCP.
I can prepare a summary of what is planned for OMA LwM2M v1.1 and post it in the distribution list if it helps.

jvermillard:
> About merging cf and core, It would be a shame to lost the abstraction we introduce between LWM2M and transport layer. It could help us to implement LWM2M over other implementation or transport layer (CoAP/TCP).
It's not about removing the abstraction, just keeping everything in one module, you are free to not use the Cf impl.
I think I'm a bit out of topic here :)

mhemmeter:
Seems like there is a veto from @sbernard31 for creating a separate clustering module. So what about moving the clustering implementation to `leshan-server-core`? If there are no objections we will create the pull request.

sophokles73:
Ok, here's what I propose:

* we create a module `leshan-server-clustering` in which create and evolve the clustering support. From my point of view it is totally acceptable to have code in there that *is not (yet) perfect* (in the end, which of our code is?).
* indedpendently from that we can discuss which (other) modules to merge. I see the value of keeping the basic modules clean of dependencies that are only used for clustering, thus I prefer to have a separate module for it.

sbernard31:
I cleaned the commit history and rebased it on master. I also move "clustering" code to a new  **leshan-server-cluster** module.
We are waiting now for a Californium 2.0.0-M1 release to integrate this PR in master.

sbernard31:
This is now integrated in master :)

tomaszszlek:
Sorry if this is a stupid question but what kind of advantages brings cluster based on Redis over usage of Bootstrap server ? My understanding is that Bootstrap server can act as load balancer and distribute clients between Leshan server instances. Will appreciate answer on this.

sbernard31:
The bootstrap interface aims to _provide essential information into the LWM2M Client to enable the LWM2M Client to perform the operation “Register” with one or more LWM2M Servers._

The idea is more about initializing information like credentials and server url.

You can have more details [here](https://medium.com/@vrmvrm/device-key-distribution-with-lightweight-m2m-36cdc12e5711#.llzeug2dt).

tomaszszlek:
Thank you for answer, I read this article and I was wondering if Bootstrap server can be used as Load Balancer. It seems that someone who created idea of Bootstrap server didn't took it under consideration.
Is it in your opinion good alternative to Leshan cluster which is based on Redis ? Actually we can put any logic in Bootstrap server, for example some Round Robin which could direct client to particular server based on hash.

sophokles73:
We should move this discussion to the mailing list instead of having it on a merged PR ...

PR #117, opened 2016-04-14
Added Redis URL command line option support with authentication

MKremserBosch:
Hi all,
added support for Redis password authentication in the demo server.
Changed command line option `r` (redis) to location of Redis database as URL.
Where the URL is in the format of: `redis://:password@hostname:port/db_number`
The fields password and db_number after the scheme are optional, default is database 0.
Example without database and password: `redis://localhost:6379`

Signed-off-by: Michael Kremser <Michael.Kremser@bosch-si.com>


sbernard31:
Thx gizmo ;)
thx gizmo

PR #123, opened 2016-04-26
Queue mode

balsmn:
# Intro

A **Queue mode** is an operation mode of LWM2M protocol (since the early versions

    A Client has to register using specified binding mode (e.g. using "UQ" for "UDP+Queue");
    A Server has to understand this binding mode (it is mandatory);
    A client is not expected to be always online - instead, a server must queue the requests and only allowed to send downlink requests one by one to the client when server received client's registration update or notification.

-    **org.eclipse.leshan.server.queue** - contains base interfaces to operate with queue mode, i.e. public API for sending downlink requests and for holding messages in a message store;
-     **org.eclipse.leshan.server.queue.impl** - provides an InMemory implementation of the message store and for deliverying the messages to the client and processing of responses.

# org.eclipse.leshan.server.queue - Queue Mode Public API
## MessageStore
Provides methods for adding a message for a client, removing, getting the first message and for emptying the queue of a client. Once the message is put in the queue, it will eventually be delivered to a client if it is reachable or when the client becomes reachable. If a client de-registers or its life time has expired, then the queue is emptied.

## QueuedRequest

A single element of a queue is represented by a QueuedRequest. Amongst the downlink request  it also holds the request Id for tracking the response context and the client's endpoint.

## QueuedRequestFactory

Provides factory methods to create the above defined QueuedRequest that could be stored in a message store.

## ClientState

Contains different states of a client that are relevant for delivery of queued messages. There are certain valid state transitions that concrete implementations needs to enforce so as to get a consistent processing of queued messages. The valid state transitions are as explained below.

- Once the client gets connected to the LWM2M Server it is considered to be **_REACHABLE_**. 
- Client is in state **_RECEIVING_** before receving any messages.
- When no more messages are left in the queue, the client is in state **_REACHABLE_**.
- When a client times out on a message, it is in state **_UNREACHABLE_**.
- When a client sends a register-update or a notify, then client state is changed from **_UNREACHABLE_** to **_REACHABLE_**.

# org.eclipse.leshan.server.queue.impl - Queue Mode Implementation

## InMemoryMessageStore

Provides an InMemory implementation of the MessageStore for holding queued messages until the client is available again or client has de-registered (either by sending a de-register message or when client life time is up).

## QueuedRequestSender

QueuedRequestSender is an entry point for queue mode related message sending.
This is basically an extension of LwM2mRequestSender, which is capable of maintaining a queue of requests for every client registered with queue binding mode and delivers the messages to the client one after other when they are online.

## ClientStateTracker

Provides methods for state transitions and enforces valid state transitions define above in ClientState.

Also-by: Daniel Maier daniel.maier@bosch-si.com


balsmn:
Closing this PR due to ip-validation error. One possiblity is that I have my email id with brackets

sbernard31:
You don't need to close the PR. Just use `git rebase --interactive` to fix the issue then do a `git push --force`.

alex-ellwein:
@sbernard31, this is what it feels like if you ``git push --force``... 
SCNR ;)
pushforce

sbernard31:
:D !

balsmn:
@sbernard31 yes @boaks  explained me that I don't have to close it everytime. I created one today but again failed with IP validation errors. I am trying to get it right without closing.
Yes, I am doing with --rebase now to fix it.

PR #126, opened 2016-04-27
Queue mode

balsmn:
# Intro

A **Queue mode** is an operation mode of LWM2M protocol (since the early versions

    A Client has to register using specified binding mode (e.g. using "UQ" for "UDP+Queue");
    A Server has to understand this binding mode (it is mandatory);
    A client is not expected to be always online - instead, a server must queue the requests and only allowed to send downlink requests one by one to the client when server received client's registration update or notification.

-    **org.eclipse.leshan.server.queue** - contains base interfaces to operate with queue mode, i.e. public API for sending downlink requests and for holding messages in a message store;
-     **org.eclipse.leshan.server.queue.impl** - provides an InMemory implementation of the message store and for deliverying the messages to the client and processing of responses.

# org.eclipse.leshan.server.queue - Queue Mode Public API
## MessageStore
Provides methods for adding a message for a client, removing, getting the first message and for emptying the queue of a client. Once the message is put in the queue, it will eventually be delivered to a client if it is reachable or when the client becomes reachable. If a client de-registers or its life time has expired, then the queue is emptied.

## QueuedRequest

A single element of a queue is represented by a QueuedRequest. Amongst the downlink request  it also holds the request Id for tracking the response context and the client's endpoint.

## QueuedRequestFactory

Provides factory methods to create the above defined QueuedRequest that could be stored in a message store.

## ClientState

Contains different states of a client that are relevant for delivery of queued messages. There are certain valid state transitions that concrete implementations needs to enforce so as to get a consistent processing of queued messages. The valid state transitions are as explained below.

- Once the client gets connected to the LWM2M Server it is considered to be **_REACHABLE_**. 
- Client is in state **_RECEIVING_** before receving any messages.
- When no more messages are left in the queue, the client is in state **_REACHABLE_**.
- When a client times out on a message, it is in state **_UNREACHABLE_**.
- When a client sends a register-update or a notify, then client state is changed from **_UNREACHABLE_** to **_REACHABLE_**.

# org.eclipse.leshan.server.queue.impl - Queue Mode Implementation

## InMemoryMessageStore

Provides an InMemory implementation of the MessageStore for holding queued messages until the client is available again or client has de-registered (either by sending a de-register message or when client life time is up).

## QueuedRequestSender

QueuedRequestSender is an entry point for queue mode related message sending.
This is basically an extension of LwM2mRequestSender, which is capable of maintaining a queue of requests for every client registered with queue binding mode and delivers the messages to the client one after other when they are online.

## ClientStateTracker

Provides methods for state transitions and enforces valid state transitions define above in ClientState.

Also-by: Alexander Ellwein <alexander.ellwein@bosch-si.com>
Also-by: Daniel Maier <daniel.maier@bosch-si.com>
Signed-off-by: Balasubramanian Azhagappan <balasubramanian.azhagappan@bosch-si.com>


balsmn:
@sbernard31 I have signed CLA and have checked its validity. Also tried with multiple Signed-off-by with my full name, no name just email id, without line breaks, with line breaks after Signed-Off-by. I also checked my email id in my profile. I have only one email in my account. Still IP Validation doesn't go through. 

Is there any special format for Signed-Off-by. I used eclipse IDE to Signed-Off.

balsmn:
I checked with another pull request, where I am the only Author. There it seems to work. 

https://github.com/eclipse/leshan/pull/127 

So now I have to check why it fails when there is code from multiple authors.

sbernard31:
Very weird ...
I can't find any problems when I check it manually.
Maybe there is a bug in the ip-validation script. You can try to [open a bug](https://bugs.eclipse.org/bugs/enter_bug.cgi?product=Community) at eclipse foundation eventually.

balsmn:
@sbernard31 Thanks for the bug tracker link. There I found a similar bug, https://bugs.eclipse.org/bugs/show_bug.cgi?id=491373 where the possible solution is to "git squash" commits from more than one authors in to one commit. I tried and it worked.

balsmn:
Looks like NamedThreadFactory is already implemented in Californium with the PR https://github.com/eclipse/californium/pull/39. When it becomes part of the official californium release, Leshan can refer to it and we don't have the need for another NamedThreadFactory here.

sophokles73:
I guess it depends on just how close you want to tie the implementation to Californium. In leshan we have tried to depend on Californium only by means of well defined interfaces. So you might also consider to define a separate NamedThreadFactory for lesahn in order to decrease (or better: not increase) coupling ....

balsmn:
@sophokles73 I see your point. Ok, lets leave it this way or we could also move it to a utility package.

leshan-integration-tests/credentials/client.csr
```diff
@@ -1,9 +1,9 @@
------BEGIN NEW CERTIFICATE REQUEST-----
+-----BEGIN NEW CERTIFICATE REQUEST-----
```
sophokles73: What do we need this for?

leshan-integration-tests/credentials/server.csr
```diff
@@ -1,9 +1,9 @@
------BEGIN NEW CERTIFICATE REQUEST-----
+-----BEGIN NEW CERTIFICATE REQUEST-----
```
sophokles73: What do we need this for?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
```diff
@@ -0,0 +1,282 @@
...
+     * Creates a new QueueRequestSender using given builder.
+     *
+     * @param builder sender builder
+     */
+    private QueuedRequestSender(Builder builder) {
+        this.messageStore = builder.messageStore;
+        this.queuedRequestFactory = builder.queuedRequestFactory;
+        this.clientRegistry = builder.clientRegistry;
+        this.observationRegistry = builder.observationRegistry;
+        this.delegateSender = builder.delegateSender;
```
sophokles73: move copying of fields to `Builder.build()` in order to only have `Builder` depend on `QueuedRequestSender` but not the other way around as well.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
```diff
@@ -0,0 +1,282 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * This sender is a special implementation of a {@link LwM2mRequestSender}, which is aware of a LwM2M client's binding
+ * mode. If the client supports "Q" (queue) mode, then the request is processed using the internal queue, i.e. the
+ * request is enqueued first and then sent if a client is back online. If the client does not support the queue mode in
+ * its binding, then the delegate sender is called in order to send the request.
+ */
+public class QueuedRequestSender implements LwM2mRequestSender, Stoppable {
+    private static final Logger LOG = LoggerFactory.getLogger(QueuedRequestSender.class);
```
sophokles73: Is it really necessary to make this a wrapper around the  *standard* `LwM2mRequestSender`? It feels like we are introducing a lot of unnecessary code (the listeners) or am I mistaken?
Wouldn't we instead want to add the queue mode functionality to the *standard* `LwM2mRequestSender`?

leshan-integration-tests/credentials/client.csr
balsmn: It is actually from @sbernard31 and may be because of a rebase/building it locally once.

leshan-integration-tests/credentials/server.csr
balsmn: It is actually from @sbernard31 and may be because of a rebase/building it locally once.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
DanielMaier-BSI: I don't get your point here. Isn't one goal of builder pattern to have all fields final?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
balsmn: Yes you are correct. As queue mode is a mandatory feature that the server should support, it would be great to have the queue mode functionality as part of the standard `LwM2mRequestSender` and use the AsyncRequestObserver . Thought of doing one step at a time. As of now left this Sender as an extension to `LwM2mRequestSender`. It can be done as the next PR once this one is accepted. Also as discussed by @sophokles73 and @jvermillard in https://github.com/eclipse/leshan/pull/47#issuecomment-157369748, it would be great to have API change in `LwM2mRequestSender`.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
sophokles73: Well, it certainly hasn't been my goal in the past :-) My goal has always been to make instantiation easier and make sure that properties are set in a consistent way.

DanielMaier-BSI: But you loose the advantage that you had before you introduced the builder: Setting fields on construction time. You don't have to bother about thread-safety when accessing these fields. Otherwise you might need to make these fields volatile.

sophokles73: I'm sorry but I do not seem to follow ... can you give an example of what you want to achieve?

DanielMaier-BSI: Mh, perhaps a link with an example is better to get a common understanding: http://www.informit.com/articles/article.aspx?p=1216151&seqNum=2

The only thing I wanted to point out is that if you do not pass the builder itself to the constructor you are not able to make the fields of the class final, thus the object immutable, which may introduce some effort to guarantee the visibility of the fields when accessed from multiple threads.

sophokles73: I still don't get it. The class has no setters, does it? Its state is only changed by means of the its other methods which are there regardless of whether the fields are defined final or not.

DanielMaier-BSI: The final keyword has some impact to the Java Memory model. With final fields it is guaranteed that each thread sees the correct values of the fields without synchronization or volatile stuff.

https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.5

sophokles73: Two things:
* if Thread safety is an important and relevant feature of this class then it should be documented in the JavaDoc.
* I am not sure if making the fields final actually makes this class immutable (and thus thread safe). The fields only contain references to the objects (e.g. the `clientRegistry` field contains a reference to an instance of `ClientRegistry`). I do not read the section of the spec you refer to as guaranteeing that any invocation of methods on the `ClientRegistry` instance will always be thread-safe as well, or does it? I think it only guarantees that two threads that read the `clientRegistry` field will get a reference to exactly the same version of the object (the `ClientRegistry` instance, that is). Or am I mistaken?

DanielMaier-BSI: Yes I think thread-safety is an important feature of this class because it gets used by foreign code that should not be responsible to wrap this class in a thread-safe manner. Of course this should be stated in the JavaDoc.

Yes it is only about visibility from different threads, not about thread-safety altogether. Of course there still can be multi-threading issues regarding other invariants. But it is also guaranteed that threads see referenced fields at least as up-to-date as the final fields are.

Besides of this visibility issue, from my point of view it is always a good thing to make fields final wherever is possible to emphasize that this fields should not be modified after initialization. If I refactor code and introduce a new setter I think twice about it if the field was final before.

sophokles73: Absolutely, I agree with you regarding the desirability of making fields final.
Anyway, we got carried away a little and after giving it some additional thought I guess it is also fine you did it now because the constructor is private anyways and nobody will need to create a `Builder` instance and pass it to the constructor.

sbernard31:
I looked at the PR. 
I see there is a MessageStore interface. I suppose the idea is to allow to implement a persistent MessageStore. Probably to support failover or clustering. This is a good thing because this is something we want for Leshan.
But If the messageStore allow us to not lost or share requests in the queue. Our current in memory callback API for response does not allow us to get response of request after a failover or in a cluster.

So this means we should probably break the API or extend it.

leshan-integration-tests/credentials/server.csr
jvermillard: all the windows fault no? :)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/ClientStatusTracker.java
```diff
@@ -0,0 +1,77 @@
...
+        clientStatus.remove(endpoint);
+    }
+
+    private boolean transitState(String endpoint, ClientState from, ClientState to) {
+        boolean updated = clientStatus.replace(endpoint, from, to);
+        if(updated) {
+            LOG.debug("Client {} state update {} -> {}", endpoint, from, to);
+        }
+        else {
+            LOG.debug("Cannot update Client {} state {} -> {}. Current state is {}",
```
jvermillard: if isDebugEnabled is missing here

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/NamedThreadFactory.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Implementation of {@link ThreadFactory} that sets thread names according to given name format. All threads are
+ * created by {@link Executors#defaultThreadFactory() #newThread(Runnable)}.
+ */
+final class NamedThreadFactory implements ThreadFactory {
```
jvermillard: go to util?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/RequestSendingTask.java
```diff
@@ -0,0 +1,137 @@
...
+        if (firstRequest != null) {
+            try {
+                // TODO send with COAP retries ok? See
+                // https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/32
+                final DownlinkRequest<LwM2mResponse> downlinkRequest = firstRequest.getDownlinkRequest();
+                LOG.debug("Sending request: {}", downlinkRequest);
+                final Client client = clientRegistry.get(firstRequest.getEndpoint());
+                if (client == null) {
+                    // client not registered anymore -> ignore this request
+                    LOG.debug("Client {} not registered anymore: {}", firstRequest.getEndpoint(), downlinkRequest);
```
jvermillard: isDebugEnabled

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/RequestSendingTask.java
```diff
@@ -0,0 +1,137 @@
...
+        this.responseContextHolder = responseContextHolder;
+    }
+
+    @Override
+    public void run() {
+        final QueuedRequest firstRequest = messageStore.retrieveFirst(endpoint);
+
+        if (firstRequest != null) {
+            try {
+                // TODO send with COAP retries ok? See
```
jvermillard: Can be removed since the discussion is closed?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/RequestSendingTask.java
```diff
@@ -0,0 +1,137 @@
...
+    private void processException(final QueuedRequest request, final Exception exception) {
+        LOG.debug("Received error response {}", request);
+        messageStore.deleteFirst(endpoint);
+        processingExecutor.execute(new ResponseProcessingTask(request, responseContextHolder, exception));
+        processingExecutor.execute(new RequestSendingTask(clientRegistry, requestSender, processingExecutor,
+                clientStatusTracker, messageStore, endpoint, responseContextHolder));
+    }
+
+    private void timeout(final Client client) {
+        LOG.debug("Client {} timed out", client.getEndpoint());
```
jvermillard: isDebugEnabled

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
```diff
@@ -0,0 +1,282 @@
...
+                        queueProcessingExecutorTerminated);
+            }
+        } catch (InterruptedException e) {
+            LOG.debug("Interrupted while stopping. Abort stopping.");
+            Thread.currentThread().interrupt();
+        }
+    }
+
+    private RequestSendingTask newRequestSendingTask(final String endpoint) {
+        return new RequestSendingTask(clientRegistry, delegateSender, processingExecutor,
```
jvermillard: IMO you should wrap your runnable here for catching exception in place of creating your own executor class

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    private final class QueueModeObservationRegistryListener implements ObservationRegistryListener {
+        @Override
+        public void newValue(Observation observation, LwM2mNode value) {
+            Client client = clientRegistry.findByRegistrationId(observation.getRegistrationId());
+            if (isQueueMode(client.getBindingMode())
+                    && clientStatusTracker.setClientReachable(client.getEndpoint())
+                    && clientStatusTracker.startClientReceiving(client.getEndpoint())) {
+                LOG.debug("Notify from {}. Sending queued requests.", client.getEndpoint());
```
jvermillard: isDebugEnabled

sbernard31:
After some discussions with julien, we think a way to solve the clustering issue could be a new API like that:
``` java
long requestTicket = helper.generateRequestTicket();
server.addListerner(new ResponseListener(){
    public void onReponse(long requestTicket, LwM2mResponse response){}
    public void onError(long requestTicket, LwM2mResponse response){}
  }
)
server.send(requestTicket, new LwM2mRequest());
```
Clearly less user-friendly than the current one, but for now this seems to us the only one which could work.
But don't worry we will do that later.
Overall the PR seems ok, so the idea is to integrate your code in master quickly as it has no impact with the rest of the code.  I will go a much deeper review now.
I will go for a more details review.

leshan-integration-tests/credentials/client.csr
sbernard31: It seems the end of line has changed from unix style to windows style. unix style should be used.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/ClientStatusTracker.java
```diff
@@ -0,0 +1,77 @@
...
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *     Balasubramanian Azhagappan, Daniel Maier (Bosch Software Innovations GmbH)
+ *                                  - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue.impl;
+
+import java.util.ArrayList;
```
sbernard31: unused import

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/MessageStore.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Contributors:
+ *     Alexander Ellwein, Daniel Maier (Bosch Software Innovations GmbH)
+ *                                - initial API and implementation
+ *     Balasubramanian Azhagappan (Bosch Software Innovations GmbH)
+ *                                  - added removeAll method
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+import java.util.List;
```
sbernard31: unused import

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/ResponseProcessingTask.java
```diff
@@ -0,0 +1,102 @@
...
+    private final QueuedRequest request;
+
+    /**
+     * Creates a new task for processing response on exception.
+     *
+     * @param request request being processed
+     * @param responseContext response context map for mapping response ID to the callback
+     * @param exception exception to propagate
+     */
+    ResponseProcessingTask(QueuedRequest request, Map<Long, ResponseContext> responseContext,
```
sbernard31: about, responseContext, we don't need the whole map, just the context for the corresponding request ?

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/QueuedModeLeshanClient.java
```diff
@@ -0,0 +1,202 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * This client has a special ability to have its GET handler instrumented by a test by which different client
+ * scenarios like sleep, wake up can be simulated.
+ */
+public class QueuedModeLeshanClient implements LwM2mClient {
```
sbernard31: Later, we should find a way to not duplicate the whole LeshanClient

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/QueueModeLeshanServer.java
```diff
@@ -0,0 +1,183 @@
...
+        if (clientRegistry instanceof Stoppable) {
+            ((Stoppable) clientRegistry).stop();
+        }
+        if (securityRegistry instanceof Stoppable) {
+            ((Stoppable) securityRegistry).stop();
+        }
+        if (observationRegistry instanceof Stoppable) {
+            ((Stoppable) observationRegistry).stop();
+        }
+
```
sbernard31: we should stop the queueRequestSender ?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/ExceptionAwareExecutorService.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.concurrent.TimeoutException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This kind of ExecutorService is able to catch an unchecked exception while running a task and log the
+ * exception trace.
+ */
+class ExceptionAwareExecutorService implements ExecutorService {
```
sbernard31: Like julien, I'm not sure we really need this class, but if we need it we should move it in org.eclipse.leshan.util

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
```diff
@@ -0,0 +1,282 @@
...
+        return delegateSender.send(destination, request, requestTimeout);
+    }
+
+    @Override
+    public <T extends LwM2mResponse> void send(Client destination, DownlinkRequest<T> request,
+            final ResponseCallback<T> responseCallback, final ErrorCallback errorCallback) {
+
+        LOG.trace("send()");
+        // safe because DownlinkRequest does not use generics itself
+        @SuppressWarnings("unchecked")
```
sbernard31: it seems generics and Queue mode are not really good friends.. and this will not be better with the new "send" API I proposed before :/

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/ClientStatusTracker.java
balsmn: With SLF4J, you just defer toString() invocation and string concatenation until really needed, which is when you really have to log the message. As the objects "endpoint, from, to" are already instantiated here, I don't see the need to wrap it with isDebugEnabled().

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/NamedThreadFactory.java
balsmn: I was also thinking in the same line. See my comments above. Shall I move it to org.eclipse.leshan.util as part of this PR itself?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/RequestSendingTask.java
balsmn: Yes, here we have the unnecessary "firstRequest.getEndpoint()" call if debug is disabled.

I can change it to 
`LOG.debug("Client {} not registered anymore: {}", endpoint, downlinkRequest);`

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/RequestSendingTask.java
balsmn: Yep, I can add this to save the cost of client.getEndpoint() call.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
balsmn: Yep, I can add this to save the cost of client.getEndpoint() call.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/ResponseProcessingTask.java
balsmn: Yes, we need just the response context here. But later when `run()` is complete, the response context for this request should be cleaned up.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/QueuedModeLeshanClient.java
balsmn: This is used only for testing purposes so that we can set the client to or not to respond to GET. It is not part of the leshan-server-core.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/QueueModeLeshanServer.java
balsmn: Yes, I will add it.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/NamedThreadFactory.java
sbernard31: Yes please.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/ResponseProcessingTask.java
sbernard31: You're right

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/QueuedModeLeshanClient.java
sbernard31: Yes I saw but, I would like to avoid to duplicate all this code even if this is for tests only. But we will do that later.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/ClientStatusTracker.java
jvermillard: you don't delegate the "clientStatus.get(endpoint)" call

leshan-integration-tests/credentials/server.csr
sophokles73: if we didn't change the cert itself we should remove this from the commit ...

balsmn: Yes, I will revert it to the previous state

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/RequestSendingTask.java
jvermillard: :+1:

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
balsmn: Thought of having the try/catch around run() instead of each executable Task. Anyways, we have only two tasks in Queue mode and I can remove, when this ExceptionAwareExecutorService is not used anywhere else in leshan.

balsmn:
I have so far updated the PR with the recommendations and clean up except for the Generics part. May be the generics usage in LWM2MRequestSender can be reworked.

sbernard31:
This PR is integrated in master  (commit d275f53216a59e4ca9602877b48355bba2dce9f8) since Friday, but I didn't close the PR as I got [build failures](https://hudson.eclipse.org/leshan/job/leshan/208/console) since this integration.
I finally find the issue \o/ ... That was because QueueModeIntegrationTestHelper modify the standard Network Config of Californium (commit f87250ba08f75f08e623626dffb0759a8f46e709)
This "standard network configuration" concept is not really good idea. This is probably something which should be removed of californium.

Thx guys for the contributions :)

PR #129, opened 2016-04-28
Refactor bootstrap resource to put LWM2M logic in server-core

sbernard31:
This refactoring aims to move LWM2M logic from `leshan-server-cf` to `leshan-server-core` as much as possible.

This kind of job was done for `client` and `server` but never for `bootstrap server`. (This is done now ^^)

There is some crappy code [here](https://github.com/eclipse/leshan/blob/shim_bootstrap_server/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mBootstraRequestSender.java#L62), I propose to clean this latter when clustering PR will be integrated. (If we succeed to integrate the clustering `observe PR`   ... :sweat: )

I think there are some issues in the specification at bootstrap level about rootpath and spec version.
A client should declare its rootpath and the version of the spec when it start a bootstrap session ?

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/request/BootstrapRequest.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/request/BootstrapRequest.java
@@ -16,6 +16,7 @@
 package org.eclipse.leshan.core.request;
 
 import org.eclipse.leshan.core.response.BootstrapResponse;
+import org.eclipse.leshan.util.Validate;
 
 /**
  * The request to send to start a bootstrap session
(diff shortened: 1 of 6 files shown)


sbernard31:
[Here](https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/49) the OMA issue related to issues raised above.

msangoi:
👍 LGMT
Ideas for a second step: split the DownlinkRequestVisitor to have a response builder for bootstrap responses only.

sbernard31:
integrated in master (commit f5e551768319d22b635815174f9389f5e095b491)

PR #132, opened 2016-05-20
Specify the Coap Server for LeshanClient

madmax88:
Allows for the Coap server to be specified for a new instance of a LashanClient. 

In LeshanClientBuilder, the method `setCoapServer` was added, allowing for the specification of a Coap Server to be used by the LeshanClient.

In LeshanClient, a new constructor was added that allows for an existing Coap Server to be used.

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
@@ -131,6 +131,69 @@ public class LeshanClient implements LwM2mClient {
         // Create CoAP resources needed for the bootstrap sequence
         clientSideServer.add(new BootstrapResource(bootstrapHandler));
     }
+    
+     public LeshanClient(final String endpoint, final InetSocketAddress localAddress,
+            InetSocketAddress localSecureAddress, final List<? extends LwM2mObjectEnabler> objectEnablers, CoapServer server) {
+
+        Validate.notNull(endpoint);
+        Validate.notNull(localAddress);
+        Validate.notNull(localSecureAddress);
+        Validate.notEmpty(objectEnablers);
+
+        // Create Object enablers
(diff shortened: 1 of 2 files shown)


sbernard31:
Hi @madmax88,
  We currently not able to do that because we need to create a [custom root resource](https://github.com/eclipse/leshan/blob/master/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java#L115) to handle delete on "/".
  There is a `getCoapServer()` on `LeshanClient`, maybe this could fit your need ?
  If not, could you explain us what is your use case ?
Thx ;)

madmax88:
Hi @sbernard31,

I need to have several thousand Clients running on a single host. Currently, that's not possible. My thinking was that creating a new Coap server for each client was too much, so the patch was intended to work around this problem. Obviously, this won't work because of the reason you described. Do you have any recommendations?

Thanks for your response!

sbernard31:
I try to create 1000 Leshan clients in the same java VM. This use ~350MB on my computers. (Less than Firefox ;))

sbernard31:
I checked with visualVM the heap space is between 50/150MB for 1000 clients

madmax88:
I'm able to get 1000 clients working fine, but I need something more along the lines of 10,000 clients :) That's where I'm having problems.

Any recommendations?

sbernard31:
I tested with 10,000 clients, and you're right that doesn't works ..
I get a `java.lang.OutOfMemoryError: unable to create new native thread`
We create too many threads, I think we are closed to 10 threads by client ... :'(
This clearly exceed the OS limitation , you could try to increase the limit (I believe this is possible on linux with `ulimit`).
But clearly LeshanClient seems not really adapted to this :/

madmax88:
Thanks for the prompt replies @sbernard31, I really appreciate it. I'm closing the request because its clear I need to rethink how I'm approaching this problem, and this doesn't belong here.

Again, thank you for your help:)

PR #135, opened 2016-05-30
Add  API to observe registration life cycle at client side

sbernard31:


leshan-client-core/src/main/java/org/eclipse/leshan/client/observer/LwM2mClientObserver.java
```diff
@@ -0,0 +1,51 @@
...
+    void updateFailed(DmServerInfo server, ResponseCode responseCode, String errorMessage);
+
+    void updateTimeout(DmServerInfo server);
+
+    void deregisterSucceed(DmServerInfo server, String registrationID);
+
+    void deregisterFailed(DmServerInfo server, ResponseCode responseCode, String errorMessage);
+
+    void deregisterTimeout(DmServerInfo server);
+}
```
sophokles73: Most handlers use a naming pattern for methods following the `onSomeEvent` pattern. What do you think about using this here as well, e.g. `onBootstrapSucceeded` and `onRegistrationFailed` etc?

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
```diff
@@ -98,18 +103,33 @@ public LeshanClient(final String endpoint, final InetSocketAddress localAddress,
         // Create sender
         requestSender = new CaliforniumLwM2mClientRequestSender(secureEndpoint, nonSecureEndpoint);
 
-        // Create registration engine
-        bootstrapHandler = new BootstrapHandler(this.objectEnablers);
-        bootstrapHandler.addBootstrapListener(new BootstrapListener() {
```
msangoi: The BootstrapListener is probably not used anymore. We should remove it and rely on the LwM2mClientObserver events.

leshan-client-core/src/main/java/org/eclipse/leshan/client/observer/LwM2mClientObserver.java
sbernard31: I will do that.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
sbernard31: I will do that too.

PR #146, opened 2016-06-20
Northbound Cluster API

sbernard31:
(work in progess)

### DIFF
--- a/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisTokenHandler.java
+++ b/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisTokenHandler.java
@@ -26,8 +26,8 @@ import redis.clients.jedis.Jedis;
 import redis.clients.util.Pool;
 
 /**
- * Only one cluster instance can be responsible of a given LWM2M client. (This restriction is mainly due to the DTLS
- * session)</br>
+ * Only one cluster instance can be responsible of a given LWM2M client at a given moment. (This restriction is mainly
+ * due to the DTLS session)</br>
  * This class store the couple Cluster instance / LwM2M client in a Redis Store.</br>
  * Each Cluster instance is identify by a unique UI and each device by its endpoint.
  */
--- a/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
+++ b/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
@@ -48,11 +48,11 @@ public class LeshanClusterServer {
         Options options = new Options();
 
         options.addOption("h", "help", false, "Display help information.");
-        options.addOption("n", "instanceID", true, "Set the ID of this instance in the cluster.");
+        options.addOption("n", "instanceID", true, "Set the unique identifier of this instance in the cluster.");
         options.addOption("lh", "coaphost", true, "Set the local CoAP address.\n  Default: any local address.");
         options.addOption("lp", "coapport", true, "Set the local CoAP port.\n  Default: 5683.");
-        options.addOption("slh", "coapshost", true, "Set the secure local CoAP address.\nDefault: any local address.");
-        options.addOption("slp", "coapsport", true, "Set the secure local CoAP port.\nDefault: 5684.");
+        options.addOption("slh", "coapshost", true, "Set the local secure CoAP address.\nDefault: any local address.");
+        options.addOption("slp", "coapsport", true, "Set the local secure CoAP port.\nDefault: 5684.");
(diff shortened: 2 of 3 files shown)


leshan-server-cluster/src/main/java/org/eclipse/leshan/server/demo/cluster/serialization/ClientUpdateSerDes.java
```diff
@@ -0,0 +1,76 @@
...
+
+import org.eclipse.leshan.LinkObject;
+import org.eclipse.leshan.server.client.ClientUpdate;
+
+import com.eclipsesource.json.Json;
+import com.eclipsesource.json.JsonArray;
+import com.eclipsesource.json.JsonObject;
+
+/**
+ * Functions for serialize and deserialize a ClientUpdate in JSON for storage.
```
jvermillard: storage?

jvermillard:
some comments:

1. I would like to have a topic prefix, like "leshan." so we can namespace all the leshan PUB/SUB traffic, because the broker is probably shared for multiple systems.
2. Having some unique abstraction called "EventDispatcher" or something alike would make it easier to make pluggable implementation of the dispatcher using different technology

balsmn:
I had took a look in to this PR. If I understand correctly, the JSON request message structure passed via Redis looks like 

```
{
    "ticket": xxxx
    "ep": xxxxx
    "req": {
            ... json representation of LWM2M DownlinkRequest ....
   }
}
```
I was wondering whether you also considered multi-tenancy for the northbound API or may be you have it for the future PRs.

sbernard31:
The documentation is available [here](https://github.com/eclipse/leshan/wiki/Cluster#redis).
There is a discussion about it on the [mailing list](https://dev.eclipse.org/mhonarc/lists/leshan-dev/msg00562.html)

About multi-tenancy did you see any specific point in the implementation which could be a problem for multi-tenancy architecture ?

balsmn:
Yes, I think so. At first glance, the tenant meta-data, say tenant id is not part of the request/response. But thats something that can be added later too. Not a big issue.

Part of Multi-tenancy is also to ensure that the response for a request doesn't reach a wrong tenant. As Redis pub/sub has no relation to key space, we have only limited options. For example, this could be achieved having separate channel prefix for each tenant. As far as I can see, the channels defined (for REQUEST_CHANNEL and RESPONSE_CHANNEL) in a generic way without any tenant relation. right?

sbernard31:
You're right. I think multi-tenancy is cleary not in the scope for now.

ming2377:
Really hope this pull-request will be processed soon... Thanks in advance.

sbernard31:
This PR is awaiting for a review, but keep in mind this is still an experimental feature.

sbernard31:
I rebased it on master.

ming2377:
Thanks for the rebase.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisTokenHandler.java
```diff
@@ -0,0 +1,82 @@
...
+import org.eclipse.leshan.server.client.Client;
+import org.eclipse.leshan.server.client.ClientRegistryListener;
+import org.eclipse.leshan.server.client.ClientUpdate;
+import org.eclipse.leshan.util.Validate;
+
+import redis.clients.jedis.Jedis;
+import redis.clients.util.Pool;
+
+/**
+ * Only one cluster instance can be responsible of a given LWM2M client. (This restriction is mainly due to the DTLS
```
jvermillard: nly one cluster instance can be responsible of a given LWM2M client. ""at a given moment"" ?

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisTokenHandler.java
```diff
@@ -0,0 +1,82 @@
...
+ * session)</br>
+ * This class store the couple Cluster instance / LwM2M client in a Redis Store.</br>
+ * Each Cluster instance is identify by a unique UI and each device by its endpoint.
```
jvermillard: identified

leshan-server-cluster/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jvermillard: should we let the user bring his own log config y?

sbernard31:
This is now integrated in master but this is still very experimental and maybe will not be deliver in the first release of Leshan.

jvermillard:
R.I.P. my review :sob:

PR #147, opened 2016-06-22
Cluster ready Queue Mode

balsmn:
This PR aims to make Eclipse Leshan QueueMode cluster capable and also tries to simplify the `LwM2mRequestSender `API. The API change is necessary to achieve QueueMode clustering capability.

## LwM2mRequestSender
The new `LwM2mRequestSender ` is a listener based instead of callbacks. Callbacks are not cluster friendly. All the previous send methods using callbacks are marked deprecated. It also has two new methods to add/remove ResponseListener. A `ResponseListener `gets notified when a response or error response is received from a LWM2M client. Each request and its response is correlated by means of a RequestTicket. 

There is now a concrete implementation `LwM2mRequestSenderImpl `of the `LwM2mRequestSender` interface which is aware of the client binding mode and delegates the call to either to a `QueuedRequestSender `or to a `CaliforniumLwM2mRequestSender`.  `LwM2mRequestSenderImpl `not only processes requests but also invokes `ResponseListener ` based on clients binding mode. 

### RequestTicket
RequestTicket has to be generated by the consumer of `LwM2mRequestSender `interface. The consumer is also responsible for making sure that the RequestTicket is cluster wide unique.

## QueuedRequestSender
`QueuedRequestSender `doesn't implement the deprecated methods of `LwM2mRequestSender ` and is also not aware of clients binding mode. This responsiblity is now moved to `LwM2mRequestSenderImpl ` as described above.


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -44,14 +46,15 @@
     // A map which contains all pending CoAP requests
     // This is mainly used to cancel request and avoid retransmission on de-registration
     private final ConcurrentNavigableMap<String/* registrationId#requestId */, Request /* pending coap Request */> pendingRequests = new ConcurrentSkipListMap<>();
+    private final ConcurrentLinkedQueue<ResponseListener> responseListeners = new ConcurrentLinkedQueue<ResponseListener>();
```
jvermillard: Diamond operator: ConcurrentLinkedQueue<>();

balsmn: Done. Also did the same in LwM2mRequestSenderImpl. I will push it along with any other changes/comments together.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -80,34 +83,57 @@ public CaliforniumLwM2mRequestSender(final Set<Endpoint> endpoints, final Observ
...
+    /*
+     * (non-Javadoc)
+     * 
+     * @see org.eclipse.leshan.server.request.LwM2mRequestSender#send(org.eclipse.leshan.server.client.Client,
+     * java.lang.String, org.eclipse.leshan.core.request.DownlinkRequest)
+     */
+    @SuppressWarnings("unchecked") // Java generic usage and strict type has to be removed from Leshan
+    @Override
+    public <T extends LwM2mResponse> void send(final Client registrationInfo, final String requestTicket,
+            final DownlinkRequest<T> request) {
```
balsmn: requestTicket is not yet passed to the CoAP layer. We might need it when we also want to use it for correlating with an observation token.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/QueueModeLeshanServer.java
```diff
@@ -144,19 +142,21 @@ public void destroy() {
     @Override
     public <T extends LwM2mResponse> T send(final Client destination, final DownlinkRequest<T> request)
             throws InterruptedException {
-        return queueRequestSender.send(destination, request, sendTimeout);
+        throw new UnsupportedOperationException(
+                "This is a deprecated message and better not use it for testing purposes");
```
sophokles73: I am not sure what to make of this message
> better not use it for testing purposes

Maybe something along the lines of

*Sender supports asynchronous requests only*

The fact that this method cannot be used is already clear because of the exception, isn't it?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
sophokles73: Also, I think you should use the abstract interface instead of the concrete class when declaring the field, i.e.
```
private final Queue<ResponseListener> responseListeners = new ConcurrentLinkedQueue<>();
```

unless you are referring to some non-Queue-defined methods ...

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
```diff
@@ -61,16 +64,16 @@ public CaliforniumLwM2mRequestSender(final Set<Endpoint> endpoints, final Observ
     }
 
     @Override
-    public <T extends LwM2mResponse> T send(final Client destination, final DownlinkRequest<T> request, Long timeout)
-            throws InterruptedException {
+    public <T extends LwM2mResponse> T send(final Client registrationInfo, final DownlinkRequest<T> request,
```
sophokles73: I think we should stick with the original name of the parameter but (soon) change the name of `Client` to `RegistrationInfo` instead. The parameter object in this case still plays the role of the *destination*, doesn't it?

leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/LwM2mRequestSenderImpl.java
```diff
@@ -0,0 +1,170 @@
...
+ * Client binding mode aware request sender. i.e: provides a wrapper delegating the actual invocation to either
+ * {@link QueuedRequestSender} or a default {@link LwM2mRequestSender} depending upon whether the client has connected
+ * in Queue mode or not.
+ */
+public class LwM2mRequestSenderImpl implements LwM2mRequestSender, Stoppable {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LwM2mRequestSenderImpl.class);
+    private final ExecutorService processingExecutor = Executors
+            .newCachedThreadPool(new NamedThreadFactory("leshan-response-processingExecutor-%d"));
+    private final ConcurrentLinkedQueue<ResponseListener> responseListeners = new ConcurrentLinkedQueue<ResponseListener>();
```
sophokles73: use abstract type for declaration?

leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/LwM2mRequestSenderImpl.java
```diff
@@ -0,0 +1,170 @@
...
+    public <T extends LwM2mResponse> T send(final Client registrationInfo, final DownlinkRequest<T> request,
+            final Long timeout) throws InterruptedException {
+        return defaultRequestSender.send(registrationInfo, request, timeout);
+    }
+
+    @SuppressWarnings("deprecation")
+    @Override
+    public <T extends LwM2mResponse> void send(final Client registrationInfo, final DownlinkRequest<T> request,
+            final ResponseCallback<T> responseCallback, final ErrorCallback errorCallback) {
+        if (registrationInfo.getBindingMode().equals(BindingMode.UQ)) {
```
sophokles73: Introduce a `Client.usesQueueMode()` method or something like that instead of traversing the inner structure of Client...

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestFactoryImpl.java
```diff
@@ -64,10 +67,20 @@ public long getRequestId() {
...
+        @Override
+        public String getRequestTicket() {
+            return this.requestTicket;
+        }
+
         @Override
         public String toString() {
-            return "QueuedRequestImpl [downlinkRequest=" + downlinkRequest + ", endpoint=" + endpoint + ", requestId="
-                    + requestId + "]";
+            return "QueuedRequestImpl [requestTicket=" + requestTicket + ", downlinkRequest=" + downlinkRequest
```
sophokles73: better use a StringBuilder?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
```diff
@@ -89,91 +87,148 @@ public static Builder builder() {
...
+    @SuppressWarnings("deprecation")
     @Override
-    public <T extends LwM2mResponse> T send(Client destination, DownlinkRequest<T> request, Long requestTimeout)
-            throws InterruptedException {
-        // Send without callbacks are intended to be synchronous calls and will not be queued.
-        // using the delegate sender to send it immediately.
-        return delegateSender.send(destination, request, requestTimeout);
+    public <T extends LwM2mResponse> T send(final Client registrationInfo, final DownlinkRequest<T> request,
+            final Long requestTimeout) throws InterruptedException {
+        throw new UnsupportedOperationException("QueueMode doesn't support sending sending of messages synchronously");
```
sophokles73: "sending sending" ..

sophokles73:
Is there a particular reasons why we can't add the *request ticket* property to the `DownlinkRequest` class and instead pass it along as a separate parameter in all of the sender's `send()` methods?

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/QueueModeLeshanServer.java
balsmn: Yes, you are right.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
balsmn: I thought of preparing here for the name changes we discussed. But here, yes, it still represents a destination. I can revert it.

leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/LwM2mRequestSenderImpl.java
balsmn: Yes, Good idea. Sounds clean.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
balsmn: Done.

leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/LwM2mRequestSenderImpl.java
balsmn: Done.

leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/LwM2mRequestSenderImpl.java
balsmn: Done.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestFactoryImpl.java
balsmn: Yep, missed it. Done.

balsmn:
@sophokles73 

> Is there a particular reason why we can't/shouldn't add the request ticket property to the DownlinkRequest class and instead pass it along as a separate parameter in all of the sender's send() methods?

Doing this means, breaking the existing send API and fixing all the integration tests which I think @sbernard31 said is not a good idea to do now.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java
sbernard31: Currently in Leshan, we get notifications through the `ObservationRegistry`. So no ticket for now. We will see later if we remove ObservationRegistry and get all notification through `ResponseListener` API. (One step at a time)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestFactoryImpl.java
sbernard31: In the leshan we try to use String.format.
See [Leshan Code rules](https://github.com/eclipse/leshan/wiki/Code-&-design-guidelines).
_"toString should be implemented using String format"_

sbernard31:
I think we should not add the ticket to  at `DownlinkRequest` because the ticket notion make sense only at server side and not at client side or bootstrap server side.

sbernard31:
I will extract the modification about the new send API (with ticket) and but the implementation for queuemode in another branch to allow you to continue to work on it.

leshan-server-core/src/main/java/org/eclipse/leshan/server/client/Client.java
```diff
@@ -171,33 +172,33 @@ public InetSocketAddress getRegistrationEndpointAddress() {
             return null;
         }
 
-        LinkObject[] res = Arrays.copyOf(objectLinks, objectLinks.length);
+        final LinkObject[] res = Arrays.copyOf(objectLinks, objectLinks.length);
```
sbernard31: Please do not add final keyword everywhere.
In Leshan we flavour `final` for class attributes only.

sbernard31:
So the new **send API** with ticket is available in [northbound](https://github.com/eclipse/leshan/tree/northbound) branch.
I put code about cluster support for queue mode in [cluster_queue_mode](https://github.com/eclipse/leshan/tree/cluster_queue_mode) branch.
I also remove `final` keyword on variables and method parameters (I keep it for class attributes).

2 questions:

 1.  in QueueRequest there is requestTicket and requestID ? Should we need both ?
 2. Now ticket id is chosen by user should we still have a QueueRequestFactory ?

balsmn:
@sbernard31 I am a bit confused now. Work on the new cluster_queue_mode means, I should open another PR for further changes and leave this PR?

sbernard31:
Yes exactly, I propose you open another PR based on the cluster_queue_mode](https://github.com/eclipse/leshan/tree/cluster_queue_mode) branch.
Are you ok with that ? :)

balsmn:
I am ok with working on the new branch and Thanks for removing all the final from variables and methods. I was not aware of the coding-convention. I will keep a watch for it.

@sbernard31 I am trying to understand how we would move forward. Now that we have two branches northbound and cluster_queue_mode, both can't be treated separately. i.e:- you can't merge northbound in to master as it will break the queue_mode implementation. For that you need cluster_queue_mode. Are you planning to bring both the branches in to master together?

balsmn:
Regarding your two questions.
>> 1. in QueueRequest there is requestTicket and requestID ? Should we need both ?
Actually not. Thats exactly what I wanted to do next.
>> 2. Now ticket id is chosen by user should we still have a QueueRequestFactory ?
Probably not. I will look in to it. I am happy to have less code.

sbernard31:
I create 2 branches to allow you to continue to work on queue_mode while I continue on northbound API.
I don't see any problem to integrate this 2 branches separately.
It's not a problem for me if we have northbound API without QueueMode or QueueMode without northbound API.

balsmn:
Closing this PR to continue work on cluster_queue_mode branch.

PR #152, opened 2016-07-06
Cluster ready Queue Mode

balsmn:
This PR aims to implement the the new send API (with RequestTicket) for QueueMode.

## LwM2mRequestSender

The new `LwM2mRequestSender `is a listener based instead of callbacks. Callbacks are not cluster friendly. All the previous send methods using callbacks are marked deprecated. It also has two new methods to add/remove `ResponseListener`. A `ResponseListener `gets notified when a response or error response is received from a LWM2M client. Each request and its response is correlated by means of a RequestTicket.

There is now a concrete implementation `LwM2mRequestSenderImpl` of the `LwM2mRequestSender` interface which delegates the send call to either to a `QueuedRequestSender `or to a `CaliforniumLwM2mRequestSender`.

## QueuedRequestSender

`QueuedRequestSender `doesn't implement the deprecated methods of `LwM2mRequestSender `and is also not aware of clients binding mode. This responsiblity is now moved to `LwM2mRequestSenderImpl `as described above.


balsmn:
@sbernard31 I am finished with implementing the new send API for QueueMode. Observations are also working as expected. Upon review If there are no further changes required, please go ahead merging the changes in to the cluster_queue_mode branch and the branch itself in to master. Thanks.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
```diff
@@ -192,7 +188,12 @@ public void onError(String clientEndpoint, String requestTicket, Exception excep
         };
     }
 
-    private void processResponse(String clientEndpoint, String requestTicket, final LwM2mResponse response) {
+    private QueuedRequestImpl newQueueRequestEntity(String endpoint, DownlinkRequest<LwM2mResponse> request,
```
sbernard31: I think this method could be removed.

sbernard31:
Some remarks about this PR/branch (clustering and queue mode) :

 1. Should we [delete request from queue on exception](https://github.com/eclipse/leshan/blob/cluster_queue_mode/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java#L201) ?
 2. We must use `usesQueueMode()` instead of `getBindingMode().equals(BindingMode.UQ)`  in LwM2mRequestSenderImpl .
 3. About `LwM2mRequestSenderImpl.createBindingAwareResponseListener()`, I think this is not the responsability to LwM2mSender to listen response and to store listeners it must delegate it to its internal implementation `queuedRequestSender` and `defaultRequestSender`.

From my point of view, after we are done with clustering and queue mode the next steps will be :

1. Avoid duplication code from `LeshanServer` to `QueueModeLeshanServer`. I think we should remove `QueueModeLeshanServer` and change `LeshanServer` to allow to use custom request sender. 
2. Avoid duplication code from `LeshanClient` to `QueuedModeLeshanClient` by adding a protected `createObjectResource` method in `LeshanClient`.
3. Do not use fixed/static port for QueueMode tests. Use free port like the others tests.

Each of those last 3 points could be done in separate PR to facilitate discussion.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
balsmn: Done.

balsmn:
> 1. Should we delete request from queue on exception ?

Except TimeoutException, I don't see any reason to keep it in the queue as it is up to the application layer to decide further actions which could also be a requeue. 

> 2. We must use usesQueueMode() instead of getBindingMode().equals(BindingMode.UQ) in LwM2mRequestSenderImpl .
Done.

balsmn:
hi Simon,

About the third question, on first look your question seems very logical and valid. But there is a reason why it is done the other way. 
> About LwM2mRequestSenderImpl.createBindingAwareResponseListener(), I think this is not the responsability to LwM2mSender to listen response and to store listeners it must delegate it to its internal implementation queuedRequestSender and defaultRequestSender

To facilitate discusson on your third question,  I have put forth two sequence diagrams. 

The following one shows the current state of implementation in which the LwM2mRequestSenderImpl has a single responsibility of delegating the calls (add/remove response listeners and also dispatching response/errors) based on client's binding mode **(Binding Aware responsiblity)**.

Request-Response-Processing

The next one is as per your suggestion (just delegating add/remove listener calls to concrete implementation), in which case the CaliforniumLwM2mRequestSender, on error should check whether client is in queue mode and the exception is not a TimeoutException before notifying the listeners. Because if a client is in queue mode and the exception is a TimeoutException, the Queue mode will eventually retry.

Request-Response-Processing-Alternative

As you see, now CaliforniumLwM2mRequestSender also takes on the responsiblity to check whether the client is in queue mode. So the **Binding Aware Responsibility** is split across CaliforniumLwM2mRequestSender and LwM2mRequestSenderImpl which I don't think is good.

What do you think?

sbernard31:
Ok I see the problem.
But from my point of view, the first solution has several issues.
1. `LwM2mRequestSenderImpl` should not know how `QueueRequestSender` is implemented. So this class should not know that all responses will be received by `defaultRequestSender`. So it should listens the 2 senders.
2.`QueueRequestSender`  implements `LwM2mRequestSender` but does not respect the "contract" as you can not add or remove listener on it.

**So what should we do ?**  
I propose to change `LwM2mRequestSenderImpl` to make it only depend of `LwM2mRequestSender` not `QueueRequestSender` and delegate add/remove listener to delegate senders.
To fix the issue you raised about timeoutException, I suppose this must be handle in  `QueueRequestSender`.  In this case you must implement something like what you do in `LwM2mRequestSenderImpl`. You could even remove the access of `ClientRegistry` and the check about queue mode if the `QueueRequestSender` use its own instance of `CaliforniumLwM2mRequestSender`.

balsmn:
Ok, I see your point. I have two follow up questions.

1. As per your proposal, still in method `LwM2mRequestSenderImpl.send()`, we need to check whether the client is in QueueMode and not delegate to all the senders which when done will send the message twice. Checking for client binding and TimeoutException on error can be done in `QueueRequestSender `as you said. right?
2. Also, I am not comfortable with exposing the implementation details of `LwM2mRequestSenderImpl `to the application, which, in order to use `LwM2mRequestSenderImpl `should instantiate a `QueueRequestSender` and a `CaliforniumLwM2mRequestSender`. One purpose of `LwM2mRequestSenderImpl ` can also be to remove the burden of depending on the concrete implementations from the application. Is it not? A solution to this might be that the `LwM2mRequestSenderImpl ` instantiates `QueueRequestSender` and `CaliforniumLwM2mRequestSender`.

balsmn:
hi Simon,

I tested your proposal (except for instantiating the CaliforniumLwM2mRequestSender in `QueueRequestSender`, instead passed a second instance of `CaliforniumLwM2mRequestSender` to `QueueRequestSender` from my demo application layer). It looks good. Now `QueueRequestSender` is also fully compliant with the `LwM2mRequestSender` contract and `LwM2mRequestSenderImpl` just delegates the add/remove listeners to both the delegate senders and doesn't register itself as listeners with any of the delegate senders. See my latest commit.

Still would be good if you can answer my questions above. Thanks.

sbernard31:
1. Yes: _"Checking for client binding and TimeoutException on error can be done in QueueRequestSender"_
2. I think `QueueRequestSender` should not be depend of Californium.
3. I'm not sure to understand the problem with `LwM2mRequestSenderImpl` ? I suppose this should be the default implementation, so the complex instantiation will be done by `LeshanServerBuilder`. If some one want to tweak sender behavior by implementing its own sender, it's ok for me if it's a bit complex.

balsmn:
Ok. Good. Then lets leave the `LwM2mRequestSenderImpl` as it now (compared to the very first commit in this branch, it is now very lean). 

Yes, as you say, we could leave the complex instantiation to LeshanServerBuilder.

If you are ok with the state here, please do merge and I can proceed with the next PR's as you suggested above (test/util cleanups).

IMO, we need one more PR's. Currently `LeshanServer` is using `CaliforniumLwM2mRequestSender.cancelPendingRequests()` to cancel requests on the CoAP layer when a device is deregistered. I would like to move this method to `LwM2mRequestSender` interface. Then QueueRequestSender can also use this feature and I can also integrate `LwM2mRequestSenderImpl `in LeshanServer instead of `CaliforniumLwM2mRequestSender`. What do you think?

leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/LwM2mRequestSenderImpl.java
```diff
@@ -101,65 +75,20 @@ public LwM2mRequestSenderImpl(LwM2mRequestSender defaultRequestSender, QueuedReq
...
-                        processingExecutor.execute(new ResponseProcessingTask(clientEndpoint, requestTicket,
-                                responseListeners, exception));
-                    }
-                }
-            }
-        };
-
+        if (queuedRequestSender instanceof Stoppable) {
+            ((Stoppable) queuedRequestSender).stop();
+        }
```
sbernard31: we should also stop defaultRequestSender (if this is an instance of Stoppable)

sbernard31:
Except my last comment, I think we are ok to merge it. I will do that as soon as you fixed that.

I agree about the fourth PR.

leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/LwM2mRequestSenderImpl.java
balsmn: Thought of it and left it as it is not an instance of Stoppable. No problem. I can add it.

balsmn:
hi Simon, 

I just pushed your last request. Thanks for looking in to it so fast.

As the 3 further PRs (test/util cleanup) and the 4th PR (cancelling of request on coap layer) related to Queue mode in general and not to clustering of queue mode, I would like to proceed on it with master as the base.

Could you also merge this branch to master? or should I open a separate PR for that?

sbernard31:
Pushed on master (commit b91f346e3f01d4d8d07cc84e926fb6dd4d42c411)

PR #157, opened 2016-07-28
InMemoryMessageStore is not thread save

ingoschaal:
Fixed a check-then-act situation in InMemoryMessageStore at add(final QueuedRequest entity)...
Also added some Junit Test for InMemoryMessageStore.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
@@ -39,17 +39,17 @@ public class InMemoryMessageStore implements MessageStore {
     private final ConcurrentMap<String, BlockingQueue<QueuedRequest>> requestQueueMap = new ConcurrentHashMap<>();
 
     @Override
-    public void add(final QueuedRequest entity) {
+    public void add(QueuedRequest entity) {
         LOG.debug("Add entity {}", entity);
-        final String endpoint = entity.getEndpoint();
-        final BlockingQueue<QueuedRequest> requestQueue = new LinkedBlockingQueue<>();
+        String endpoint = entity.getEndpoint();
+        BlockingQueue<QueuedRequest> requestQueue = new LinkedBlockingQueue<>();
 
         // If two Threads want to add a entity at the exact instant, each will
         // retrieve its own instance of
         // Queue and add the entity to the respective instances. But only one of
         // them wins the putIfAbsent method.
         // To not lose messages, this check will prevent.
-        final BlockingQueue<QueuedRequest> allreadyExisting = requestQueueMap.putIfAbsent(endpoint, requestQueue);
+        BlockingQueue<QueuedRequest> allreadyExisting = requestQueueMap.putIfAbsent(endpoint, requestQueue);
         if (allreadyExisting != null) {
             allreadyExisting.add(entity);
         } else {
@@ -58,25 +58,27 @@ public class InMemoryMessageStore implements MessageStore {
     }
 
     @Override
-    public boolean isEmpty(final String endpoint) {
-        if (queueExist(endpoint)) {
-            return requestQueueMap.get(endpoint).isEmpty();
+    public boolean isEmpty(String endpoint) {
+        LOG.trace("Checking for empty Queue {}", endpoint);
+        BlockingQueue<QueuedRequest> requests = requestQueueMap.get(endpoint);
+        if (requests != null) {
+            return requests.isEmpty();
         }
         return true;
-
     }
 
     @Override
-    public QueuedRequest retrieveFirst(final String endpoint) {
+    public QueuedRequest retrieveFirst(String endpoint) {
         LOG.trace("Retrieve first for endpoint {}", endpoint);
-        if (queueExist(endpoint)) {
-            return requestQueueMap.get(endpoint).peek();
+        BlockingQueue<QueuedRequest> requests = requestQueueMap.get(endpoint);
+        if (requests != null) {
+            return requests.peek();
         }
         return null;
     }
 
     @Override
-    public void removeAll(final String endpoint) {
+    public void removeAll(String endpoint) {
         LOG.debug("Emptying messages for client {}", endpoint);
         // If client has registers and de-registers without any other messages
         // then the queue would not have been initialized.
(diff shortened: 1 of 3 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
```diff
@@ -39,40 +39,54 @@
     private final ConcurrentMap<String, BlockingQueue<QueuedRequest>> requestQueueMap = new ConcurrentHashMap<>();
 
     @Override
-    public void add(QueuedRequest entity) {
+    public void add(final QueuedRequest entity) {
```
sbernard31: pls, remove all unnecessary final keyword.
(see https://github.com/eclipse/leshan/wiki/Code-&-design-guidelines)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
```diff
@@ -39,40 +39,54 @@
...
+        // retrieve its own instance of
+        // Queue and add the entity to the respective instances. But only one of
+        // them wins the putIfAbsent method.
+        // To not lose messages, this check will prevent.
+        final BlockingQueue<QueuedRequest> allreadyExisting = requestQueueMap.putIfAbsent(endpoint, requestQueue);
+        if (allreadyExisting != null) {
+            allreadyExisting.add(entity);
+        } else {
+            requestQueue.add(entity);
+        }
```
sbernard31: it seems better :)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
```diff
@@ -39,40 +39,54 @@
...
 
     @Override
-    public boolean isEmpty(String endpoint) {
-        return getMessageQueueForEndpoint(endpoint).isEmpty();
+    public boolean isEmpty(final String endpoint) {
+        if (queueExist(endpoint)) {
+            return requestQueueMap.get(endpoint).isEmpty();
+        }
+        return true;
+
```
sbernard31: It seems that we can get NPE if between `if` and `return` the value for `endpoint` is removed from the map.
So I prefer something like : 
```java
  BlockingQueue<QueuedRequest> requests = requestQueueMap.get(endpoint);
  if (requests != null){
      requests.isEmpty();
  }
  return true;
```

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
```diff
@@ -39,40 +39,54 @@
...
     @Override
-    public QueuedRequest retrieveFirst(String endpoint) {
+    public QueuedRequest retrieveFirst(final String endpoint) {
         LOG.trace("Retrieve first for endpoint {}", endpoint);
-        BlockingQueue<QueuedRequest> requests = getMessageQueueForEndpoint(endpoint);
-        return requests.peek();
+        if (queueExist(endpoint)) {
+            return requestQueueMap.get(endpoint).peek();
+        }
+        return null;
```
sbernard31: same as above

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
```diff
@@ -39,40 +39,54 @@
...
-        }
+        requestQueueMap.remove(endpoint);
     }
 
     @Override
-    public void deleteFirst(String endpoint) {
+    public void deleteFirst(final String endpoint) {
         LOG.debug("Delete first entity of endpoint {}", endpoint);
-        BlockingQueue<QueuedRequest> requests = requestQueueMap.get(endpoint);
+        final BlockingQueue<QueuedRequest> requests = requestQueueMap.get(endpoint);
```
sbernard31: same as above

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
```diff
@@ -84,19 +98,26 @@ public void deleteFirst(String endpoint) {
...
+            return Collections.emptyList();
         }
     }
 
-    private BlockingQueue<QueuedRequest> getMessageQueueForEndpoint(String endpoint) {
-        return requestQueueMap.get(endpoint) != null ? requestQueueMap.get(endpoint)
-                : new LinkedBlockingQueue<QueuedRequest>();
+    private boolean queueExist(final String endpoint) {
+        return requestQueueMap.get(endpoint) != null ? true : false;
+    }
```
sbernard31: we can remove this method (if my remarks above make sense)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
```diff
@@ -84,19 +98,26 @@ public void deleteFirst(String endpoint) {
      * @param endpoint client's endpoint
      * @return list of queue request entities in order of processing.
      */
-    public List<QueuedRequest> retrieveAll(String endpoint) {
+    public List<QueuedRequest> retrieveAll(final String endpoint) {
         LOG.debug("Retrieve all for endpoint {}", endpoint);
-        BlockingQueue<QueuedRequest> requests = getMessageQueueForEndpoint(endpoint);
-        if (requests.isEmpty()) {
-            return Collections.emptyList();
+        if (queueExist(endpoint)) {
```
sbernard31: same as above

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
```diff
@@ -84,19 +98,26 @@ public void deleteFirst(String endpoint) {
...
+        return requestQueueMap.get(endpoint) != null ? true : false;
+    }
+
+    /**
+     * For test onyl.
+     * 
+     * @return ConcurrentMap<String, BlockingQueue<QueuedRequest>>
+     */
+    protected ConcurrentMap<String, BlockingQueue<QueuedRequest>> getRequestQueueMap() {
+        return this.requestQueueMap;
```
sbernard31: I don't really like the idea to expose the requestQueueMap just for testing.
Maybe we can add a getSize() on this class, with retrieveAll.getSize(), we will cover all your needs.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
ingoschaal: yes i will do this of course.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
ingoschaal: Ok, that is  possible, you are right.
My thought was not to have so much duplicate code, therefore the method. I will change this and remove the method.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
ingoschaal: yes, as mentioned above

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/InMemoryMessageStore.java
balsmn: FYI, I have added a similar `getQueueSize(endpoint)` method and removed `retrieveAll(endpoint)` method from InMemoryMessageStore as part of the PR #158.

ingoschaal: OK good point.  Now I added this getQueueSize method, looks good also for testing use.

sbernard31:
Integrated in master commit (ddaa6e4744d35ed8c9afe34a37ef44a85ec76a0f)

PR #158, opened 2016-07-29
Cancelling all pending messages support

balsmn:
`CaliforniumLwM2mRequestSender `implementation had this method (cancelPendingMessages) and this concrete implementation was directly used in LeshanServer. This prevented from using the new `LwM2mRequestSender` API (containing support for both queue and non-queue mode) in Leshan Server. This PR aims to add the `cancelPendingMessages` to `LwM2mRequestSender` API and implements the same in all the concrete implementations (`LwM2mRequestSenderImpl`, `QueuedRequestSender` and `CaliforniumLwM2mRequestSender`). With this PR, Leshan Server will officially have the QueueMode support.

Signed-off-by: Bala Azhagappan <balasubramanian.azhagappan@bosch-si.com>

PR #160, opened 2016-08-09
Queue mode sender concurrency fix

balsmn:
This PR is based on the source of #158 (contains the first three commits) and aims to fix the concurrency problem in

(https://github.com/eclipse/leshan/blob/master/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java#L124

 if we assume two threads t1 (user thread that calls `LwM2mRequestSender.send()`) and t2 (another user thread that calls `cancelPendingMessage()`):
- t1 passes the check mentioned above, but request was not added to messageStore yet
- t2 removes all request from queue (could be because client was unregistered)
- t1 adds the request to message store (we have a request in the store without a corresponding registered client, i.e. this request may stay there forever)

PS: I will update this PR if #158 is accepted.

Signed-off-by: Bala Azhagappan balasubramanian.azhagappan@bosch-si.com

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
@@ -69,7 +69,7 @@ public class QueuedRequestSender implements LwM2mRequestSender, Stoppable {
     private final ClientStatusTracker clientStatusTracker;
     private final Collection<ResponseListener> responseListeners = new ConcurrentLinkedQueue<>();
 
-    private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock(false);
+    private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
 
     /**
      * Creates a new QueueRequestSender using given builder.
@@ -91,6 +91,11 @@ public class QueuedRequestSender implements LwM2mRequestSender, Stoppable {
         delegateSender.addResponseListener(createResponseListener());
     }
 
+    /**
+     * fluent api builder to construct an instance of {@link QueuedRequestSender}
+     * 
+     * @return instance of {@link QueuedRequestSender.Builder}
+     */
     public static Builder builder() {
         return new Builder();
     }
@@ -120,7 +125,7 @@ public class QueuedRequestSender implements LwM2mRequestSender, Stoppable {
 
         readWriteLock.readLock().lock();
         try {
-            // Accept messages only when the client is already known to ClientRegistry
+            // Check whether client is still known to ClientRegistry
             if (clientRegistry.get(endpoint) != null) {
                 QueuedRequest queuedRequest = new QueuedRequestImpl(endpoint, castedDownlinkRequest, requestTicket);
                 messageStore.add(queuedRequest);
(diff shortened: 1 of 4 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
```diff
@@ -67,6 +69,8 @@
     private final ClientStatusTracker clientStatusTracker;
     private final Collection<ResponseListener> responseListeners = new ConcurrentLinkedQueue<>();
 
+    private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock(false);
```
sbernard31: use new ReentrantReadWriteLock()

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
```diff
@@ -114,17 +118,25 @@ public static Builder builder() {
...
+                // If Client is reachable and this is the first message, we send it
+                // immediately.
+                if (clientStatusTracker.startClientReceiving(endpoint)) {
+                    processingExecutor.execute(newRequestSendingTask(endpoint));
+                }
+            } else {
+                String message = String.format("message received in Queue Mode for the unknown client [%s]", endpoint);
+                LOG.warn(message);
+                // notify application layer that the message will not be sent with an exception.
+                throw new RequestCanceledException(message);
```
sbernard31: The canceled notification should not be raised like other errors ? (through `ResponseListener`)

leshan-server-core/src/main/java/org/eclipse/leshan/server/response/ResponseProcessingTask.java
```diff
@@ -89,8 +89,7 @@ private void executeAction() {
         for (ResponseListener listener : responseListeners) {
             if (listener != null) {
                 if (hasException) {
-                    LOG.debug("invoke response listener for requestTicket {} with exception {}", requestTicket,
-                            exception);
+                    LOG.debug("invoke response listener for requestTicket throws exception", requestTicket, exception);
```
sbernard31: Missing curly bracket `{}` for requestTicket ?

DanielMaier-BSI: If the last param is an exception, slf4j prints since version 1.6.0 the stacktrace anyways. See http://www.slf4j.org/faq.html#paramException (exmaple is with error log, but also true for debug statements)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
balsmn: Yes makes sense.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/impl/QueuedRequestSender.java
balsmn: Not a problem

leshan-server-core/src/main/java/org/eclipse/leshan/server/response/ResponseProcessingTask.java
sbernard31: I just talk about requestTicket parameter, not about exception.

balsmn: oh..yes...

balsmn:
Also updated this PR after #158 was closed. Rebase on to master did not work due to conflicts between #158 and #157. I had to merge it with the changes in this PR.

sbernard31:
you shall not pass
Sry we try to keep a clean history, so no merge allowed ;)

Maybe you could try to restart from master and cherry pick the last commit. (This should limits conflict)

balsmn:
Hi Simon,

It worked. Thanks for the tip. I still have some conflicts with master which I think is due to #167 which got merged today. I will rebase on to master again tomorrow and probably have to adapt the implementations too as per new API.

sbernard31:
Integrated in master (commit  	a8df823)

PR #161, opened 2016-08-10
Objlnk support for tlv and text format

DevendraKurre:


sbernard31:
Hi @DevendraKurre,
  Thx for the PR. 
  
  I think we should create a specific class for `ObjectLink` instead of exposing byte[] at API side.

  There is also some formatting issues (mixed tab/spaces). Have a look [here](https://github.com/eclipse/leshan#code-with-eclipse) to configure formatter.

  We don't not accept PR with merge commit, could you please clean your branch history :).

   And last point, You need to sign the [eclipse CLA](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git). 

   I hope all of this is not too scary :)

PR #168, opened 2016-08-22
fixed duplicate observations for the same resource

balsmn:
This PR fixes the issues reported in https://github.com/eclipse/leshan/issues/163 and added few javadocs.

Signed-off-by: Bala Azhagappan <balasubramanian.azhagappan@bosch-si.com>

Issue #163, opened 2016-08-15
Multiple observations on same resource


Christian-Schmid:
Hi,

I may have found an issue with behavior of the current ObservationRegistry implementation.
In earlier versions of leshan (till leshan-0.1.11-M10) it was not possible to have more than one Observation (for the same path) on a client resource. When adding a new observation, any potential existing observation was cancelled: 

https://github.com/eclipse/leshan/blob/leshan-0.1.11-M10/leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/ObservationRegistryImpl.java#L63

Starting M11 this check was removed:

https://github.com/eclipse/leshan/blob/leshan-0.1.11-M11/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumObservationRegistryImpl.java#L71

Therefore now it is possible to have multiple observations on the same resource. 

Was there any specific reason for removing this check? (Having the coap spec in mind, this should still not be allowed: https://tools.ietf.org/html/rfc7641#page-10 ) 

I would appreciate any comment :-)

Many thanks!
Chris

sbernard31:
I can see any good reason why this check was removed. I suppose you find a bug.
We probably must add it again.

(This would be interesting to see how this behave at californium side too)

sbernard31:
is there any chance you provide a patch for this one ?  :)

balsmn:
yes, we will provide one

sbernard31:
Integrated in master (commit aa764a9)

PR #169, opened 2016-08-24
Fixed a race condition while adding and cancelling an observation

balsmn:
This PR fixes a race condition in `CaliforniumObservationRegistryImpl` while invoking `addObservation` and `cancelObservation` calls. Due to this race condition it could still be possible to have a duplicate observation for the same resource and client reported in https://github.com/eclipse/leshan/issues/163

Due to change that came in as a bug fix for issue #168 there can be at most only one active observation for a given resource. So the `ObservationRegistry` interface was changed to reflect the same.

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumObservationRegistryImpl.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumObservationRegistryImpl.java
@@ -64,6 +64,7 @@ public class CaliforniumObservationRegistryImpl
 
     private final List<ObservationRegistryListener> listeners = new CopyOnWriteArrayList<>();
     private final Map<KeyToken, Observation> observationsByToken = new ConcurrentHashMap<>();
+    private final Object monitor = new Object();
 
     /**
      * Creates an instance of {@link CaliforniumObservationRegistryImpl}
@@ -83,12 +84,27 @@ public class CaliforniumObservationRegistryImpl
 
     @Override
     public void addObservation(Observation observation) {
-        // cancel any existing observations for the same path and registration id.
-        cancelObservations(observation.getRegistrationId(), observation.getPath().toString());
+        Observation existingObservation;
+        synchronized (monitor) {
+            // cancel any existing observations for the same path and registration id.
+            existingObservation = getObservation(observation.getRegistrationId(), observation.getPath().toString());
+            if (existingObservation != null) {
+                cancelObservationsOnEndpoints(existingObservation.getId());
+                observationsByToken.remove(new KeyToken(existingObservation.getId()));
+            }
 
-        // add the new observation
-        observationsByToken.put(new KeyToken(observation.getId()), observation);
+            // add the new observation
+            observationsByToken.put(new KeyToken(observation.getId()), observation);
+        }
 
+        // notify listeners if existing observation was cancelled.
+        if (existingObservation != null) {
+            for (ObservationRegistryListener listener : listeners) {
+                listener.cancelled(observation);
+            }
+        }
+
+        // notify about the newly added observation.
         for (ObservationRegistryListener listener : listeners) {
             listener.newObservation(observation);
         }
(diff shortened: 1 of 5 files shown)


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumObservationRegistryImpl.java
```diff
@@ -64,6 +66,7 @@
 
     private final List<ObservationRegistryListener> listeners = new CopyOnWriteArrayList<>();
     private final Map<KeyToken, Observation> observationsByToken = new ConcurrentHashMap<>();
+    private final Lock observationLock = new ReentrantLock();
```
jvermillard: Why us a Lock and not a synchronize(this) ?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumObservationRegistryImpl.java
```diff
@@ -112,52 +133,65 @@ public int cancelObservations(Client client) {
             return 0;
 
         Set<Observation> observations = getObservations(registrationId);
+        int removed = 0;
         for (Observation observation : observations) {
-            cancelObservation(observation);
+            removed = doCancelObservation(observation) ? ++removed : removed;
```
jvermillard: You mean:
```
if (doCancelObservation(observation)) {
   removed++;
}
```

balsmn: yes

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumObservationRegistryImpl.java
balsmn: notifying Observation listeners is outside the locks. At the end the same can also be achieved using synchronized blocks. At the end its just a religious thing which one you chose.

DanielMaier-BSI: I think it is better to use a lock that is private to the class. Otherwise someone else could also synchronize on "this" and we might get a deadlock (more a theoretical and style discussion).

Of course we can also synchronize on a private object and use synchronized blocks.

jvermillard: There is no good reason to use a lock here, you can use a private monitor. "private Object monitor = new Object();" if you really want to isolate it.

A synchronize on monitor is superior here because you don't have to rely on the try/finally and you are not using the extra feature of Lock missing from synchronize block.

That's no religious/style thing, just better idiomatic java :smile:

DanielMaier-BSI: This is what I meant with the last sentence. We can use a private monitor.

balsmn:
@jvermillard as you suggested, I have pushed the updated version with the two changes.

sbernard31:
Manu is modifying this code and your PR will not be so relevant in the new code.

But finally this constraint of duplicate observation should not be handle at Leshan level. This is a CoAP constraint so this should be done in californium.

balsmn:
I agree with you regarding only one observation per resource restriction belonging to californium. The fact that we have an `ObservationStore` in Californium layer and an `ObservationRegistry` in Leshan leads to replicate the californium restriction to Leshan layer. 

Could you give a heads up what kind of changes are to be expected in the new code? Is it only reimplementing `ObservationRegistry` or we would not have an `ObservationRegistry`?

sbernard31:
The Observe CoAP spec [§3.1][1] say:
"A client MUST aggregate such requests and MUST NOT register more than once for the same target resource.  The target resource is identified by all options in the request that are part of the Cache-Key. This includes, for example, the full request URI and the Accept Option."

Currently in californium, there is no limitation about that. An obversation is just identifying by a Token.

The spec say that a target resource is identified by all the options [§5.10][2] that are part of the cache key (= all  options except size1). But, as we deviate a little from the spec by allowing observation from different endpoint (This is the case where the CoAP server is behind a NAT), Using the Cache-Key is not so relevant...

So I'm not so sure pushing this restriction in californium is still a good idea... 

> Could you give a heads up what kind of changes are to be expected in the new code? Is it only reimplementing ObservationRegistry or we would not have an ObservationRegistry?

Manu will probably push a PR before the end of the week.

[1]:https://tools.ietf.org/html/rfc7641#section-3.1
[2]:https://tools.ietf.org/html/rfc7252#section-5.10

sbernard31:
So Manu pushed its code in the [observe_context](https://github.com/eclipse/leshan/commits/observe-request-context) branch which needs this [additionnal feature](https://github.com/eclipse/californium/pull/86) in Californium.

Unfortunately, this will not fix the race condition you find here.
A possible solution could be to let the ObservationStore removing old observation(s?) as it is in charge to keep the store in a consistent state.  Ideally, it should notify Californium that the request is cancelled.

There is another race condition between observe and registration, but we think we found a miss-conception in Leshan. As observation is strongly linked to the registration, we think that observation should be store in ClientRegistry(RegistrationStore)

What do you think ?

balsmn:
Yes I agree that Observation is strongly linked to a registration.

> we think that observation should be store in ClientRegistry(RegistrationStore)

You mean to say that instead of having a separate ObservationRegistry, just have the RegistrationStore implement the [CaliforniumObservationRegistry](https://github.com/eclipse/leshan/blob/master/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/CaliforniumObservationRegistry.java) or even the new [LwM2mObservationStore ](https://github.com/eclipse/leshan/blob/observe-request-context/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LwM2mObservationStore.java) interface introduced in the branch ([observe_context](https://github.com/eclipse/leshan/commits/observe-request-context)) and clear the observations when a de-registration happens?

balsmn:
> Unfortunately, this will not fix the race condition you find here.
A possible solution could be to let the ObservationStore removing old observation(s?) as it is in charge to keep the store in a consistent state. Ideally, it should notify Californium that the request is cancelled.

I agree. If [`InMemoryObservationStore`](https://github.com/eclipse/leshan/blob/observe-request-context/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/InMemoryLwM2mObservationStore.java) while adding, can also cancel any existing observations. That means, the `add(Observation)` of `ObservationStore` should return list of observations removed/canceled while adding.  right? 

When it is implemented that way and merged in to master, then this PR will not be relevant any more, if not it would still be good to have this PR.

I have a small request. As the refactoring done in `CaliforniumObservationRegistry` as part of the observe-request-context has nothing to do with the observe request's extention with a context info, it would be good to separate this refactoring from the actual observe request context implementations when creating PR's. What do you think ?

sbernard31:
>You mean to say that instead of having a separate ObservationRegistry, just have the RegistrationStore implement the CaliforniumObservationRegistry or even the new LwM2mObservationStore interface introduced in the branch (observe_context) and clear the observations when a de-registration happens?

Something like this but I need to think more about this.

msangoi:
> I have a small request. As the refactoring done in CaliforniumObservationRegistry as part of the observe-request-context has nothing to do with the observe request's extention with a context info, it would be good to separate this refactoring from the actual observe request context implementations when creating PR's. What do you think ?

I'm not sure to understand why it is needed to separate this. I agree that the registry refactoring could have been done in a separate PR, but then the code handling the observe context would have been duplicated.

sbernard31:
Manu changes was integrated in master so the PR si no more relevant.
I open the issue to track that this race condition is still present. #176

PR #177, opened 2016-09-21
Added notification of early failure

phtrivier:
The `fail()` API of `BootstrapManager` was not called when the BS failed early (for example, when no  no `BootstrapConfig` exists. 
(This could be useful for logging or reporting.)


sbernard31:
Integrated in master (commits 387bee8bca180333b22782ed6b9ae74730524e7c,18d3271a61df6e8d2e171aae3f9f5d14b0eb9c53)

PR #190, opened 2016-10-18
Avoid to spit error 500 on empty object list

jvermillard:
By initializing the list of object, it permits to be resilient to registration without any objects and to not spit "unexpected errors" even if the device is not supposed to do that.


sbernard31:
Integrated in #241 (commit 348026440e46120b91104a16068294528dfc2f3f)

PR #195, opened 2016-11-07
Add support for TCP.

boaks:
First approach. Requires californium "bugfix_tcp_notify".

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
@@ -12,19 +12,30 @@
  * 
  * Contributors:
  *     Zebra Technologies - initial API and implementation
+ *     Bosch Software Innovations - add TCP endpoints
  *******************************************************************************/
 package org.eclipse.leshan.client.californium;
 
 import java.net.InetSocketAddress;
+import java.security.GeneralSecurityException;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+
 import org.eclipse.californium.core.CoapServer;
 import org.eclipse.californium.core.network.CoapEndpoint;
 import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.core.server.resources.Resource;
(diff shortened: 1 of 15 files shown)


leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
```diff
@@ -153,6 +164,83 @@ protected Resource createRoot() {
         clientSideServer.add(new BootstrapResource(bootstrapHandler));
     }
 
+    public LeshanClient(final boolean tcp, final String endpoint, final InetSocketAddress localAddress,
```
boaks: OK, the client requires more work, But for very first tests, it should do it.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
```diff
@@ -140,6 +154,46 @@ public LeshanServer(final Set<Endpoint> endpoints, final ClientRegistry clientRe
         createServer(endpoints);
     }
 
+    public LeshanServer(final ClientRegistry clientRegistry, final SecurityRegistry securityRegistry,
```
boaks: Though the CoapEndpoint are highly depending on "internal" stuff, creating endpoints outside the LeshanServer is very complex (or impossible).

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
```diff
@@ -162,24 +216,89 @@ public LeshanServer(InetSocketAddress localAddress, InetSocketAddress localSecur
...
+            final SecurityRegistry securityRegistry, final ClientRegistry clientRegistry,
+            final CaliforniumObservationRegistry observationRegistry) {
+
+        // secure endpoint
+        Connector connector = null;
+        NetworkConfig config = NetworkConfig.getStandard();
+        String algorithm = Security.getProperty("ssl.KeyManagerFactory.algorithm");
+
+        try {
+            KeyStore ks = KeyStore.getInstance("JKS");
```
boaks: Next to change ... 
And peer check on registration still missing.

boaks:
Hi,

this PR is not really intended to be merged, but I hope its a base for discussion 
(and may be someone will also use it for early  testing).

So it mainly contains two changes:
- remove explizit "secureEndpoint" and "noSecureEndpoint" and replace it with a "Set<Endpoint>"
  (BootstrapServer right now not done).
  Are there any objections against the set?

- remove the parameter for explizit secure and none secure endpoints and replace them with a list of URI
  (use "*" for any-interface).
  Are there any objections against using a list of URIs?

boaks:
More work done, functionality improved see #203

PR #196, opened 2016-11-07
use slf4j-simple in place of log4j2 for client

jvermillard:
log4j2 is not compatible with JRE embedded
so for the client it's preferable to use the lighter sl4j-simple


sbernard31:
If we do that we should remove the log4j2.xml config file and add a "simplelogger.properties" file.
(http://www.slf4j.org/api/org/slf4j/impl/SimpleLogger.html)

sbernard31:
Integrated in master (commit 18f3063b7b0f6efcb049410f7374aad6dd1b263d)

PR #197, opened 2016-11-07
Cluster improvement

jvermillard:
Implemented in multiple commit most of the improvement suggested in the "northbound api" PR

### DIFF
--- a/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
+++ b/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
@@ -15,8 +15,8 @@
  *******************************************************************************/
 package org.eclipse.leshan.server.cluster;
 
-import java.net.BindException;
 import java.net.URI;
+import java.net.URISyntaxException;
 
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.DefaultParser;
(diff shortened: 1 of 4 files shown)


leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
```diff
@@ -30,30 +30,32 @@
...
+ * 
+ * Will start a LWM2M server which would be part of a large cluster group. This group of cluster node will share state
+ * and communicate using Redis. So they all need too be connected to the same redis server.
+ */
 public class LeshanClusterServer {
-    private static final Logger LOG = LoggerFactory.getLogger(LeshanClusterServer.class);
 
     private final static String USAGE = "java -jar leshan-server-cluster.jar [OPTION]";
-    private final static String FOOTER = "All options could be passed using environment variables.(using long option name in uppercase)";
+    private final static String FOOTER = "Question? Bug report? Eclipse Leshan : https://eclipse.org/leshan";
```
sbernard31: Why removing information about environment variable ?

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
```diff
@@ -30,30 +30,32 @@
...
-        options.addOption("slh", "coapshost", true, "Set the local secure CoAP address.\nDefault: any local address.");
-        options.addOption("slp", "coapsport", true, "Set the local secure CoAP port.\nDefault: 5684.");
-        options.addOption("wp", "webport", true, "Set the HTTP port for web server.\nDefault: 8080.");
+        options.addOption("h", "help", false, "Display this help message.");
+        options.addOption("n", "instanceID", true, "Sets the unique identifier of this instance in the cluster.");
+        options.addOption("l", "coapaddressport", true,
+                "Sets the local CoAP bind address and port in form [IP]:[PORT].\n  Default: 0.0.0.0:5683.");
+        options.addOption("sl", "coapsaddressport", true,
+                "Sets the local secure CoAP bind address and port, same format.\n  Default: 0.0.0.0:5684.");
         options.addOption("r", "redis", true,
```
sbernard31: Using lh, lp, slh, slp was to be homogeneous with LeshanServerDemo, LeshanClientDemo and wakaama.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
```diff
@@ -77,83 +79,93 @@ public static void main(String[] args) {
 
         // Abort if unexpected options
         if (cl.getArgs().length > 0) {
-            System.out.println("Unexpected option or arguments : " + cl.getArgList());
+            System.err.println("Unexpected option or arguments : " + cl.getArgList());
```
sbernard31: I prefer using sys.err for error. I don't consider putting bad option as an error this seems to me a normal use case of a command line tool.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
```diff
@@ -77,83 +79,93 @@ public static void main(String[] args) {
...
             formatter.printHelp(USAGE, null, options, FOOTER);
             return;
         }
 
         // get cluster instance Id
-        String clusterInstanceId = System.getenv("INSTANCEID");
```
sbernard31: Why removing environment variable ?
(I better understand why environment information was removed above ^^)

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
jvermillard: I think the number of option on the clluster node will grow so having 6 options for port/iface mapping was large.
and -slh -slp :grin: looks so simple. 
BTW I did it for reducing the number of port/address options (6) but then I removed the web port only config which is not only inconsistent (no iface config) but also unused. So we can go back to the previous solution if you remove the web port binding option.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
jvermillard: it's not an error but it stop the the processing. for me it's the definition of a fatal error.

jvermillard:
bytebuffer is a standard class (well tested) and wrap the byte[] an provide the needed methods for array.
I think you are prematurely optimizing. If we reimplement a class from the std lib I think we should have strong rationales.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/LeshanClusterServer.java
sbernard31: If we want to remove it. We should also remove it in bsserver-demo and server-demo. We should also change the script on the sandbox which use this envvar.

sbernard31:
Array.hashcode() and Array.equals is standard class too (well tested too).

sbernard31:
I integrated this changes on master 
(commit 37496e72fec9f38e5bcd149433ab217fb34299c2,  0e8518139beddb8dc5b2aea81f6439df7ca95435, ae582948f9800c99e36ec3355ba34d1b5ad792ff, 91b115d5f14914b32ac9bc1f220b8723253bf344,  61b9a9c66ceeddd2f23ffc35f5ace9b9fa46844c )
except for `bytebuffer.wrap`,  removing env var, unified lh/lp and  slh/slp.

ming2377:
Sorry, not clear if this is actually merged into master, seems some conflicts remain. Thanks.

sbernard31:
The comment above seems clear.
All is integrated excepted : bytebuffer.wrap, removing env var, unified lh/lp and slh/slp for reason exposed above.

I think removing env var, unified lh/lp and slh/slp could be done in another PR. (changing this for all Leshan command line tools) I create  issues about this #208, #209

I close this issue. 
Please feel free to reopen it or comment if it's still not clear

PR #203, opened 2016-11-11
Support tcp notifies

boaks:
A small next experimental step towards CoAP over TCP experience.

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
@@ -12,19 +12,30 @@
  * 
  * Contributors:
  *     Zebra Technologies - initial API and implementation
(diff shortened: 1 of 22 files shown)


boaks:
Requires the californium branch "bugfix_tcp_notify", for those who want to try it.

sophokles73:
Is there a relation between adding TCP support and the original changes the `improve_builder` branch has been created for? My guess is that this should be done on a separate branch ...

boaks:
https://dev.eclipse.org/mhonarc/lists/leshan-dev/msg00615.html

`But what if we want to use a secure endpoint for coap and one for coap+tcp?
`

So my impression was, that the branch prepares for coap+tcp.
The first part of the work was to introduce a way to easier add endpoints (URI set), because they are more related to "leshan internals" then it seems from outside.
But, if you then just want to use "coap+tcp", then you will be faced, that the observe/notify needs to be fixed.

`My guess is that this should be done on a separate branch ...`

Cool, so how do I get such a branch?

PR #204, opened 2016-11-22
ContentFormat default value

peteryang:
I tested some LwM2m client for observing /objectid/instid/resourceid

LwM2m device does not set content format. WIth content format to be null, Leshan shows exception

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumObservationRegistryImpl.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumObservationRegistryImpl.java
@@ -268,7 +268,7 @@ public class CaliforniumObservationRegistryImpl implements CaliforniumObservatio
                 LwM2mModel model = modelProvider.getObjectModel(client);
 
                 // get content format
-                ContentFormat contentFormat = null;
+                ContentFormat contentFormat = ContentFormat.TEXT;
                 if (coapResponse.getOptions().hasContentFormat()) {
                     contentFormat = ContentFormat.fromCode(coapResponse.getOptions().getContentFormat());
                 }
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LwM2mResponseBuilder.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LwM2mResponseBuilder.java
@@ -352,7 +352,7 @@ public class LwM2mResponseBuilder<T extends LwM2mResponse> implements DownlinkRe
         LwM2mNode content;
         try {
             // get content format
-            ContentFormat contentFormat = null;
+            ContentFormat contentFormat = ContentFormat.TEXT;
             if (coapResponse.getOptions().hasContentFormat()) {
                 contentFormat = ContentFormat.fromCode(coapResponse.getOptions().getContentFormat());
             }
--- a/leshan-server-demo/pom.xml
+++ b/leshan-server-demo/pom.xml
@@ -86,10 +86,7 @@ Contributors:
     </dependencies>
     <build>
         <plugins>
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>buildnumber-maven-plugin</artifactId>
-            </plugin>
+
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>


sophokles73:
Hi Peter,

please follow the instructions in our [contribution guide](https://github.com/eclipse/leshan/blob/master/CONTRIBUTING.md) in order to sign the Eclipse CLA and make sure that your commits contain a correct *signed-off-by* footer.

sophokles73:
I agree that leshan should be able to handle the lack of a content format option in a better way than throwing an exception. However, the content format option is mandatory for the client (device) to set in its answers. So you might want to fix that in the XDK as well...

leshan-server-demo/pom.xml
```diff
@@ -86,10 +86,7 @@ Contributors:
     </dependencies>
     <build>
         <plugins>
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>buildnumber-maven-plugin</artifactId>
-            </plugin>
+
```
sophokles73: why are you removing this plugin?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumObservationRegistryImpl.java
```diff
@@ -268,7 +268,7 @@ public void onNotification(Request coapRequest, Response coapResponse) {
                 LwM2mModel model = modelProvider.getObjectModel(client);
 
                 // get content format
-                ContentFormat contentFormat = null;
+                ContentFormat contentFormat = ContentFormat.TEXT;
```
sophokles73: I do not think that this is the correct point in the code to fix the problem. At this point we are just dealing with a symptom. The lack of the content format option needs to be detected deeper down the stack when we receive a response from a client. According to the latest spec the content format is mandatory for clients to set in their responses, so we should think about a general strategy for dealing with this. I can imagine to reject or ignore such response messages (the spec doesn't seem to give any hint on this) or trying to *determine* or *guess* the content format from the payload ...

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LwM2mResponseBuilder.java
```diff
@@ -352,7 +352,7 @@ private LwM2mNode decodeCoapResponse(final LwM2mPath path, final Response coapRe
         LwM2mNode content;
         try {
             // get content format
-            ContentFormat contentFormat = null;
+            ContentFormat contentFormat = ContentFormat.TEXT;
```
sophokles73: this seems to be the right place to deal with a missing content format. However, I am not sure that assuming TEXT is the right strategy.

sbernard31:
@sophokles73  is right, the spec say :
_"A Server WRITE request as well as a Client response message MUST specify a content type in using one of the supported Media Type."_

If there is an issue this is maybe in the way we handle this, but I'm not sure we can do something better than what we do now.

When the server receive a response, there is no way to send an error. we can just ACK to say we receive the message (if this is a confirmable response) and RST to stop this observe relation.

Currently we catch the error, we log it and  we return it to the application layer. The user is free to cancel or not the observation.

sophokles73:
@sbernard31,

so you're saying that the NPE that @peteryang observed is actually already caught and handled in the way you described it?

sbernard31:
I think so when I read the code, but I didn't test it, so maybe I'm wrong :).

@peteryang  could you provide the log about this NPE ?

peteryang:
@sbernard31, @sophokles73, 

SEVERE: Exception in protocol stage thread: The validated object is null
java.lang.IllegalArgumentException: The validated object is null
	at org.eclipse.leshan.util.Validate.notNull(Validate.java:222)
	at org.eclipse.leshan.util.Validate.notNull(Validate.java:205)
	at org.eclipse.leshan.core.node.codec.DefaultLwM2mNodeDecoder.decodeTimestampedData(DefaultLwM2mNodeDecoder.java:79)
	at org.eclipse.leshan.server.californium.impl.CaliforniumObservationRegistryImpl.onNotification(CaliforniumObservationRegistryImpl.java:278)
	at org.eclipse.californium.core.network.CoapEndpoint$NotificationDispatcher.onNotification(CoapEndpoint.java:568)
	at org.eclipse.californium.core.network.UdpMatcher$3.onResponse(UdpMatcher.java:417)
	at org.eclipse.californium.core.coap.Request.setResponse(Request.java:451)
	at org.eclipse.californium.core.server.ServerMessageDeliverer.deliverResponse(ServerMessageDeliverer.java:169)
	at org.eclipse.californium.core.network.stack.CoapUdpStack$StackTopAdapter.receiveResponse(CoapUdpStack.java:206)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveResponse(AbstractLayer.java:103)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveResponse(AbstractLayer.java:103)
	at org.eclipse.californium.core.network.stack.ObserveLayer.receiveResponse(ObserveLayer.java:146)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveResponse(AbstractLayer.java:103)
	at org.eclipse.californium.core.network.stack.BlockwiseLayer.receiveResponse(BlockwiseLayer.java:322)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveResponse(AbstractLayer.java:103)
	at org.eclipse.californium.core.network.stack.ReliabilityLayer.receiveResponse(ReliabilityLayer.java:265)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveResponse(AbstractLayer.java:103)
	at org.eclipse.californium.core.network.stack.CoapUdpStack.receiveResponse(CoapUdpStack.java:145)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl.receiveResponse(CoapEndpoint.java:807)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl.receiveMessage(CoapEndpoint.java:709)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl.access$1000(CoapEndpoint.java:668)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl$1.run(CoapEndpoint.java:682)
	at org.eclipse.californium.core.network.CoapEndpoint$5.run(CoapEndpoint.java:858)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LwM2mResponseBuilder.java
peteryang: I have tested other formats like OPAQUE and TLV. But it seems that TEXT has the most chance to decode successfully.

sbernard31:
I reread the code and I was wrong! For observation we don't handle it correctly.
We don't catch it and we don't make it accessible to the end user...

As @sophokles73 proposed, we should create an issue about that and close this PR.

FMPOV, we should handle response of observation like other response (catch+log+return it to application layer)

sophokles73:
@peteryang 

would you care to open an issue for this as proposed by me and @sbernard31 and close this PR?

peteryang:
new Issue is created

Issue #192, opened 2016-10-24
Merge of Client and Observation registries


msangoi:
Because the observation lifecycle is tied to the client registration, it makes sense to have a common store.

This change is related to the race condition issues: #176 and #77

sbernard31:
I'm working on it. 
Very first stage available in the [refactor_merge_store](https://github.com/eclipse/leshan/commits/refactor_merge_stores) branch.

sbernard31:
This is done (#210)

PR #210, opened 2016-11-29
Create a new store which handles registrations and observations.

sbernard31:
I refactor the code to create a new store which handles registrations and observations.

This aims to fix several issues about registrations and observations lifecycle. Merging the store allow to easily handle atomicity on registration/observations operations. (E.g. deregister will removed current observations)
more details : #192

This PR is focused on introducing the store without too many impact on current user API. (ClientRegistry and ObservationRegistry).

But this reveal some issues in the current API (e.g. `ClientRegistry` expose methods like `registerClient()` or `updateClient()` which is internal API). So the next step will probably to clean the API (with some renaming :-/ ).

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/ExpirationListener.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/ExpirationListener.java
@@ -1,0 +1,26 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Sierra Wireless and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *     Sierra Wireless - initial API and implementation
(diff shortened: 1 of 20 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationStore.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param address of the client registered.
+     * @return the registration or null if there is no client registered with this socket address.
+     */
+    Collection<Client> getRegistrationByAdress(InetSocketAddress address);
+
+    /**
+     * @return all registrations in this store.
+     * @Deprecated should be replaced by an iterator.
+     */
+    // TODO should be removed
```
msangoi: Should be replaced by an iterator version.

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/ExpirationListener.java
```diff
@@ -0,0 +1,31 @@
...
+import org.eclipse.leshan.server.client.Client;
+
+/**
+ * A listener to be aware of registration expiration.
+ * 
+ * @see RegistrationStore
+ */
+public interface ExpirationListener {
+
+    void registrationExpire(Client registration, Collection<Observation> observation);
```
msangoi: registrationExpired? or expiredRegistration?

msangoi:
Ok for me to merge so that we can move to the next refactoring steps (observe api clean-up, renaming...)

sbernard31:
This is now integrated in master :)

PR #211, opened 2016-12-02
introducing RequestRejectException to identify a rejected request

balsmn:
Specific exceptions helps to identify concrete reason why a request failed like a request was rejected by the LWM2M client or the request was cancelled. 


jvermillard:
I'm not sure to understand the use case. this should occur only if a device reject the request by sending a CoAP RESET ? I don't think LWM2M permit this.

balsmn:
> I don't think LWM2M permit this.

I am not sure about that. In LWM2M Spec Section 8.1 under Transport layer bindings, it is mentioned that

`The Reset message is used as a message layer error message in response to a malformed Confirmable message.`

As this applies to the Transport Layer (CoAP), when server sends a malformed CON, then it could get a RST from the client.

leshan-core/src/main/java/org/eclipse/leshan/core/request/exception/RequestRejectedException.java
```diff
@@ -0,0 +1,37 @@
...
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *     Bosch Software Innovations GmbH
+ *               - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.leshan.core.request.exception;
+
+/**
+ * Exception indicating that the message was rejected by the LWM2M client.
```
sbernard31: As we use it, at LWM2M client side too. I think we can not use _"LWM2M client"_ in this javadoc. Maybe we can use _"Request receiver"_ ?  Better proposition is welcome ;)

Should we use _"request"_ instead of _"message"_ too ?

leshan-core/src/main/java/org/eclipse/leshan/core/request/exception/RequestRejectedException.java
```diff
@@ -0,0 +1,37 @@
...
+public class RequestRejectedException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    public RequestRejectedException() {
+        super();
+    }
+
+    /**
+     * @param message the message returned by the server or <code>null</code> if the server did not return a message
```
sbernard31: I think there is not message payload for RST message in CoAP spec. Even if it exists the Californium API don't expose it. So we can remove this parameter.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumLwM2mClientRequestSender.java
```diff
@@ -163,12 +164,12 @@ public void onTimeout() {
 
         @Override
         public void onCancel() {
-            errorCallback.onError(new RequestFailedException("Canceled request"));
+            errorCallback.onError(new RequestCanceledException("Canceled request"));
```
sbernard31: I'm not sure we need a message parameter for `RequestCanceledException` ? (same thing for sender at server side)

leshan-core/src/main/java/org/eclipse/leshan/core/request/exception/RequestRejectedException.java
balsmn: Yes I agree. I am fine with Request receiver.

leshan-core/src/main/java/org/eclipse/leshan/core/request/exception/RequestRejectedException.java
balsmn: Good point. I will remove this constructor.

balsmn:
@sbernard31 As we have now specific exceptions like RequestCanceledException and RequestFailedException, does a generic exception like RequestFailedException is still needed? It is not used anywhere in leshan code now. I can remove it too.

sbernard31:
If this is not used, I suppose we can remove it.

balsmn:
@sbernard31 I also tried to rebase on to the leshan master and didn't get any conflict.  I don't know what this `QueuedRequestSender` conflict is. I am not able to view it.

balsmn:
@sbernard31 please ignore my previous comment. I found it and I will push it shortly resolving it.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/servlet/ClientServlet.java
```diff
@@ -194,7 +193,7 @@ protected void doPut(HttpServletRequest req, HttpServletResponse resp) throws Se
             LOG.warn("Invalid request or response", e);
             resp.setStatus(HttpServletResponse.SC_BAD_REQUEST);
             resp.getWriter().append(e.getMessage()).flush();
-        } catch (ResourceAccessException | RequestFailedException e) {
+        } catch (ResourceAccessException e) {
```
sbernard31: We should replace RequestFailedException by RequestRejectedException else we change the behavior and I'm not sure this is what we want.

balsmn:
@sbernard31  can you verify it and close it. Thanks.

sbernard31:
Integrated in master (commit a7a0b0bf6b84f9365fe5059c57c1a1258bb41d24)

PR #212, opened 2016-12-02
fixed humitidy minimum value to be float instead of string

joakimeriksson:
Signed-off-by: Joakim Eriksson <joakime@sics.se>

I Had some issues with the humidity sensors and it seems to be the JSON description file that was configured for string intstead of float for the minimum humidity value. This is a fix for that.

BTW: Is there anyone planning to add the other IPSO objects that are not yet supported?


jvermillard:
Mmaybe we need to resynch with the last XML file version because it's fixed here:
https://github.com/IPSO-Alliance/pub/blob/master/reg/xml/3304.xml

PR #215, opened 2016-12-07
Bugfix for Observation checking for observe option 0.

boaks:


boaks:
Would only be required in the future for californium 2.0.0-M3.

PR #216, opened 2016-12-07
Add implementations for getKey(ServerNames serverNames, String identity).

boaks:
Dummy, just call getKey(String identity).


boaks:
Would only be required in the future for californium 2.0.0-M3.

PR #219, opened 2016-12-15
Bugfix extracting CN from x509.

boaks:
If CN  is the last entry, its not terminated by a ','.

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
@@ -254,9 +254,9 @@ public class RegisterResource extends CoapResource {
                 return Identity.rpk(peerAddress, publicKey);
             } else if (senderIdentity instanceof X500Principal) {
                 // Extract common name
-                Matcher endpointMatcher = Pattern.compile("CN=.*?,").matcher(senderIdentity.getName());
+                Matcher endpointMatcher = Pattern.compile("CN=(.*?)(,|$)").matcher(senderIdentity.getName());
                 if (endpointMatcher.find()) {
-                    String x509CommonName = endpointMatcher.group().substring(3, endpointMatcher.group().length() - 1);
+                    String x509CommonName = endpointMatcher.group(1);
                     return Identity.x509(peerAddress, x509CommonName);
                 } else {
                     return null;
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/ResourceUtil.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/ResourceUtil.java
@@ -66,9 +66,9 @@ public class ResourceUtil {
                 return Identity.rpk(peerAddress, publicKey);
             } else if (senderIdentity instanceof X500Principal) {
                 // Extract common name
-                Matcher endpointMatcher = Pattern.compile("CN=.*?,").matcher(senderIdentity.getName());
+                Matcher endpointMatcher = Pattern.compile("CN=(.*?)(,|$)").matcher(senderIdentity.getName());
                 if (endpointMatcher.find()) {
-                    String x509CommonName = endpointMatcher.group().substring(3, endpointMatcher.group().length() - 1);
+                    String x509CommonName = endpointMatcher.group(1);
                     return Identity.x509(peerAddress, x509CommonName);
                 } else {
                     // TODO may be a other Identity type is better?


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
```diff
@@ -254,9 +254,9 @@ private Identity extractIdentity(CoapExchange exchange) {
                 return Identity.rpk(peerAddress, publicKey);
             } else if (senderIdentity instanceof X500Principal) {
                 // Extract common name
-                Matcher endpointMatcher = Pattern.compile("CN=.*?,").matcher(senderIdentity.getName());
+                Matcher endpointMatcher = Pattern.compile("CN=(.*?)(,|$)").matcher(senderIdentity.getName());
                 if (endpointMatcher.find()) {
-                    String x509CommonName = endpointMatcher.group().substring(3, endpointMatcher.group().length() - 1);
+                    String x509CommonName = endpointMatcher.group(1);
```
sophokles73: Why do we need this code twice?

boaks: What do you mean with "twice"?
One for the client one for the server?
We only have "client-cf" and "server-cf" which contains the californium specific stuff.
A common package for "cf" is currently not used.

sophokles73: There is the `leshan-core` module which both *client* and *server* depend on and which contains a `util` package.
I can image putting this functionality there instead ...

sophokles73: I am not talking about dependencies to `Californium`. This seems to be a simple utility method for extracting the DN from a subject, right? All I am asking is that we put it somewhere in a utility class in the `leshan-core` module. Or am I completely mistaken?

sbernard31: We can just move an utility method in `leshan-core`, but maybe it's time to ask again for a `leshan-core-cf` project.

We didn't create it at the beginning because we didn't have so much common code. After a quick look, I think we have 6,7 classes which could go in a `leshan-core-cf` package.

Should we create a new  `leshan-core-cf` project ?

boaks: @sophokles73 

ResourceUtil:

    private static Identity extractIdentity(CoapExchange exchange) {
        InetSocketAddress peerAddress = new InetSocketAddress(exchange.getSourceAddress(), exchange.getSourcePort());

        Principal senderIdentity = exchange.advanced().getRequest().getSenderIdentity();
        if (senderIdentity != null) {
            if (senderIdentity instanceof PreSharedKeyIdentity) {
                return Identity.psk(peerAddress, senderIdentity.getName());
            } else if (senderIdentity instanceof RawPublicKeyIdentity) {
            } else if (senderIdentity instanceof X500Principal) {
                // Extract common name
                Matcher endpointMatcher = Pattern.compile("CN=.*?,").matcher(senderIdentity.getName());
                if (endpointMatcher.find()) {
                    String x509CommonName = endpointMatcher.group().substring(3, endpointMatcher.group().length() - 1);
                } else {
                    // TODO may be a other Identity type is better?
                    LOG.warn("x509 certificate without CN from " + peerAddress);
                    return Identity.unsecure(peerAddress);
                }
            }
        }
        return Identity.unsecure(peerAddress);
    }

RegisterResource:

    private Identity extractIdentity(CoapExchange exchange) {
        InetSocketAddress peerAddress = new InetSocketAddress(exchange.getSourceAddress(), exchange.getSourcePort());

        Principal senderIdentity = exchange.advanced().getRequest().getSenderIdentity();
        if (senderIdentity != null) {
            if (senderIdentity instanceof PreSharedKeyIdentity) {
                return Identity.psk(peerAddress, senderIdentity.getName());
            } else if (senderIdentity instanceof RawPublicKeyIdentity) {
            } else if (senderIdentity instanceof X500Principal) {
                // Extract common name
                Matcher endpointMatcher = Pattern.compile("CN=.*?,").matcher(senderIdentity.getName());
                if (endpointMatcher.find()) {
                    String x509CommonName = endpointMatcher.group().substring(3, endpointMatcher.group().length() - 1);
                } else {
                    return null;
                }
            }
        }
        return Identity.unsecure(peerAddress);
    }


So the code duplication is much larger then only x509 CN.
But sure, that small part x509 we could move to core.

@sbernard31 

> Should we create a new leshan-core-cf project ?

If we want to reduce code duplications, yes :-)

boaks:
Just decide how to continue:

- util fo x509 in core
- leshan-core-cf

For "leshan-core-cf": I would prefer, if this could be created by a commiter.

sbernard31:
@sophokles73, @msangoi , @jvermillard  are you ok to create a new `leshan-core-cf` project ?

jvermillard:
:+1: or elect Achim as committer and force him to do it :D

sophokles73:
> We didn't create it at the beginning because we didn't have so much common code. After a quick look, I think we have 6,7 classes which could go in a leshan-core-cf package.

@sbernard31: and these 6, 7 classes are Californium specific and should therefore not go into `leshan-core`, right?

sbernard31:
@sophokles73 exactly !

sophokles73:
Then +1 for creating a `leshan-core-cf` module :-)

boaks:
> :+1: or elect Achim as committer and force him to do it :D

But this would delay the fix for days ;-).

By the way, I think we should "slighty" change the process:
If some lines fixing a small issue and make an other issue visible, I think we should merge the fix and create a reminder issue.

sbernard31:
> If some lines fixing a small issue and make an other issue visible, I think we should merge the fix and create a reminder issue.

I agree. (To be honest this is what we try to do :p)

I  create a new issue about a new `leshan-core-cf` project.(#222) and I will integrate your fix ASAP

PR #223, opened 2016-12-19
Move Startable, Stopable, Destoyable feature on the stores.

sbernard31:
We should not test if registrationService or observationService are startable/stopable/destroyable as they can not be implemented by users anymore. Tests should be done on stores only.


sbernard31:
integrated in master (commit 67ec454)

PR #226, opened 2016-12-21
Add authorizer interface to allow custom LWM2M client authorization

sbernard31:
Since we merged registrationStore with observationStore, we doesn't have a way to customize LWM2M client authorization anymore.

So we add a new interface responsible to do that : `Authorizer`.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/Authorizer.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/Authorizer.java
@@ -1,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Sierra Wireless and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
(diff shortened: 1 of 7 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/security/Authorizer.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param registration the registration linked to the received request.<br>
+     *        For register request this is the registration which will created<br>
+     *        For update request this is the registration before the update was done.
+     * @param senderIdentity the {@link Identity} used to send the request.
+     * @param securityStore the securityStore which contains securityInfo for a given endpoint.
+     * 
+     * @return true if this request is authorized.
+     */
+    boolean isAuthorized(UplinkRequest<?> request, Registration registration, Identity senderIdentity,
+            SecurityStore securityStore);
```
msangoi: Why do you pass the securityStore here?
It is not possible to give it to the implementation with a constructor?

sbernard31:
:gift: for @msangoi, now the `Authorizer` does not depend on `SecurityStore` anymore. The securityStore is set as constructor of the `BasicAuthorizer`

sbernard31:
:santa: I renamed `BasicAuthorizer` in `DefaultAuthorizer` too :christmas_tree:

msangoi:
👍

PR #235, opened 2017-01-04
cleaning up current queue mode implementation.

balsmn:
Cleaning up the existing Queue mode implementation and provides an API for the simplified queue mode as per recent LWM2M specification. Only API is defined as part of this PR and no implementation provided. It will be done as the next step once we agree on the API.


sbernard31:
Integrated in master (commit 28268410a0b88676ba8e52cf9c70fe90e4ac6cd0) without the new interface.

@balsmn  please create a new PR to discuss about the new API, but the interface you propose seems to be a good start.

Issue #244, opened 2017-01-13
Handle unsupported content format on notification correctly


joakimeriksson:
I am running a leshan server and a Contiki based LWM2M client and when I do an observation on for example sensor values I get exceptions as soon as the notification comes in to leshan. It is related to a missing null pointer check at line 245 org.eclipse.leshan.server.californium.impl.ObservationServiceImpl.onNotification

Where there is a piece of code like this:
```
 // decode response
                List<TimestampedLwM2mNode> timestampedNodes = decoder.decodeTimestampedData(coapResponse.getPayload(),
                        contentFormat, observation.getPath(), model);

                // create lwm2m response
                ObserveResponse response;
                if (timestampedNodes.size() == 1 && !timestampedNodes.get(0).isTimestamped()) {
                    response = new ObserveResponse(ResponseCode.CONTENT, timestampedNodes.get(0).getNode(), null,
                            observation, null, coapResponse);
                } else {
                    response = new ObserveResponse(ResponseCode.CONTENT, null, timestampedNodes, observation, null,
                            coapResponse);
                }

```
This code assumes that decoder will always return an object with timestamped nodes but I guess it is not doing that in some cases. (line 245 is the if (timestamped...).

The exception I get is the following:
```

Jan 13, 2017 8:43:57 AM org.eclipse.californium.core.network.stack.BlockwiseLayer$BlockCleanupTask run
INFO: Block2 transfer timed out: {1}
Jan 13, 2017 8:43:57 AM org.eclipse.californium.core.network.CoapEndpoint$5 run
SEVERE: Exception in protocol stage thread: null
java.lang.NullPointerException
	at org.eclipse.leshan.server.californium.impl.ObservationServiceImpl.onNotification(ObservationServiceImpl.java:245)
	at org.eclipse.californium.core.network.CoapEndpoint$NotificationDispatcher.onNotification(CoapEndpoint.java:568)

```

joakimeriksson:
Found out the reason - I was still sending old plain-text format (1541). Maybe there could be an improved exception for this!?

sbernard31:
Yes you're right we need to provide more feedback when an error occur on notification, there is an issue about that https://github.com/eclipse/leshan/issues/206

We should handle unsupported format in better way than just returning null, maybe by rising an exception ? WDYT ?

joakimeriksson:
Good improvement - this is how all things should be handled to provide useful feedback to anyone using  Leshan! Many thanks!

[...]

PR #245, opened 2017-01-13
Handle error on notification correctly

### LINKED ISSUE #206: Give feedback to end user when something wrong occurs on notification reception.
Even though the specs requires the client MUST send content format in the response,  the server could have better handling strategy if this happens.  

Please refer to the discussion in the pull request:  https://github.com/eclipse/leshan/pull/204

One possible solution is to add Not spec compatible related APIs:

public class NotSpecCompatibleAPIs{
   static void setResponseContentFormatIfNull(CaliforniumObservationRegistryImpl reg, ContentFormat cf){

}
}
I choose CaliforniumObservationRegistryImpl as the parameter because it will not change the public  ObservationRegistry interface.

  It is not elegant, but it helps to developers easy because it takes time to ask device manufacturers to fix their code.

sbernard31:
This PR aims to fix #206 and #244.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/observation/ObservationListener.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/observation/ObservationListener.java
@@ -30,5 +30,13 @@ public interface ObservationListener {
      * @param observation the observation for which new data are received
      * @param reponse the lwm2m response received
      */
-    void newValue(Observation observation, ObserveResponse response);
+    void onResponse(Observation observation, ObserveResponse response);
+
+    /**
+     * Called when an error occurs on new notification.
+     * 
+     * @param observation the observation for which new data are received
+     * @param error the exception raised when we handle the notification
+     */
+    void onError(Observation observation, Exception error);
 }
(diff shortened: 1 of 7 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/observation/ObservationListener.java
```diff
@@ -30,5 +30,13 @@
      * @param observation the observation for which new data are received
      * @param reponse the lwm2m response received
      */
-    void newValue(Observation observation, ObserveResponse response);
+    void onResponse(Observation observation, ObserveResponse response);
```
sophokles73: I like the newly introduced error handler method and I also think that we should rename `newValue()` :-)
My understanding is that `onResponse` will be invoked with ANY response received from the LWM2M client that we are able to process whereas `onError` is invoked when we cannot process the response, right? In particular this means that responses sent form the client containing error codes (4.xx) will also be delivered via the `onResponse` method if the content can be processed correctly, right?

sbernard31: That's right. I tried to have a similar approach than our current [send API](https://github.com/eclipse/leshan/blob/master/leshan-server-core/src/main/java/org/eclipse/leshan/server/LwM2mServer.java#L79).

sophokles73: We need to make sure that we document this properly in these methods' JavaDoc because I am afraid that `onError()` might otherwise be easily mistaken as a handler for 4.xx responses...

sbernard31: You're right. I think this is a general issue. We clearly need to spend time to better document our API. This is something we should do for the 1.0.0.

balsmn: Just wondering what happened to `cancelled` method here. How will the listener gets notified about an observation cancellation?

sbernard31:
Integrated in master (commit c7c25f5e30dbdd2a72dcf9c4172ffbc0e8b56c73, 35d10e3b00bf6db66af719cd32db37bff11be3c3, 19de0abf97d2c96499d8ba60ac444c0ece12b24a, 5bbca50a8e1dab722d4644aaa5b98c32001cb5ca).

Issue #247, opened 2017-01-17
ClientRegistryListener execute during client registration


tomaszszlek:
Hi,
We discovered issue with custom ClientRegistryListener.
We implemented some logic in registered method which is time consuming.
The issue is that registered method code is invoked by the same thread which handle client registration.
It means that when client is registering, server should register client and then invoke Listener method.
We saw that javadoc of registered method says 

> Invoked when a new client has been registered on the server.

while the method is invoked before client complete registration.

sbernard31:
If you want to do time consuming task, you should do that in new thread. 

About Leshan, I don't think it could be a good idea to create a new thread automatically, I rather prefer to let the choice to Leshan users.

About complete registration, the response to the register request is not confirmable, this means that we can not know if the device receive the response or not.  So at server point of view we consider that the client is registered as soon as we add the registration in the store. So I think the javadoc is correct. (but maybe we should be more explicit)

Anyway, we should probably send the response before invoking listener method, but keep in mind that packet could be lost or recieve in wrong order with UDP.

tomaszszlek:
Thank you for answer.
I agree about time consuming task but let's imagine scenario when in listener you would like to read some particular resources of client - you can't do it because client is pending registration and server will receive timeout.

Regarding javadoc - for me registered "client registered on the server" = server sent response, this is something that made me confused.

Because of scenario I described above I think that sending response before invoking listener makes a lot of sense.

About UDP - when packets will be damaged during server registration response, client will need to resend registration request because response will be incorrect ? How Leshan will behave when request/response sent by server will be damaged due to UDP nature ?
Thank you in adavance.

sbernard31:
>I agree about time consuming task but let's imagine scenario when in listener you would like to read some particular resources of client - you can't do it because client is pending registration and server will receive timeout.

Not exactly because of  retransmission mechanism ([see](https://github.com/eclipse/wakaama/pull/24#issuecomment-78451687))

> Because of scenario I described above I think that sending response before invoking listener makes a lot of sense.

I will see if I can change the code in that sense.

>About UDP - when packets will be damaged during server registration response

damaged ? With UDP, packets can be reordered or lost but not damaged.

tomaszszlek:
Yes sorry, i meant package reorder.

sbernard31:
Yes it should.

sbernard31:
I try to fix this in the #249 PR. Please tell us, if it behave better now.

tomaszszlek:
[...]

PR #249, opened 2017-01-18
Send registration event after sending response to device

sbernard31:
This PR was motivated by #247.

It aims to send registration event (register, update, deregister) after we effectively send the LWM2M response to the device. Even if UDP does not guarantee ordered and like it is explain [here](https://github.com/eclipse/wakaama/pull/24#issuecomment-78451687) it is not mandatory to make it works, this could be an improvement most of the time. 

It is largely inspired by what I have done [here](https://github.com/eclipse/leshan/commit/776c8e48595e050569553a7b6c5b089a18485697) for #16, but trying to keep the `LwM2mResponse` immutable.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationHandler.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationHandler.java
@@ -26,6 +26,7 @@ import org.eclipse.leshan.core.response.DeregisterResponse;
 import org.eclipse.leshan.core.response.RegisterResponse;
 import org.eclipse.leshan.core.response.UpdateResponse;
 import org.eclipse.leshan.server.impl.RegistrationServiceImpl;
+import org.eclipse.leshan.server.impl.SendableResponse;
 import org.eclipse.leshan.server.security.Authorizer;
 import org.eclipse.leshan.util.RandomStringUtils;
 import org.slf4j.Logger;
@@ -47,7 +48,7 @@ public class RegistrationHandler {
         this.authorizer = authorizer;
     }
 
-    public RegisterResponse register(Identity sender, RegisterRequest registerRequest,
+    public SendableResponse<RegisterResponse> register(Identity sender, RegisterRequest registerRequest,
             InetSocketAddress serverEndpoint) {
 
         Registration.Builder builder = new Registration.Builder(RegistrationHandler.createRegistrationId(),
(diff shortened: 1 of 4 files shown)


tomaszszlek:
I have tested your PR and it works a little bit strange.
I have following code in my registration listener:
`    
   @Override

    public void registered(Registration registration) {
        LOG.info("Client should be registered here at this point");
        try {
            Thread.currentThread().sleep(1000 * 30);
        }
        catch (InterruptedException e) {
            throw new RuntimeException(e);
        }
        LOG.info("Register listener finished his work");
    }
`

I would expect that client will start register, then server receive the response, finally after 30 seconds my listener code will execute but behavior is different.
On client side I see registration request, after that server is not responding client waits for registration.
When 30 seconds have passed then server is responding to client and registration is finished. 
Client logs:

> 
> ...
> 
> 2017-01-19 14:36:15,701 INFO RegistrationEngine - Trying to register to coap://localhost:5683
> 
> ...
> 
> 2017-01-19 14:36:45,814 INFO RegistrationEngine - Registered with location '/rd/YIgPGdm9d7'.
>

sbernard31:
You're right. I thought that "exchange complete" notification was called after the response was sent, but it seems I was wrong ... : / .

So, if I understand correctly how californium( the JAVA CoAP library we use) works, I can't see any good solution for this problem...

@sophokles73 What did you think about that ? Did you see any solution to sent registration notification after we send the CoAP response ?

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationHandler.java
```diff
@@ -62,82 +63,95 @@ public RegisterResponse register(Identity sender, RegisterRequest registerReques
...
+        final Deregistration deregistration = registrationService.getStore().addRegistration(registration);
 
-        // notify new registration and de-registration
+        // Create callback to notify new registration and de-registration
         LOG.debug("New registration: {}", registration);
-        if (deregistration != null) {
-            registrationService.fireUnregistered(deregistration.getRegistration());
-        }
-        registrationService.fireRegistred(registration);
+            Runnable whenSended = new Runnable (){
```
jvermillard: whenSent

boaks:
@sbernard31 
I think, the solution depends on the wanted QoS :-).
Though californium tries to decouple the CoAP message logic from the "message transfer" (element-connector), there are two possibilities:
1. when CoAP core hands over the message to the connector (CoapEndpoint.OutboxImpl)
2. when the connector rely sends the message

So working on californium https://github.com/eclipse/californium/pull/202 I would guess 2 will be the right. So may be we introduce/extend a "callback", which informs us, if the message is send or droped.

sbernard31:
>So may be we introduce/extend a "callback", which informs us, if the message is send or droped
I like the idea.

sophokles73:
Sorry for joining the party this late but I am not sure if I understand the intention fully. Do you guys want to make sure that the event is only fired once we are sure that the LWM2M client has received the server's response to its registration request? If not then IMHO it doesn't really make a difference which *stage* of  Cf processing the response we use as the trigger for firing the event. The most likely reason for the client not receiving the response will be that the message gets lost during transmission. It is pretty unlikely that the message gets lost while flowing through Cf's stack...

sbernard31:
>Do you guys want to make sure that the event is only fired once we are sure that the LWM2M client has received the server's response to its registration request? 

No, we want to make sure event is only fired once we sent the message.

sophokles73:
Ok, I have taken a close look at your code @sbernard31. The problem is that `ExchangeObserver.completed()` gets invoked in `Matcher.sendResponse()` **before** the CoAP response is actually sent via the connector (by the `Outbox`). This is because the `Matcher` is *higher* in the stack than the `Outbox` (which is actually at the bottom of the stack).
Why don't you simply remove the `ExchangeObserverAdapter` and invoke `SendableResponse.sent()` as the last step from within `RegisterResource.handleRegister()` right after you have sent back the response to the client?

```
...
        // Create CoAP Response from LwM2m request
        // -------------------------------
        if (response.getCode() == org.eclipse.leshan.ResponseCode.CREATED) {
            exchange.setLocationPath(RESOURCE_NAME + "/" + response.getRegistrationID());
            exchange.respond(ResponseCode.CREATED);
        } else {
            exchange.respond(fromLwM2mCode(response.getCode()), response.getErrorMessage());
        }
        sendableResponse.sent();
}
```

sbernard31:
Because, `exchange.respond()` does not sent synchronously the response.

- The message will be pushed in a [queue of the connector](https://github.com/eclipse/californium/blob/master/element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java#L187) and will be handled later by a thread.
or
- The message will be handled asynchronously by the coapendpoint if there is a [ `customExecutor`](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java#L459).

(This is not the case here but we can also imagine that the response is sent using block mode and we would like to be notify when the last block is sent)

sbernard31:
In case we have no `customExecutor`, your proposition is probably better than using `ExchangeObserver.completed()`, because the event callback is called later. 
And in all case event handling will not block message delivery.

I will do that and @tomaszszlek would be able to test it and tell us if it's better.

sophokles73:
> Because, exchange.respond() does not synchronously sent the response.

> The message will be pushed in a queue of the connector and will be handled later by a thread.
or
The message will be handled asynchronously by the coapendpoint if there is a customExecutor.

That's true but you cannot rely on the order the UDP packets will be delivered to the client anyway so I do not see why this should be a problem ...

boaks:
> It is **pretty unlikely** that the message gets **lost** while flowing through Cf's stack...

According https://github.com/eclipse/californium/issues/104 messages **should be dropped**, if they were "sent" into the wrong DTLS session. Though the message process flow seems to be potential asynchron/parallel, the right time point in my opinion would be the one, when the DTLS session used for encryption is known and either matching on not matching (and therefore the message should be dropped).

sbernard31:
> That's true but you cannot rely on the order the UDP packets will be delivered to the client anyway so I do not see why this should be a problem ...

I agree but most of the time users want to send request just after registration is done. (using registered event). Like David explained [here](https://github.com/eclipse/wakaama/pull/24#issuecomment-78451687) it works even with disordered message . But most of the time, messages are sent in right order, so this change can avoid the first retransmission. 
This is just a potential optimization.

sbernard31:
@tomaszszlek I pushed the solution described by @sophokles73. Does it better for you ?

sophokles73:
@sbernard31,

any news on this? FMPOV this is an improvement anyway so I would be ok to merge this even without further feedback from @tomaszszlek ...

sbernard31:
Finally, integrated in master (commit ece1c3e2bd7c0295c76a38dece3b6202e700e1af)

PR #250, opened 2017-01-19
Replace "getAllRegistrations()" deprecated API  by "iterator()"

sbernard31:


leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationStore.java
```diff
@@ -65,15 +66,15 @@
...
-     * @Deprecated should be replaced by an iterator.
+     * Returns an iterator over the registration of this store. There are no guarantees concerning the order in which
+     * the elements are returned (unless the implementation provides a guarantee).
+     *
+     * @return an <tt>Iterator</tt> over the registration in this store
      */
-    // TODO Should be replaced by an iterator.
-    @Deprecated
-    Collection<Registration> getAllRegistration();
+    Iterator<Registration> iterator();
```
msangoi: I would keep getAllRegistrations() as method name.

sbernard31: You're right. It's probably better.

leshan-server-core/src/main/java/org/eclipse/leshan/server/client/RegistrationService.java
```diff
@@ -39,13 +39,12 @@
...
-     * @return the registrations.
-     * @deprecated should be replace by an iterator
+     * Returns an iterator over all registrations. There are no guarantees concerning the order in which the
+     * elements are returned.
+     *
+     * @return an <tt>Iterator</tt> over registrations
      */
-    @Deprecated
-    Collection<Registration> getAllRegistrations();
+    Iterator<Registration> getAllRegistrations();
```
sophokles73: What's the difference between `RegistrationService` and `RegistrationStore`? Both seem to have this method but it is unclear to me why we need both.
I also wonder whether this method is used by any other client than the demo server web UI for retrieving the list of registered devices. In that case, returning an iterator doesn't seem to be a good fit because it does not provide any support for efficient paging, does it?

sbernard31: `RegistrationService` is the API exposed to Leshan users : you can add listener on registration event, access to registrations ...  `RegistrationService` should not be implemented by users, it just help to hide internal API.

`RegistrationStore` is the API needed by Leshan. Users should implement to be able to persist registration using  technology they want. 

About the iterator, It's a good question. IMHO, the `RegistrationService` should expose an API to go through registrations. The previous one was clearly not compatible with a large number of registrations.  So we decide to replace it by **iterator** API. **paging** API is also a possible alternative. Hard to know which are the more suitable approach. **Users feedback about that are welcomed !**

In all case, users are free to implement their own way to browse registrations in their store. This will just be not so easy to access than if we put it in the `RegistrationService` API

jvermillard:
@sophokles73 
> returning an iterator doesn't seem to be a good fit because it does not provide any support for efficient paging, does it?

I think it depends on how you implement your iterator:
https://stackoverflow.com/questions/12557837/paginated-queries-iterator-recipe

You had something specific in mind for supporting paging?

sbernard31:
@sophokles73  waiting for you to integrate this PR ? FMPOV we can start with an iterator and see if this satisfies users.

sophokles73:
If you want to put a UI in front of the `RegistrationStore` (e.g. the Demo Server Web UI) then I currently do not see how we could implement a paging mechanism based on this Iterator returned by `getAllRegistrations()`, simply because there is no way for the UI to specify which page it wants (e.g. by specifying a specific index), or am I missing something?
This is not an issue for the demo server where we will probably only have a dozen devices registered but for production this seems to be insufficient. Or would we not use the `RegsitrationStore` interface in a production system anyways but instead delegate device registration to a separate system and only use the `RegistrationService` to look up information for individual clients, e.g. for authentication purposes?

sbernard31:
>If you want to put a UI in front of the RegistrationStore (e.g. the Demo Server Web UI) then I currently do not see how we could implement a paging mechanism based on this Iterator returned by getAllRegistrations(), simply because there is no way for the UI to specify which page it wants (e.g. by specifying a specific index), or am I missing something?

If you want to create a UI which presents all the registration by page you **can not do that with this API**, but you can create an infinite scroll UI.
Anyway, if you need this kind of paging, you also probably need sorting and searching features. All of this will probably depend of the implementation of your store.  Up to you to implement thoses methods  for your RegistrationStore. We could try to define an abstraction for that but it seems to me a bigger business than what we try to achieve here.

The legitimate question is : "So why adding this API to RegistrationService ? Why not just remove it ?"
FMPOV, `RegistrationService` is one of the first Leshan API expose to new users. Proposed an easy way to go through all registrations can make senses for beginner. And everybody don't need to handle millions devices.

sophokles73:
I see.
On a side note: I have to admit that I have not spent a lot of time recently on leshan. However, when I take a look at the `remove_getallregistration` branch I find it hard to understand which interface is supposed to be responsible for what. In particular, I stumbled across the `RegistrationHandler`'s dependency on `RegistrationServiceImpl`. Is there a particular reason why it cannot depend on other interfaces only? Or does this simply mean that we didn't get the interfaces right yet?

sbernard31:
`RegistrationHandler` in internal class : aim to handle registration logic (a kind of controller)

`RegistrationService` is the API exposed to Leshan users : you can add listener on registration event, access to registrations ...  `RegistrationService` should not be implemented by users, it just help to hide internal API.

`RegistrationStore` is responsible to persist registration and observations.  Users could implement this interface.

sbernard31:
I didn't really reply to the question :
>In particular, I stumbled across the RegistrationHandler's dependency on RegistrationServiceImpl. Is there a particular reason why it cannot depend on other interfaces only? Or does this simply mean that we didn't get the interfaces right yet?

`RegistrationHandler` depends on `RegistrationServiceImpl` especially because of "fire event" methods. We don't want to expose thoses methods to users, so we don't add it to the interface.
The `RegistrationServiceImpl` is a kind of vestige of the old ClientRegistryImpl, currently it hold `RegistrationListener`. We could maybe organize code in a better way, but I can not find a good way to do for now and all of this is internal. I'm currently focus on external API.

sbernard31:
integrated in master (commit 70ce28b302d9c08b3ea14326b6ec0fa565623166, ca83975f77c1505bd8949baa48fd0ebc7f591adb, 05d647d7be6ab41a9810a9d55be6537dd3f1b84f )

PR #254, opened 2017-01-25
Add registration parameters for observations event

sbernard31:
Add the registration in parameters of observations events. This allow for example to easily send request when we receive a new notification.

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/ObservationServiceImpl.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/ObservationServiceImpl.java
@@ -80,7 +80,7 @@ public class ObservationServiceImpl implements ObservationService, NotificationL
         this.decoder = decoder;
     }
 
-    public void addObservation(Observation observation) {
+    public void addObservation(Registration registration, Observation observation) {
         // cancel any other observation for the same path and registration id.
         // delegate this to the observation store to avoid race conditions on add/cancel?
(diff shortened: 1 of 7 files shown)


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/ObservationServiceImpl.java
```diff
@@ -220,13 +220,13 @@ public void onNotification(Request coapRequest, Response coapResponse) {
...
-            try {
-                // get registration
-                Registration registration = registrationStore.getRegistration(observation.getRegistrationId());
-                if (registration == null)
-                    // TODO Should we clean registrationIDs maps ?
-                    return;
+            // get registration
+            Registration registration = registrationStore.getRegistration(observation.getRegistrationId());
+            if (registration == null)
+                // TODO Should we clean registrationIDs maps ?
```
jvermillard: we should log something no?

sbernard31: Yes, we should ! I will add this. ERROR ?

msangoi: Yes, I think ERROR is fine, this case should not happen.

sbernard31:
Integrated in master (commit 450abdcbb0364a5862ba221767916e7039df393e)

PR #255, opened 2017-01-27
Raise CodecException when error occured in decoder/encoder

### LINKED ISSUE #228: LwM2mNodeEncoder should raise a checked exception if the node is invalid
As an example, if the node cannot be encoded because of an incompatibility between the node and the model (e.g. invalid value type), a runtime `IllegalArgumentException` is raised. 
It would be cleaner to have a checked exception with a proper message so that the caller is able to handle this specific failure.

sbernard31:
This PR aims to fix #228.

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/ObjectResource.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/ObjectResource.java
@@ -41,7 +41,7 @@ import org.eclipse.leshan.core.model.LwM2mModel;
 import org.eclipse.leshan.core.node.LwM2mNode;
 import org.eclipse.leshan.core.node.LwM2mObjectInstance;
 import org.eclipse.leshan.core.node.LwM2mPath;
-import org.eclipse.leshan.core.node.codec.InvalidValueException;
+import org.eclipse.leshan.core.node.codec.CodecException;
 import org.eclipse.leshan.core.node.codec.LwM2mNodeDecoder;
 import org.eclipse.leshan.core.node.codec.LwM2mNodeEncoder;
 import org.eclipse.leshan.core.request.BootstrapWriteRequest;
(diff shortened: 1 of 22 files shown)


leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/Lwm2mNodeEncoderUtil.java
```diff
@@ -141,12 +142,16 @@ public static Object convertValue(Object value, Type currentType, Type expectedT
                 // let's assume we received an hexadecimal string
                 LOG.debug("Trying to convert hexadecimal string {} to byte array", value);
                 // TODO: check if we shouldn't instead assume that the string contains Base64 encoded data
-                return Hex.decodeHex(((String) value).toCharArray());
+                try {
+                    return Hex.decodeHex(((String) value).toCharArray());
+                } catch (IllegalArgumentException e) {
+                    throw new CodecException(String.format("Unalble to convert hexastring [%s] to byte array", value));
```
msangoi: typo

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/ObjectResource.java
```diff
@@ -216,7 +216,7 @@ public void handlePUT(final CoapExchange coapExchange) {
                 }
 
                 return;
-            } catch (InvalidValueException e) {
+            } catch (CodecException e) {
                 LOG.warn("Unable to decode payload to write", e);
                 coapExchange.respond(ResponseCode.INTERNAL_SERVER_ERROR);
```
msangoi: BAD_REQUEST?

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvEncoder.java
```diff
@@ -185,7 +186,7 @@ private Tlv encodeResource(LwM2mResource resource) {
             case OBJLNK:
                 return TlvEncoder.encodeObjlnk((ObjectLink) value);
```
msangoi: A try/catch block (for IllegalArgumentException) should be added here around all TlvEncoder.encode() methods. 
And it may be cleaner to raise TlvException instead of IllegalArgumentException in the TlvEncoder class.

Issue #238, opened 2017-01-10
LwM2mRequestSender /  String requestTicket


boaks:
Are there any plans to (optionaly) use the CoAP token for this requestTicket to reduce the states of the communication stack?

balsmn:
I am not sure about using the CoAP token in place for a requestTicket. RequestTicket can be used by Applications from Northbound interface to correlate the read/write with the responses. On the otherhand CoAP tokens, for example, can be long living in case of observations. Correct me if I am wrong.

boaks:
Hm, why should observe/notfies be mapped to north in a different way then a single request/reponse pair?
RFC 7641 specifies them just as request with multiple responses using the same token.
So if north want to use Information Reporting, it should naturally use the requestTicket as well.

May be the main issues would be the scope of the uniqueness and the responsibilities to generate them.

balsmn:
Yes, I agree with you regarding the responsiblity to generate them. Actually the requestTicket is generated at the time when the request comes in from the Application through the Northbound interface. If we could set that requestTicket as a token for the CoAP request then it would reduce the states as you mentioned.

sbernard31:
Now we don't store requests in a queue anymore (because of #191). Do we still need the request ticket in the leshan java API ?

sbernard31:
I rethink about that and I can not see a good reason to keep this API : 
`send(Registration destination, String requestTicket, DownlinkRequest<T> request)`

This made sense when we persisted request in queue in leshan, now we can remove it.
If there is no objection, I will remove this API in the next few days.

boaks:
I was irritated by the  `@Deprecated` in `LwM2mRequestSender` for the other sends.
So, if the `requestTicket` should be removed, the `@Deprecated` may also be corrected :-)

sbernard31:
Of course :)

sbernard31:
As #256 is integrated in master, should we close this issue ?

boaks:
your right.

PR #256, opened 2017-01-31
Remove send with ticket API

sbernard31:
As we don't need to persist request in queue anymore, we can remove the send with ticket API. 
(https://github.com/eclipse/leshan/issues/238#issuecomment-272207682)


sbernard31:
Integrated in master (commit 2036cd922f8cca6da517ef708abf2be772fb1dfb)

PR #259, opened 2017-02-02
Do no create default SecurityStore

sbernard31:
`SecurityStore` parameter is not mandatory anymore to create a `LeshanServer`. If there is no `SecurityStore` there is no secure endpoint.
The builder don't create a `FileSecurityStore` by default anymore.


sbernard31:
Integrated in master (commit 6203018f36f13b087c6f41f15c8ae7d939f49377)

Issue #191, opened 2016-10-24
Queue mode change following last spec changes


jvermillard:
Now

> Queue Mode offers functionality for a LWM2M Client to inform the LWM2M Server that it may be disconnected for an extended period of time and also when it becomes reachable again.

So the LWM2M spec doesn't specify the need to keep a queue of messages as part of the protocol implementation.

How we change that?

We should plug the "message queue" implementation on top of leshan API.
Also Leshan API could provide an API for being informed of the online/offline status of  queue mode devices and also to directly fail requests which are sent to offline devices (in place of trying for 2 minutes to send a CoAP message).

balsmn:
hi Julien,

As Queue Mode is not explicity defined for any queued messages, I propose to complete remove the Queue Mode implementation from Leshan and let the application layer decide what it wants when it receives a timeout for a request (indicating the device is **offline**) and when it receives an registration-update (**online**).

sbernard31:
it sounds good to me.

sophokles73:
@jvermillard,

I agree that we should remove the *queue mode* implementation from leshan's core and (if at all) move it on top of it. However, I agree with @balsmn in that we should probably better remove it completely from leshan and instead make sure that an application gets all the information it needs to implement desired behavior for Q devices. My understanding is that this is what you intend to do with 
>Also Leshan API could provide an API for being informed of the online/offline status of queue mode devices and also to directly fail requests which are sent to offline devices (in place of trying for 2 minutes to send a CoAP message).

I am struggling though with how the leshan server can know about the online/offline status of the device.
It seems obvious that it can call back the application about the fact that the device has updated its registration and will therefore be *online* for at least some time now. However, how does the server know when the device is going to be *offline*? I can imagine that the leshan server *caches* the outcome of its latest request to the device and use that information to tell the application that it currently thinks that the device is offline. But for how long would you consider this information valid?

sbernard31:
I suppose this should be a configurable value and the default value should be something at least equals to MAX_TRANSMIT_WAIT.

sbernard31:
@balsmn did you start to work on this new (online/offline) API ?

balsmn:
@sbernard31 yes I started but have not yet completed. I am integrating the new listener API in to RegistrationServiceImpl as an example to start the discussion. But I can also provide the API alone suggested earlier in PR #235 as a separate PR without any concrete use of it. what do you think?

sbernard31:
[...]

PR #265, opened 2017-02-03
introduced a simplified Queue mode API and implementation.

balsmn:
This PR aims to provide a simplified QueueMode API and implementation as defined in the latest LWM2M Specification (without actually queuing any messages instead notifying applications when the LWM2M client state changes to online or offline).

TODO: When a LWM2M client timeout on an request, no notification (offline) is sent. Once we agree on the initial API, it can be done. For example: one could have such a notification sent from `SyncRequestObserver`.

I would like to get the following questions answered as part of this PR.

1. What qualifies a LWM2M Client to be online (previous state: offline)?
    Possibility: Once we receive a notification for an observation, when we receive a registration update and when we receive a response 

2. What qualifies a LWM2M Client to be offline (previous state: online)?
    Possibility: Once a request timeout and ...?

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/LwM2mServer.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/LwM2mServer.java
@@ -25,6 +25,8 @@ import org.eclipse.leshan.core.response.LwM2mResponse;
 import org.eclipse.leshan.core.response.ResponseCallback;
 import org.eclipse.leshan.server.model.LwM2mModelProvider;
 import org.eclipse.leshan.server.observation.ObservationService;
+import org.eclipse.leshan.server.queue.PresenceListener;
+import org.eclipse.leshan.server.queue.PresenceService;
 import org.eclipse.leshan.server.registration.Registration;
 import org.eclipse.leshan.server.registration.RegistrationService;
 import org.eclipse.leshan.server.security.SecurityStore;
@@ -121,6 +123,12 @@ public interface LwM2mServer {
      * observation or cancel it.
      */
     ObservationService getObservationService();
+
+    /**
+     * Get the Presence service to get the status of LWM2M clients connected with binding mode 'Q'. You can use this
+     * object to add {@link PresenceListener} to get notified when a device comes online or offline.
+     */
+    PresenceService getPresenceService();
 
     /**
      * Get the SecurityStore containing of security information.
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/Presence.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/Presence.java
@@ -1,0 +1,26 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *     Bosch Software Innovations GmbH - initial API
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+/**
+ * Possible states of a LWM2M client registered with Queue mode binding.
+ */
+public enum Presence {
+    /** LWM2M Client is reachable and messages can be sent **/
+    ONLINE,
+    /** LWM2M Client is not reachable and no messages can be sent **/
+    OFFLINE
+}
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceListener.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceListener.java
@@ -1,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *     Bosch Software Innovations GmbH - initial API
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+import org.eclipse.leshan.server.registration.Registration;
+
+/**
+ * A listener aware of the status of LWM2M Client using queue mode binding.
+ *
+ */
+public interface PresenceListener {
+
+    /**
+     * This method is invoked when the LWM2M client with the given endpoint state changes from online to offline. This
+     * listener method will be invoked only once when the state change occurs;i.e:- it will not be invoked when the
+     * previous state of the endpoint is offline and further events from the client indicates that the client is still
+     * offline.
+     * 
+     * @param registration data of the lwm2m client.
+     */
+    void onOffline(Registration registration);
+
+    /**
+     * This method is invoked when the LWM2M client with the given endpoint state changes from offline to online again.
+     * This listener method will be invoked only once when the state change occurs;i.e:- it will not be invoked when the
+     * previous state of the endpoint is online and further events from the client indicates that the client is still
+     * online.
+     * 
+     * @param registration data of the lwm2m client.
+     */
+    void onOnline(Registration registration);
+}
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
@@ -1,0 +1,50 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *     Bosch Software Innovations GmbH - initial API
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+import org.eclipse.leshan.server.registration.Registration;
+
+/**
+ * Tracks the status of each LWM2M client registered with Queue mode binding. Also ensures that the
+ * {@link PresenceListener} are notified on state changes only for those LWM2M clients registered using Queue mode
+ * binding.
+ *
+ * @see Presence
+ */
+public interface PresenceService {
+
+    /**
+     * Add the listener to get notified when the LWM2M client goes online or offline.
+     * 
+     * @param listener target to notify
+     */
+    void addListener(PresenceListener listener);
+
+    /**
+     * Remove the listener previously added. This method has no effect if the given listener is not previously added.
+     * 
+     * @param listener target to be removed.
+     */
+    void removeListener(PresenceListener listener);
+    
+    /**
+     * Returns the current status of a given LWM2M client registration.
+     * 
+     * @param registration data to check
+     * @return true if the status is {@link Presence#ONLINE}
+     */
+    boolean isOnline(Registration registration);
+}
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
@@ -1,0 +1,100 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *     Bosch Software Innovations GmbH - initial API
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import org.eclipse.leshan.server.registration.Registration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tracks the status of each LWM2M client registered with Queue mode binding. Also ensures that the
+ * {@link PresenceListener} are notified on state changes only for those LWM2M clients registered using Queue mode
+ * binding.
+ *      
+ * @see Presence
+ */
+public final class PresenceServiceImpl implements PresenceService {
+    private static final Logger LOG = LoggerFactory.getLogger(PresenceServiceImpl.class);
+    private final ConcurrentMap<String, Presence> clientStatus = new ConcurrentHashMap<>();
+    private final List<PresenceListener> listeners = new CopyOnWriteArrayList<>();
+
+    @Override
+    public void addListener(PresenceListener listener) {
+        listeners.add(listener);
+    }
+
+    @Override
+    public void removeListener(PresenceListener listener) {
+        listeners.remove(listener);
+    }
+
+    @Override
+    public boolean isOnline(Registration registration) {
+        return Presence.ONLINE.equals(clientStatus.get(registration.getEndpoint()));
+    }
+
+    /**
+     * Sets the client identified by the given registration in {@link Presence#OFFLINE} mode if the current state is
+     * {@link Presence#ONLINE}. If the state is changed successfully, then the corresponding
+     * {@link PresenceListener} are notified.
+     * 
+     * @param registration lwm2m client registration data
+     */
+    public void setOffline(Registration registration) {
+        if (registration.usesQueueMode()
+                && transitState(registration.getEndpoint(), Presence.ONLINE, Presence.OFFLINE)) {
+            for (PresenceListener listener : listeners) {
+                listener.onOffline(registration);
+            }
+        }
+
+    }
+
+    /**
+     * Sets the client identified by the given registration in {@link Presence.ONLINE} mode if the current state is
+     * either {@link Presence#OFFLINE} or <code>null</code>. If the state is changed successfully, then the
+     * corresponding {@link PresenceListener} are notified.
+     * 
+     * @param registration lwm2m client registration data
+     */
+    public void setOnline(Registration registration) {
+        if (registration.usesQueueMode()
+                && transitState(registration.getEndpoint(), Presence.OFFLINE, Presence.ONLINE)) {
+            for (PresenceListener listener : listeners) {
+                listener.onOnline(registration);
+            }
+        }
+    }
+
+    private boolean transitState(String endpoint, Presence from, Presence to) {
+        boolean updated = clientStatus.replace(endpoint, from, to);
+        if (updated) {
+            LOG.debug("Client {} state update from {} -> {}", endpoint, from, to);
+        } else if (clientStatus.putIfAbsent(endpoint, to) == null) {
+            LOG.debug("Client {} state update to -> {}", endpoint, to);
+            updated = true;
+        } else {
+            LOG.trace("Cannot update Client {} state {} -> {}. Current state is {}", endpoint, from, to,
+                    clientStatus.get(endpoint));
+        }
+        return updated;
+    }
+}
--- a/leshan-server-core/src/test/java/org/eclipse/leshan/server/queue/PresenceServiceTest.java
+++ b/leshan-server-core/src/test/java/org/eclipse/leshan/server/queue/PresenceServiceTest.java
@@ -1,0 +1,117 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *     Bosch Software Innovations GmbH - initial API
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+
+import org.eclipse.leshan.core.request.BindingMode;
+import org.eclipse.leshan.server.registration.Registration;
+import org.junit.Test;
+
+/**
+ * tests the implementation of {@link PresenceService}
+ *
+ */
+public class PresenceServiceTest {
+
+    private PresenceServiceImpl instance = new PresenceServiceImpl();
+
+    @Test
+    public void testSetOnlineForNonQueueMode() throws Exception {
+        Registration registration = givenASimpleClient();
+        instance.addListener(new PresenceListener() {
+
+            @Override
+            public void onOnline(Registration registration) {
+                fail("No invocation was expected");
+            }
+
+            @Override
+            public void onOffline(Registration registration) {
+                fail("No invocation was expected");
+            }
+        });
+        instance.setOnline(registration);
+    }
+
+    /**
+     * Test method for
+     * {@link org.eclipse.leshan.server.queue.PresenceService#setOffline(org.eclipse.leshan.server.client.Registration)}.
+     * 
+     * @throws Exception
+     */
+    @Test
+    public void testSetOnlineAndOffline() throws Exception {
+        Registration queueModeRegistration = givenASimpleClientWithQueueMode();
+        instance.addListener(new PresenceListener() {
+            int onlineStatusCountDown = 0;
+            int offlineStatusCountDown = 0;
+
+            @Override
+            public void onOnline(Registration registration) {
+                onlineStatusCountDown++;
+                assertTrue("Expected online state to be invoked only once but invoked" + onlineStatusCountDown,
+                        onlineStatusCountDown == 1);
+            }
+
+            @Override
+            public void onOffline(Registration registration) {
+                offlineStatusCountDown++;
+                assertTrue("Expected offline state to be invoked only once but invoked" + offlineStatusCountDown,
+                        offlineStatusCountDown == 1);
+            }
+        });
+        // invoke setOnline multiple times
+        instance.setOnline(queueModeRegistration);
+        instance.setOnline(queueModeRegistration);
+        // invoke setOffline multiple times
+        instance.setOffline(queueModeRegistration);
+        instance.setOffline(queueModeRegistration);
+    }
+
+    @Test
+    public void testIsOnline() throws Exception {
+        Registration queueModeRegistration = givenASimpleClientWithQueueMode();
+
+        assertFalse(instance.isOnline(queueModeRegistration));
+        instance.setOnline(queueModeRegistration);
+        assertTrue(instance.isOnline(queueModeRegistration));
+    }
+
+    private Registration givenASimpleClient() throws UnknownHostException {
+        InetSocketAddress address = InetSocketAddress.createUnresolved("localhost", 5683);
+
+        Registration.Builder builder = new Registration.Builder("ID", "urn:client", InetAddress.getLocalHost(), 10000,
+                address);
+
+        return builder.build();
+    }
+
+    private Registration givenASimpleClientWithQueueMode() throws UnknownHostException {
+        InetSocketAddress address = InetSocketAddress.createUnresolved("localhost", 5683);
+
+        Registration.Builder builder = new Registration.Builder("ID", "urn:client", InetAddress.getLocalHost(), 10000,
+                address);
+
+        return builder.bindingMode(BindingMode.UQ).build();
+    }
+}
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
@@ -42,6 +42,7 @@ import org.eclipse.leshan.core.observation.Observation;
 import org.eclipse.leshan.core.request.DownlinkRequest;
 import org.eclipse.leshan.core.response.ErrorCallback;
 import org.eclipse.leshan.core.response.LwM2mResponse;
+import org.eclipse.leshan.core.response.ObserveResponse;
 import org.eclipse.leshan.core.response.ResponseCallback;
 import org.eclipse.leshan.server.Destroyable;
 import org.eclipse.leshan.server.LwM2mServer;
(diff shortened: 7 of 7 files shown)


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/ObservationServiceImpl.java
```diff
@@ -230,6 +232,8 @@ public void onNotification(Request coapRequest, Response coapResponse) {
             return;
         }
 
+        clientStatusTracker.setClientOnline(registration);
```
balsmn: I am not fully convinced of notifying the Queue mode status listeners here because this line will not be invoked if there are no `ObservationListener` registered. But adding this in the very begining means that the registrationStore has to invoked at the very begining.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/QueueModeClientStatusListener.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * this method is invoked when the LWM2M client with the given endpoint state changes from offline to online again.
+     * This listener method will be invoked only once when the state change occurs;i.e:- it will not be invoked when the
+     * previous state of the endpoint is online and further events from the client indicates that the client is still
+     * online.
+     * 
+     * @param endpoint of the client registered.
+     */
+    void onClientOnline(String endpoint);
```
sbernard31: I would pass the `Registration` instead of `endpoint` String and I would rename it in `online()`

sbernard31:
> What qualifies a LWM2M Client to be online (previous state: offline)?

We can start with : _Once we receive a notification for an observation and  when we receive a registration update_ and when we receive a new registration

> What qualifies a LWM2M Client to be offline (previous state: online)?

We can start with : _Once a request timeout_ . In a second time (another PR), we could also consider to set the client offline after a given period of inactivity. (period which could be configurable)

As I propose to remove the singleton pattern for `ClientStatusTraker`, I propose to expose the queuemode API like the other services by adding a `getQueueModeService` or `getPresenceService` on `LwM2mServer`

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/ClientStatusTracker.java
```diff
@@ -0,0 +1,114 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tracks the status of each LWM2M client registered with Queue mode binding. Also ensures that the
+ * {@link QueueModeClientStatusListener} are notified on state changes only for those LWM2M clients registered using
+ * Queue mode binding.
+ *
+ * @see QueueModeClientState
+ */
+public final class ClientStatusTracker {
```
balsmn: No problem. :-)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/ClientStatusTracker.java
```diff
@@ -0,0 +1,114 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(ClientStatusTracker.class);
+    private static ClientStatusTracker instance = null;
+    private final ConcurrentMap<String, QueueModeClientState> clientStatus = new ConcurrentHashMap<>();
+    private final List<QueueModeClientStatusListener> listeners = new CopyOnWriteArrayList<>();
+
+    private ClientStatusTracker() {
+        // Exists to prevent instantiation.
+    }
+
+    public static synchronized ClientStatusTracker getInstance() {
```
balsmn: Is it not already a singleton?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/ClientStatusTracker.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * sets the client identified by the given registration in {@link QueueModeClientState#OFFLINE} mode if the current
+     * state is either {@link QueueModeClientState#ONLINE}. If the state is changed successfully, then the corresponding
+     * {@link QueueModeClientStatusListener} are notified.
+     * 
+     * @param endpoint client
+     */
+    public void setClientOffline(Registration registration) {
```
balsmn: No problem. :-)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/ClientStatusTracker.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * sets the client identified by the given registration in {@link QueueModeClientState.ONLINE} mode if the current
+     * state is either {@link QueueModeClientState#OFFLINE} or <code>null</code>. If the state is changed successfully,
+     * then the corresponding {@link QueueModeClientStatusListener} are notified.
+     * 
+     * @param endpoint client
+     * @return true if the status is successfully changed, false otherwise.
+     */
+    public void setClientOnline(Registration registration) {
```
balsmn: No problem. :-)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/ClientStatusTracker.java
sbernard31: Oops I mean : we don't need to make it a singleton

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/ClientStatusTracker.java
```diff
@@ -0,0 +1,114 @@
...
+        if (registration.usesQueueMode()
+                && (transitState(registration.getEndpoint(), QueueModeClientState.OFFLINE, QueueModeClientState.ONLINE)
+                        || clientStatus.putIfAbsent(registration.getEndpoint(), QueueModeClientState.ONLINE) == null)) {
```
balsmn: No problem. :-)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/ClientStatusTracker.java
balsmn: The reason why it is made singleton is that, no matter where this tracker is used (like `LeshanServer`, `ObservationService`, `RegistrationService `etc.) they all add `PresenceListener` to the same list. If not, we have to make sure that the tracker is instantiated only once manually.

Any problem with singletons?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceTracker.java
```diff
@@ -0,0 +1,116 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(PresenceTracker.class);
+    private static PresenceTracker instance = null;
+    private final ConcurrentMap<String, Presence> clientStatus = new ConcurrentHashMap<>();
+    private final List<PresenceListener> listeners = new CopyOnWriteArrayList<>();
+
+    private PresenceTracker() {
+        // Exists to prevent instantiation.
+    }
+
+    public static synchronized PresenceTracker getInstance() {
```
balsmn: The reason why it is made singleton is that, no matter where this tracker is used (like LeshanServer, ObservationService, RegistrationServiceetc.) they all add PresenceListener to the same list. If not, we have to make sure that the tracker is instantiated only once manually.

Any problem with singletons?

balsmn:
@sbernard31 I see your point regarding not having PresenceTracker as singleton instead integrate it in to LeshanServer and provide access to the tracker through getPresenceService. I am ok with it. 

In that I would like to rename the `PresenceTracker` to `PresenceService`.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceTracker.java
balsmn: I got my answer. Please see my comments below.

sbernard31:
Ok for `PresenceService`

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/QueueModeClientStatusListener.java
balsmn: No problem. :-)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/QueueModeClientStatusListener.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * this method is invoked when the LWM2M client with the given endpoint state changes from online to offline. This
+     * listener method will be invoked only once when the state change occurs;i.e:- it will not be invoked when the
+     * previous state of the endpoint is offline and further events from the client indicates that the client is still
+     * offline.
+     * 
+     * @param endpoint of the client registered.
+     */
+    void onClientOffline(String endpoint);
```
balsmn: No problem., :-)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/QueueModeClientStatusListener.java
```diff
@@ -0,0 +1,43 @@
...
+ * Contributors:
+ *     Bosch Software Innovations GmbH - initial API
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+/**
+ * A listener aware of the status of LWM2M Client using queue mode binding.
+ *
+ */
+public interface QueueModeClientStatusListener {
```
balsmn: No problem. :-)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/QueueModeClientState.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Contributors:
+ *     Bosch Software Innovations GmbH - initial API
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+/**
+ * Possible states that a LWM2M client states that can be tracked for those clients that are using Queue mode binding.
+ */
+public enum QueueModeClientState {
```
balsmn: No problem. :-)

leshan-server-core/src/main/java/org/eclipse/leshan/server/impl/RegistrationServiceImpl.java
```diff
@@ -90,6 +92,7 @@ public void fireUpdated(RegistrationUpdate update, Registration registration) {
         for (RegistrationListener l : listeners) {
             l.updated(update, registration);
         }
+        clientStatusTracker.setClientOnline(registration);
     }
```
balsmn: No problem. :-)

balsmn:
@sbernard31 I am finished with building in your change requests. As you requested, only device online notification is implemented. Offline notification can be done as a separate PR.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceListener.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * this method is invoked when the LWM2M client with the given endpoint state changes from online to offline. This
+     * listener method will be invoked only once when the state change occurs;i.e:- it will not be invoked when the
+     * previous state of the endpoint is offline and further events from the client indicates that the client is still
+     * offline.
+     * 
+     * @param endpoint of the client registered.
+     */
+    void onOffline(Registration endpoint);
```
sbernard31: bad parameter and javadoc

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceListener.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * this method is invoked when the LWM2M client with the given endpoint state changes from offline to online again.
+     * This listener method will be invoked only once when the state change occurs;i.e:- it will not be invoked when the
+     * previous state of the endpoint is online and further events from the client indicates that the client is still
+     * online.
+     * 
+     * @param endpoint of the client registered.
+     */
+    void onOnline(Registration endpoint);
```
sbernard31: bad parameter and javadoc

leshan-server-core/src/main/java/org/eclipse/leshan/server/LwM2mServer.java
```diff
@@ -70,9 +72,8 @@
      * @throws InvalidResponseException if the response received is malformed.
      * 
      */
-    <T extends LwM2mResponse> T send(Registration destination, DownlinkRequest<T> request)
-            throws InterruptedException, CodecException, InvalidResponseException, RequestCanceledException,
-            RequestRejectedException;
+    <T extends LwM2mResponse> T send(Registration destination, DownlinkRequest<T> request) throws InterruptedException,
+            CodecException, InvalidResponseException, RequestCanceledException, RequestRejectedException;
```
sbernard31: formatting issue

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -0,0 +1,61 @@
...
+public interface PresenceService {
+
+    /**
+     * sets the client identified by the given registration in {@link Presence#OFFLINE} mode if the current state is
+     * either {@link Presence#ONLINE}. If the state is changed successfully, then the corresponding
+     * {@link PresenceListener} are notified.
+     * 
+     * @param endpoint client
+     */
+    public void setOffline(Registration registration);
```
sbernard31: I'm not sure we want to expose `setOffline` and `setOnline` in public API ? Should we just keep this in the implementation ?

leshan-server-core/src/test/java/org/eclipse/leshan/server/queue/PresenceServiceTest.java
```diff
@@ -0,0 +1,104 @@
...
+ *
+ */
+public class PresenceServiceTest {
+
+    private PresenceService instance = new PresenceServiceImpl();
+
+    @Test
+    public void testSetOnlineForNonQueueMode() throws Exception {
+        Registration registration = givenASimpleClient();
+        final CountDownLatch countDown = new CountDownLatch(2);
```
sbernard31: should we have a countdownlatch of 1 ?

leshan-server-core/src/test/java/org/eclipse/leshan/server/queue/PresenceServiceTest.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Test method for
+     * {@link org.eclipse.leshan.server.queue.PresenceService#setOffline(org.eclipse.leshan.server.client.Registration)}.
+     * 
+     * @throws Exception
+     */
+    @Test
+    public void testSetOnlineAndOffline() throws Exception {
+        Registration queueModeRegistration = givenASimpleClientWithQueueMode();
+        final CountDownLatch countDown = new CountDownLatch(2);
```
sbernard31: Why using countdownLatch as there is no asynchronous code ?

leshan-server-core/src/test/java/org/eclipse/leshan/server/queue/PresenceServiceTest.java
balsmn: yes... thats correct.

leshan-server-core/src/test/java/org/eclipse/leshan/server/queue/PresenceServiceTest.java
balsmn: Yes that's true. But to make sure that both the methods have been invoked, either you would use some flags or some counter. I am open to both.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
balsmn: I agree. May be we can add a method to get method for the current state in the public API.

sbernard31: This make sense.

leshan-server-core/src/test/java/org/eclipse/leshan/server/queue/PresenceServiceTest.java
sbernard31: I suppose a counter for each state should be ok.
You can also call several time setOffline or setOnline and see that the counter is not incremented.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
balsmn: How about `public boolean isOnline(Registration registration);`

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/Presence.java
```diff
@@ -0,0 +1,26 @@
...
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *     Bosch Software Innovations GmbH - initial API
+ *******************************************************************************/
+package org.eclipse.leshan.server.queue;
+
+/**
+ * Possible states that a LWM2M client states that can be tracked for those clients that are using Queue mode binding.
```
sophokles73: There seems to be a phrasing issue here ...

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceListener.java
```diff
@@ -0,0 +1,45 @@
...
+import org.eclipse.leshan.server.registration.Registration;
+
+/**
+ * A listener aware of the status of LWM2M Client using queue mode binding.
+ *
+ */
+public interface PresenceListener {
+
+    /**
+     * this method is invoked when the LWM2M client with the given endpoint state changes from online to offline. This
```
sophokles73: First letter should be upper-case ...

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceListener.java
```diff
@@ -0,0 +1,45 @@
...
+     * listener method will be invoked only once when the state change occurs;i.e:- it will not be invoked when the
+     * previous state of the endpoint is offline and further events from the client indicates that the client is still
+     * offline.
+     * 
+     * @param registration data of the lwm2m client.
+     */
+    void onOffline(Registration registration);
+
+    /**
+     * this method is invoked when the LWM2M client with the given endpoint state changes from offline to online again.
```
sophokles73: First letter should be upper-case ...

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -0,0 +1,50 @@
...
+ * @see Presence
+ */
+public interface PresenceService {
+
+    /**
+     * add the listener to get notified when the LWM2M client goes online or offline.
+     * 
+     * @param listener target to notify
+     */
+    public void addListener(PresenceListener listener);
```
sophokles73: No need for declaring visibility in an interface ...

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -0,0 +1,50 @@
...
+ * Tracks the status of each LWM2M client registered with Queue mode binding. Also ensures that the
+ * {@link PresenceListener} are notified on state changes only for those LWM2M clients registered using Queue mode
+ * binding.
+ *
+ * @see Presence
+ */
+public interface PresenceService {
+
+    /**
+     * add the listener to get notified when the LWM2M client goes online or offline.
```
sophokles73: uppercase

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -0,0 +1,50 @@
...
+     * 
+     * @param listener target to notify
+     */
+    public void addListener(PresenceListener listener);
+
+    /**
+     * remove the listener previously added. This method has no effect if the given listener is not previously added.
```
sophokles73: uppercase

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -0,0 +1,50 @@
...
+     * 
+     * @param listener target to be removed.
+     */
+    public void removeListener(PresenceListener listener);
+    
+    /**
+     * returns the current status of a given LWM2M client registration.
```
sophokles73: uppercase

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -0,0 +1,50 @@
...
+     * @param listener target to notify
+     */
+    public void addListener(PresenceListener listener);
+
+    /**
+     * remove the listener previously added. This method has no effect if the given listener is not previously added.
+     * 
+     * @param listener target to be removed.
+     */
+    public void removeListener(PresenceListener listener);
```
sophokles73: public ?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -0,0 +1,50 @@
...
+    
+    /**
+     * returns the current status of a given LWM2M client registration.
+     * 
+     * @param registration target registration data to check
+     * @return true if the status is {@link Presence#ONLINE}
+     */
+    public boolean isOnline(Registration registration);
```
sophokles73: public ?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+        listeners.remove(listener);
+    }
+
+    @Override
+    public boolean isOnline(Registration registration) {
+        return Presence.ONLINE.equals(clientStatus.get(registration.getEndpoint()));
+    }
+
+    /**
+     * sets the client identified by the given registration in {@link Presence#OFFLINE} mode if the current state is
```
sophokles73: uppercase

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+                && transitState(registration.getEndpoint(), Presence.ONLINE, Presence.OFFLINE)) {
+            for (PresenceListener listener : listeners) {
+                listener.onOffline(registration);
+            }
+        }
+
+    }
+
+    /**
+     * sets the client identified by the given registration in {@link Presence.ONLINE} mode if the current state is
```
sophokles73: uppercase

balsmn:
@sophokles73 regarding incorrect interface declaration. Did it in a hurry. My bad. Anyways corrected it. Thanks for pointing it out.

Regarding the `Presence` information in `Registration` object, I don't think it makes sense as `Presence` is thought only for LWM2M Clients using "Q" mode binding. Other mode devices would be mostly online and doesn't necessitate tracking. 

In other words, a device not using "Q" mode and setting it to "offline", as per LWM2M Spec, has no meaning. right?

sophokles73:
> Regarding the Presence information in Registration object, I don't think it makes sense as Presence is thought only for LWM2M Clients using "Q" mode binding. Other mode devices would be mostly online and doesn't necessitate tracking.

> In other words, a device not using "Q" mode and setting it to "offline", as per LWM2M Spec, has no meaning. right?

Hmmm, the way you have been discussing it so far, there doesn't seem to be anything special about a device registering with Q mode. When it registers it is considered _online_, when it updates its registration, it is considered _online_. When it deregisters, it is considered _offline_. Sounds like this is true for any kind of device, isn't it?

balsmn:
@sophokles73 

> When it deregisters, it is considered offline.

No, when a device de-registers, no tracking is necessary. Anyways Offline is not part of this PR.

Other that, Yes the way the status change trigger from **offline to online** is not specific to "Q" mode, but the javadoc for `Persistence`, `PersistenceService` and its relevant listener clearly mentions "Q" mode and more over `PersistenceServiceImpl` changes state for the devices only when the "Q" is used. See 

[PersistenceServiceImpl#62](https://github.com/eclipse/leshan/pull/265/files#diff-47ed08b0989553138c4a8f045a01c0cbR62) and [PersistenceServiceImpl#79](https://github.com/eclipse/leshan/pull/265/files#diff-47ed08b0989553138c4a8f045a01c0cbR79)

As this is implementation specific, we can even say, it is generic for a device to have a state (online, offline). So for non-queue mode devices, it could mean that the invocation `PersistenceService#isOnline(registration)` always returns true. 

I am open for both options but we should have a consensus.

sophokles73:
I guess we are talking about `PresenceService`. And yes, the *online* status seems to be a generic property of every device, not just of devices registering with queue mode. For those, however, it seems to be more relevant since those devices are expected to be *offline* most of the time.

balsmn:
Oops...yes...PresenceService.

balsmn:
@sophokles73 @sbernard31 and of course others who are interested in this: I can make the change (integrating `Presence` into `org.eclipse.leshan.server.registration.Registration`) if there are no objections to it.

sbernard31:
I'm ok about using `PresenceService` for both (Q and none Q device).

About moving `Presence` in `Registration`, this make sense for me **only if we want to persist this data**.  Why would we do that ?

sophokles73:
> Why would we do that?

Hmm, I thought we persist the registration info to provide for fail-over between leshan nodes, don't we? Why wouldn't the presence information be helpful in such a case as well?

sbernard31:
1) if  the server restarts,  DTLS connection is lost,  peer is maybe not behind the same IP (NAT timeout), so we probably not able to send request to QueueMode devices anymore ...
2) if this happens we don't lost so much, we can just wait the next device communication.

To be clear, if we really need it, I'm not against to persist it. But this will make the code more complex. (RegistrationStore is already complex) So I would like  to be sure we really need it.

sophokles73:
> To be clear, if we really need it, I'm not against to persist it. But this will make the code more complex. (RegistrationStore is already complex) So I would like to be sure we really need it.

I was actually proposing to include it in Registration in order to *reduce* complexity since we already have a `RegistrationStore` and means to persist it (we are talking about another boolean property, don't we)?
Introducing an additional service actually seems to be more effort to me, in particular if it is about just one boolean property ...

sbernard31:
When store is in memory this is not really a problem (even if Registration is immutable so this means we need to recreate a new object each time state changes).
But for persistent store, this means adding a new read/write access ...

It's not just adding a boolean. You need to add corresponding write operation in the store.
And this will just replace the Map in `PresenceService`.
Adding a service or not is another question ?
Using a boolean instead of an enum is another question too ...

balsmn:
Seems like this topic is in limbo. Discussed with @sophokles73. 

@sophokles73 I am writing down what we discussed. Correct me if I am wrong.

The basic understanding of the LWM2M Spec seems to be different here. According to the Spec, one can say when a device is online, it keeps sending its registration update and when it goes to sleep/offline, it sends a de-registration.

But as per Eclipse Leshan implementation, if a device de-registers (either active or timed out on its registration-update), then the registration info is removed from `RegistrationStore` and Leshan is no more aware of such a device. In this case it doesn't make sense to have the `Presence` boolean flag in `RegistrationStore`.

The question for Eclipse Leshan would be, whether `Registration` info is always a temporary info about the device or whether it could become long-term living object as initially perceived by the spec.

sophokles73:
> The question for Eclipse Leshan would be, whether Registration info is always a temporary info about the device or whether it could become long-term living object as initially perceived by the spec.

I do not think the spec *per se* defines registration information as *persistent* or *long living*. It simply doesn't comment on this aspect at all. However, given the way how leshan is implemented, for me the question is more: do we want to support applications being able to determine a device's *presence* status even when the device is no longer *registered*, i.e. once it has explicitly de-registered or its registration has timed-out? If we do, then we are basically introducing a generic *Device Registry* with device entities having a (persistent) id and an `online` property. Is that we want?

sbernard31:
This discussion is more and more confuse for me ...

> The basic understanding of the LWM2M Spec seems to be different here. According to the Spec, one can say when a device is online, it keeps sending its registration update and when it goes to sleep/offline, it sends a de-registration.

Ok we can imagine that for none Queue mode devices. what is the point here ?

> But as per Eclipse Leshan implementation, if a device de-registers (either active or timed out on its registration-update), then the registration info is removed from RegistrationStore and Leshan is no more aware of such a device. In this case it doesn't make sense to have the Presence boolean flag in RegistrationStore.

I do not understand the point here too.

>The question for Eclipse Leshan would be, whether Registration info is always a temporary info about the device or whether it could become long-term living object as initially perceived by the spec.

I don't think so. The question is just : do we need to persist this information or not ? and I already give my arguments, [here](https://github.com/eclipse/leshan/pull/265#issuecomment-279407686) and [here](https://github.com/eclipse/leshan/pull/265#issuecomment-279442209).

sbernard31:
> Do we want to support applications being able to determine a device's presence status even when the device is no longer registered ?

I don't think so.

sophokles73:
neither do I :-) So if we agree on this, wouldn't it make sense to handle the `online` property within the scope of the registration life-cycle and add it to `Registration`?

sbernard31:
I repeat myself : 
This could make sense but I explained why I have some concerns about persisting it and why persisting it seems not really useful.

sophokles73:
> This could make sense but I explained why I have some concerns about persisting it and why persisting it seems not really useful.

I have read through your earlier comments again and I see your point regarding the lack of usefulness of this information after a fail-over. Agreed :-)
However, I wonder whether it makes sense to keep track of this property for the devices (in the map) at all or if it is not sufficient to simply notify listeners about state changes occurring. In the end we do not really know whether a device is still online after, let's say, 10 minutes after we have received the latest message from it (e.g. its last registration update). So, applications will probably be interested most in the fact that the device is online *right at the time* when its listener has been notified. In practical use cases it will probably also be relevant for the listener to know which event triggered the notification, e.g. whether it was a *register*, *update registration* or *response received*, so that the application can determine whether it is time to send requests to the device or simply pass on the notification.

sbernard31:
I think the event is not enough because : 
if we have an update and 2 notifications during a short part of time, we don't want to raise the event 3 times. 

If we don't provide this service and so we send the event 3 times, this feature bring nothing new from register and observation event.

sophokles73:
But if we have an update and the client goes to sleep after e.g. 1 minute and wakes up 3 hours later and does an update registration then an application won't be notified of this event because the presence service has assumed the client to be *online* all the time in between and thus discards the update registration event (no state transition).

sbernard31:
As explained above, after a configurable time of inactivity the client will be considered as offline. So if this period is less than 3 hours - 1 minutes (I think it should), the application will be notified.

chuchiliusandc:
I have a question ... 
About this Q mode thing, there are no queue anymore for each device now ? 
the queue to save the pending requests ... 

I feel, so far, only the state of the device being discussed ...

Thanks :)

sbernard31:
Please avoid flooding ... you already ask this question at #191   -_-!

IagoFelipeTrentin:
Hello!
I am sorry in advance if this question was already clarified somewhere, but I couldn't find good answers.

- In the current Leshan implementation, what are the differences between the U and UQ modes? I mean, how the Server (or Client) behaves?
I analyzed the code, and so far it seems to be the same way for both modes U or UQ.

- Does (and How) the Server informs the application layer (when request timeout) to say the Client is Sleeping (Offline but Registered)?
By Sleeping I mean it is Alive =  lastUpdate+lifetime > current time.

Thank you!

sbernard31:
> In the current Leshan implementation, what are the differences between the U and UQ modes

If you talked about master, you're right.
If you talked about this PR, look at  [that](https://github.com/eclipse/leshan/pull/265/commits/7de1d9142a63db68c66698c1f98d8cda64161dd7#diff-47ed08b0989553138c4a8f045a01c0cbR79)

> Does (and How) the Server informs the application layer (when request timeout) to say the Client is Sleeping (Offline but Registered)?

In master, there is nothing about it.
In this PR, this is  the [offline event](https://github.com/eclipse/leshan/pull/265/commits/7de1d9142a63db68c66698c1f98d8cda64161dd7#diff-1264d9217a37c7edb9661653096a3383R27), but this is not fully implemented see [this comment](https://github.com/eclipse/leshan/pull/265#issuecomment-277318077)

sbernard31:
@balsmn  do you still plan to contribute this PR or this is not needed anymore ? I just ask to know :)

ming2377:
Seems new PresenseService works in single DM node. Any plan to support Cluster, maybe moving to RegistrationStore makes sense, so that RedisRegistrationStore can store this very much shared-state among DM clustering nodes.

ming2377:
Seems `presenceService.setOnline(updatedRegistration)` are called in the same COAP thread executes `RegistationListener::updated`, `ObservationListener::onResponse`, and `ObservationListener::onError`.

We recently run into thread blocking issue when we execute code directly in these listener methods, DM may become non-responsive, we resolved the issue by using Executors.newCachedThreadPool. There is an example ObservationListener in RedisRequestResponseHandler that uses the same strategy.

sbernard31:
Thx to @joakimeriksson and mainly @carlosgp143,  server queue mode implementation is relaunched.
(see #427 and #451 for more details)

This PR will be used as base for this work and so this is integrated [queue_mode](https://github.com/eclipse/leshan/tree/queue_mode) branch.

Thx a lot @balsmn  for your contribution.

PR #272, opened 2017-02-11
Use StandardCharsets

scop:


sbernard31:
Integrated in master (commit 4c8e9fd084a6296a9514f965c77730c21a2a8755)
(I just keep javadoc in StringUtils because this is a copy of apache commons, and I would like to limite modifications as much as possible)

PR #275, opened 2017-02-15
Add discover operation support in REST API

scop:


sophokles73:
Thanks for the submission. Going through the code I noticed that you haven't added any particular functionality to the web UI making use of the *discovery* feature. We have always considered the demo application to be exactly that: a *demo* application only, i.e. this is not intended to be used for production. As such, IMHO we only add code that makes helps making it a better *demo* application. There also has been some discussion around the need for a *real*, i.e. *production quality* REST API for leshan in the past. However, there never has been enough demand nor community support to make it happen.
So, if this is your way of stating that you would like to see a REST API for leshan evolving, then I think we should start a discussion of how to approach it again :-) Extending the *demo* application for that purpose is not the right approach FMPOV.

scop:
I have a plan to extend the demo UI to make use of the discovery feature later (e.g. disabling UI elements that discovery finds unsupported); this paves way for it and was already useful in its current form in examining/testing client discovery behavior in another use case I had, which is why I thought I'd submit it already now.

But sure, a non-demo REST API would be a nice thing to have :)

sbernard31:
This seems to be a small change and open the way to add UI for discovery feature. I think we can integrate it.

sophokles73:
Ok, then go ahead with merging, @sbernard31 :-)

sbernard31:
Done. (In master commit 73237f09c17b9d1bd09f28ffe0d5019c883fe00e)

PR #279, opened 2017-02-17
Remove unnecessary use of com.sun.corba API

scop:
It seems very unlikely to me that com.sun.corba (private API) Exceptions would be thrown by the invoked code. I suppose the addition of this in 521bf72739f7f0e6a6ff1fb4ef0697ba7a561a27 was unintentional.


leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/servlet/ClientServlet.java
```diff
@@ -167,10 +166,6 @@ private void handleException(Exception e, HttpServletResponse resp) throws IOExc
             LOG.warn("Request rejected", e);
             resp.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
             resp.getWriter().append("Request rejected:").append(e.getMessage()).flush();
-        } else if (e instanceof RequestCanceledException) {
```
sophokles73: The code is required, it's just the import that is wrong (probably due to erroneous auto completion).
What we need to import is `org.eclipse.leshan.core.request.exception.RequestCanceledException` ...

scop:
Ah! Changed to just fix the import.

PR #282, opened 2017-02-20
Security test base work

scop:
Some trivial base work to security tests, making them closer to working when client RPK/X509 support is testable.

### DIFF
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/IntegrationTestHelper.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/IntegrationTestHelper.java
@@ -150,6 +150,10 @@ public class IntegrationTestHelper {
         builder.setSecurityStore(new InMemorySecurityStore());
         server = builder.build();
         // monitor client registration
+        setupRegistrationMonitoring();
+    }
+
+    protected void setupRegistrationMonitoring() {
         resetLatch();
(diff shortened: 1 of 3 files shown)


sbernard31:
I'm ok with your cleaning commits, but I'm not sure activate those tests was a good Idea. Client are currently not able to connect using RPK or X509. So tests succeed but for bad reason. (It succed because client failed to do an handshake correctly and not because server refuse to authorize it)

scop:
Even though these tests succeed for kind of a wrong reason, enabling them is good in the sense that doing so may pinpoint failures and regressions in the related test framework level code. For example, `createServerWithRPK` and `createServerWithX509Cert` failed to do necessary initialization needed to track registrations (that's fixed in the "Set up registration monitoring ..." commit), and that would not have been noticeable without enabling some of the tests being discussed.

Anyway, just confirm if you'd still like these tests to be ignored again for now and I'll remove it.

sbernard31:
I agree but in other hand, If we don't use it it's not a big deal if there is issue with it. We could even consider to remove it.  We didn't do it because the code is mainly ok we just need a way in scandium to access to rpk and x509 certificate dynamically.

I will integrate your 3 first commits. Thx again @scop

sbernard31:
Integrated in master (commits c0e993d45990aaa1734395be9a4999d7ee87bdc6, dd0640b6670a1925899abd6394da0ac8751f0450, a28335455b8aad2643aa3fa9e6a27e82c6101732)

PR #286, opened 2017-03-01
Integration of Californium 2.0.0-M3

sbernard31:


sophokles73:
LGTM

leshan-core/src/main/java/org/eclipse/leshan/util/Key.java
```diff
@@ -0,0 +1,39 @@
+package org.eclipse.leshan.util;
```
msangoi: `util` package? why not moving it to`org.eclipse.leshan.server.californium.impl` if it used only there.

sbernard31: I put it here because this is not californium dependent and this can be reused for anyone we want to use byte array as key for a map.

sophokles73: +1 for keeping it in `util` package

PR #288, opened 2017-03-02
Add a way to configure CoAP network config

sbernard31:
Now you can provide a CoAP network config for Leshan server, bootstrap server and client.

1) By default we use the Californium default one.
2) DTLS configuration (max connections, stale connection threshold) are synchronized with coap network configuration (max active peers, max peer inactivity period). 
3) A file is not created or loaded anymore, if users want to do that it should load/store itself like this :

```java
builder.setNetworkConfig(NetworkConfig.getStandard());
```

### DIFF
--- a/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/LeshanBootstrapServerDemo.java
+++ b/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/LeshanBootstrapServerDemo.java
@@ -26,6 +26,7 @@ import org.apache.commons.cli.DefaultParser;
 import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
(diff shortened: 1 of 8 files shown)


sbernard31:
Integrated in master (commit 9552410, ac5739d, 05c623d)

cake808:
This doesn't seem to work on the latest leshan-server-demo implementation.

E.g. I'm not able to change the `COAP_PORT` in `Californium.properties`. 
It still uses the default Port when running the leshan-server-demo.

sbernard31:
This never worked.

I didn't even notice there was some COAP_PORT, COAP_SECURE_PORT in the network config (Californium.properties)

For leshan demos, the current expected way is to use : 
```
 -lp,--coapport <arg>           Set the local CoAP port.
                                Default: 5683.
 -slp,--coapsport <arg>         Set the secure local CoAP port.
                                Default: 5684.
```

sbernard31:
@cake808, if you can tell me if #847 sounds OK to you.

PR #292, opened 2017-03-17
Better support of empty payload for TLV and JSON format

sbernard31:
We currently doesn't support correctly empty payload for TLV and JSON format.
This PR fixes it and adds some tests about that.


sbernard31:
Integrated in master ( commit afed4d7b610356136438d42d87a7cb3a7566867c)

PR #293, opened 2017-03-17
Redis registration store cleanup

msangoi:
* key renaming
* iterator cleanup


jvermillard:
rationale for renaming keys?

msangoi:
Simply homogenous naming. 
Before the change, keys for registrations and observations had a different pattern (val:key or key:val).

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
```diff
@@ -265,11 +280,7 @@ public Registration next() {
                 if (!hasNext()) {
```
sbernard31: This `try (Jedis j = pool.getResource()){}` seems not useful.

msangoi:
Good catch @sbernard31. updated. thx.

sbernard31:
Integrated in master (commit f6aa2ac32ffa5368b975025a5cc18d95a22364cc)

PR #297, opened 2017-03-21
Use internal Validate instead of commons-lang one

scop:
Signed-off-by: Ville Skyttä <ville.skytta@iki.fi>

### DIFF
--- a/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapStoreImpl.java
+++ b/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapStoreImpl.java
@@ -26,10 +26,10 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
-import org.apache.commons.lang.Validate;
 import org.eclipse.leshan.server.bootstrap.BootstrapConfig;
 import org.eclipse.leshan.server.bootstrap.BootstrapStore;
 import org.eclipse.leshan.server.bootstrap.demo.ConfigurationChecker.ConfigurationException;
+import org.eclipse.leshan.util.Validate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 


leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapStoreImpl.java
```diff
@@ -26,10 +26,10 @@
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
-import org.apache.commons.lang.Validate;
 import org.eclipse.leshan.server.bootstrap.BootstrapConfig;
 import org.eclipse.leshan.server.bootstrap.BootstrapStore;
 import org.eclipse.leshan.server.bootstrap.demo.ConfigurationChecker.ConfigurationException;
+import org.eclipse.leshan.util.Validate;
```
sophokles73: Does this also let us get rid of the dependency on `commons-lang` altogether? If so, we should also remove it from the pom file.

scop: Not yet, unfortunately. There are still some StringUtils and ArrayUtils dependencies.

sbernard31: We try to limit dependencies to Leshan that's why we duplicate some apache commons classes.
But we only target jar which is intended to be reused. (leshan-core, leshan-server-core, leshan-server-cf ...)

For demo jar, I think it's better to add a maven dependency.

So in this case, using internal or apache commons validate does really matter.

About dependency, here leshan-server-cf :
```
[INFO] org.eclipse.leshan:leshan-server-cf:bundle:1.0.0-SNAPSHOT
[INFO] +- org.eclipse.leshan:leshan-core-cf:jar:1.0.0-SNAPSHOT:compile
[INFO] |  \- org.eclipse.leshan:leshan-core:jar:1.0.0-SNAPSHOT:compile
[INFO] |     \- com.google.code.gson:gson:jar:2.2.4:compile
[INFO] +- org.eclipse.leshan:leshan-server-core:jar:1.0.0-SNAPSHOT:compile
[INFO] +- org.eclipse.californium:californium-core:jar:2.0.0-M3:compile
[INFO] |  \- org.eclipse.californium:element-connector:jar:2.0.0-M3:compile
[INFO] |     \- io.netty:netty-all:jar:4.1.1.Final:compile
[INFO] +- org.eclipse.californium:scandium:jar:2.0.0-M3:compile
```
I think if we can remove the netty-all (~2,6MB) dependency this could be a good think knowing we don't really need it.

sbernard31: The netty dependency issue is a bit off topic.
Using apache commons or internal Validate does not change so mush so if you prefer this dependency. I will integrated in master.

sbernard31:
integrated in master (commit ebc4efc20619232ae08679b6a27320201d913f8e )

Issue #280, opened 2017-02-17
Use v1.0.0 object models of the LWM2M specication.


sbernard31:
Now the v1.0.0 of the LWM2M specification is out we should integrate the last object descriptions.

Those descriptions seems available [here](https://raw.githubusercontent.com/OpenMobileAlliance/openmobilealliance.github.io/master/OMNA/LwM2M/DDF.xml).  (tell us if this is not the right source)

We can use the downloader(#273 ) written by @scop to do that.

Some questions remain : 
1)Which objects should we embedded by default ?
2)We originally create our .json format because of license issue with ddf files (.xml). Now this files are licensed under BSD, should we still use our json format? should we embedded ddf files directly ?

sbernard31:
1) Personally, I think we should just keep the minimum set of object description in `leshan-core` (only the one described in the core spec). And add all the others in our demo package. Knowing that users can add their models. Using some configurations.
2) The only reason I see to keep json file could be that maybe this is less verbose, but I bet that after jar compression there is not so much differences.

jvermillard:
I think we need to fill a IP CQ for the DDF files..
but yes I see no reason to keep the json files

boaks:
> 1. Personally, I think we should just keep the minimum set of object description in leshan-core (only the one described in the core spec). And add all the others in our demo package. Knowing that users can add their models. Using some configurations.

Will the "eclipse sandbox" then still support the IPSO stuff (maybe even more IPSO objects than now, e.g. 3335)?

sbernard31:
Yes it will as the "eclipse sandbox" is just an instance of leshan-server-demo. So I propose that  all the IPSO stuff would be supported (even 3335) in our demo.

boaks:
Sounds great, we are looking forward to that.

sbernard31:
Fixed since #299 is integrated in master.

PR #299, opened 2017-03-21
Update model, load only core objects by default, demos load more objects

sbernard31:
Following #280 :
1) we use ddf format,  we don't use json format in jar anymore (but we still support it)
2) we use the last version of object definition.
3) Code library only loads core object definition (security, server, access control, device, connectivity monitoring, firmware, location, connectivity statistics)
4) Demos (server and client) now load objects define [here](https://raw.githubusercontent.com/OpenMobileAlliance/openmobilealliance.github.io/master/OMNA/LwM2M/DDF.xml)


scop:
The `MyLocation` changes from #300 should be included here too, I suppose? Or I can take care of looking at updating them afterwards once this is in.

sbernard31:
@scop, it's done. Thx you :)

leshan-core/src/main/java/org/eclipse/leshan/core/model/ObjectLoader.java
```diff
@@ -107,6 +105,52 @@ public static ObjectModel loadDdfFile(InputStream input, String streamName) {
...
+    /**
+     * Load object definition from DDF files.
+     * 
+     * @param path directory path to the DDF files
+     * @param filenames names of all the DDF files
+     */
+    public static List<ObjectModel> loadDdfFiles(String path, String[] filenames) {
+        List<ObjectModel> models = new ArrayList<>();
+        for (String filename : filenames) {
+            String fullpath = path + filename;
```
msangoi: Should be resilient to trailing '/' in path.

leshan-core/src/main/java/org/eclipse/leshan/core/model/ObjectLoader.java
```diff
@@ -39,6 +39,11 @@
 
     private static final Gson GSON;
 
+    private static String[] ddfpaths = new String[] { "LWM2M_Security-v1_0.xml", "LWM2M_Server-v1_0.xml",
+                            "LWM2M_Access_Control-v1_0.xml", "LWM2M_Device-v1_0.xml",
+                            "LWM2M_Connectivity_Monitoring-v1_0.xml", "LWM2M_Firmware_Update-v1_0.xml",
+                            "LWM2M_Location-v1_0.xml", "LWM2M_Connectivity_Statistics-v1_0.xml" };
```
msangoi: What about software object? and objects which are parts of additional specs (e.g. portofolio 😄) ?

leshan-core/src/main/java/org/eclipse/leshan/core/model/ObjectLoader.java
jvermillard: or `\`

leshan-core/src/main/java/org/eclipse/leshan/core/model/ObjectLoader.java
sbernard31: As you said, this is additional specs. So not embedded by default, but you can easily add it manually. (like we do in demos app).

msangoi: 👍

leshan-core/src/main/java/org/eclipse/leshan/core/model/ObjectLoader.java
sbernard31: We use `class.getResourceAsStream` so `/` should be used not `\` (see [javadoc](https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html#getResource%28java.lang.String%29)).

Maybe our javadoc is not clear, we should explain that we use `class.getResource()` and point the javadoc which explain this is for loading resource _in a way that is independent of the location of the code._ (loading resource in jar or war file)

About resilience of trailing `/`, I can add it but I found this is not so usefull. This method is more for loading statically resource you embedded in your jar or war. So you will use constant and quickly see if you can not load your files. (because of illegalStateException)

If you want something more dynamic (the case you don't embedded your models in your jar) you can use [loadObjectsFromDir](https://github.com/eclipse/leshan/pull/299/files/54053b03413fcbc4361ffda5b9dbe40216198581#diff-449a4a63d5ac98e293fecf026aaf8165R157) which should be public now.

sbernard31: I changed the code to support trailing `/`.
I renamed `loadDdfFiles` in `loadDdfResources` to be more explicit.
I also add some javadoc and link to `class.getResourceAsStream`

sbernard31:
Integrated in master commit (9b531b386a11df83bd48eb3d370eb281648448f1)

Issue #208, opened 2016-11-29
Removing environment variable from all Leshan command line tools


sbernard31:
see #197

sbernard31:
Fixed since #301 is integrated in master.

PR #301, opened 2017-03-22
Do not use env var anymore to configure Leshan.

sbernard31:
Removing the use of MODELS_FOLDER environment variable.
1) If Leshan is used as a Library :  builder, constructor and ObjectLoader helper  are enough to customize LWM2M model.
2) For demos, I add a new option to be able to set the path of folder including model files.

Then removing all environment variable of demos as described in #208.

### DIFF
--- a/leshan-server-cluster/src/main/resources/log4j2.xml
+++ b/leshan-server-cluster/src/main/resources/log4j2.xml
@@ -32,6 +32,9 @@ Contributors:
     <logger name="org.eclipse.leshan.server.security.SecurityCheck" level="trace" additivity="false">
         <AppenderRef ref="STDOUT"/>
     </logger>
+    <logger name="org.eclipse.leshan.core.model.LwM2mModel" level="trace" additivity="false">
+        <AppenderRef ref="STDOUT"/>
+    </logger>
     <Root level="warn">
       <AppenderRef ref="STDOUT"/>
     </Root>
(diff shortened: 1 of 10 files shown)


leshan-server-cluster/src/main/resources/log4j2.xml
```diff
@@ -32,6 +32,9 @@ Contributors:
     <logger name="org.eclipse.leshan.server.security.SecurityCheck" level="trace" additivity="false">
         <AppenderRef ref="STDOUT"/>
     </logger>
+    <logger name="org.eclipse.leshan.core.model.LwM2mModel" level="trace" additivity="false">
```
msangoi: There is no trace logs in the LwM2mModel class (almost no logs at all).
You mean ObjectLoader?

sbernard31: Actually, I would like to report to demos app users  when they override an embedded model.

I want this [log](https://github.com/eclipse/leshan/blob/master/leshan-core/src/main/java/org/eclipse/leshan/core/model/LwM2mModel.java#L48). This is `debug` not `trace`. As `trace` include `debug`, it works. But I can change it to avoid confusion.

sbernard31:
Integrated in master (commit e017ef40759342dcbaad5667406fc532db30906d, 4bc29cccf72690cc71bcdb714a4ffbc591784d18, 892f1eff368bfec617ffdca2aba536aede7bc1b9)

PR #303, opened 2017-03-25
Misc trivial fixes and cleanups

scop:


leshan-core/src/main/java/org/eclipse/leshan/core/request/CreateRequest.java
```diff
@@ -216,7 +216,7 @@ private CreateRequest(ContentFormat format, LwM2mPath target, Integer instanceId
             if (instanceId == null) {
                 instanceId = target.getObjectInstanceId();
             } else {
-                if (instanceId != target.getObjectInstanceId()) {
+                if (instanceId.equals(target.getObjectInstanceId())) {
```
sbernard31: you meant !equals ?

leshan-core/src/main/java/org/eclipse/leshan/json/JsonArrayEntry.java
```diff
@@ -155,7 +157,7 @@ public boolean equals(Object obj) {
                 return false;
         } else if (!stringValue.equals(other.stringValue))
             return false;
-        if (time != other.time)
```
sbernard31: There is an mistake here, this method was modified manually. We should regenerated with eclipse : 
```java
if (time == null) {
    if (other.time != null)
        return false;
} else if (!time.equals(other.time))
    return false;
```

leshan-core/src/main/java/org/eclipse/leshan/json/JsonArrayEntry.java
```diff
@@ -132,30 +132,15 @@ public boolean equals(Object obj) {
         if (getClass() != obj.getClass())
             return false;
```
sbernard31: This equals methods is generated code. I'm not sure we want to modify/simplify it ?

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
```diff
@@ -185,8 +185,6 @@ public static void createAndStartServer(int webPort, String localAddress, int lo
         }
 
         // Get public and private server key
-        PrivateKey privateKey = null;
-        PublicKey publicKey = null;
```
sbernard31: `publicKey` is used below out of the try/catch block : compilation error.

leshan-core/src/main/java/org/eclipse/leshan/core/request/CreateRequest.java
scop: Ah, yes.

leshan-core/src/main/java/org/eclipse/leshan/json/JsonArrayEntry.java
scop: Will remove this change.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
scop: Oops, sorry about that.

scop:
Thanks for the comments, all should be addressed now.

sbernard31:
Integrated in master. Thx @scop

PR #304, opened 2017-03-29
Add expired parameter to the unregistered event

msangoi:
Add a parameter to determine whether a client has been unregistered because of its lifetime expiration.


sbernard31:
:+1:

PR #305, opened 2017-03-30
Destroy client and server on IntegrationTestHelper dispose().

boaks:
Executing the test on a (too) small VM reports problems creating new
threads. Destroying the client and server at the end of the test frees
that resources and the test are able to run.

### DIFF
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/CreateTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/CreateTest.java
@@ -13,6 +13,7 @@
  * Contributors:
  *     Zebra Technologies - initial API and implementation
  *     Achim Kraus (Bosch Software Innovations GmbH) - add test for create security object
+ *     Achim Kraus (Bosch Software Innovations GmbH) - replace close() with destroy()
  *******************************************************************************/
 
 package org.eclipse.leshan.integration.tests;
@@ -54,8 +55,8 @@ public class CreateTest {
 
     @After
     public void stop() {
-        helper.client.stop(false);
-        helper.server.stop();
+        helper.client.destroy(false);
+        helper.server.destroy();
         helper.dispose();
     }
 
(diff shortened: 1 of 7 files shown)


leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/IntegrationTestHelper.java
```diff
@@ -246,6 +248,12 @@ public void deregisterClient() {
 
     public void dispose() {
         deregisterClient();
+        if (null != client) {
+            client.destroy(false);
+        }
+        if (null != server) {
+            server.destroy();
+        }
```
sbernard31: We already call `stop()` in all tests. So we should replace it by `destroy()` or remove all the `stop()`call and `destroy` in dispose. Personally, I prefer the first solution as this make more sense to me to create in @before and destroy in @after.

boaks: No objections, unless it's ensured, that `destroy`  is called ;-).

boaks: Using @after would cause changes in many places. Currently I started to use JUnit Rule in californium. May be such a Rule can do a better job then @before and @after.

sbernard31: In all case, you need to change @after : to remove `stop()` or to replace it by `destroy()`.  (I do a call hierarchy, we should change it at 9 locations.)

About using Rule, you can do that in a second step, if you think this can simplify tests.

boaks: It's not forbidden to call stop and additionally destroy.
But I will go to change the PR.

boaks:
Now with destroy() in @after.

sbernard31:
Integrated in master (commit 83fbb9f8c570370dae4543c070dcaa239cf2c5dd)
I add missing destroy in DiscoveryTest, SecurityTest, BootstrapTest.

PR #311, opened 2017-03-31
Remove secondary index on ADD if endpoint already register in redisStore

sbernard31:
We didn't remove secondary index on ADD if endpoint already register in redisStore.
It should be fixed now.


sbernard31:
Integrated in master (commit 660201d19440cbae5cbb7ee12245d695d420a8ff)

PR #314, opened 2017-04-04
Add javadoc to LeshanServerBuilder API.

sbernard31:


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
```diff
@@ -62,6 +68,17 @@
...
+    /**
+     * <p>
+     * Set the address/port for unsecured CoAP Server.
+     * </p>
+     * 
+     * By default a wildcard address and the default CoAP port(5683) is used
+     * 
+     * @param hostname The address to bind. If null wildcard address is used.
+     * @param port A valid port value is between 0 and 65535. A port number of zero will let the system pick up an
+     *        ephemeral port in a bind operation.
```
sophokles73: add a `@throws IllegalArgumentException` for port out of legal range?
add a `@return` tag?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
```diff
@@ -71,11 +88,29 @@ public LeshanServerBuilder setLocalAddress(String hostname, int port) {
         return this;
     }
 
+    /**
+     * <p>
+     * Set the address for unsecured CoAP Server.
+     * </p>
+     * 
+     * By default a wildcard address and the default CoAP port(5683) is used.
+     */
```
sophokles73: `@return` ?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
```diff
@@ -71,11 +88,29 @@ public LeshanServerBuilder setLocalAddress(String hostname, int port) {
...
+    /**
+     * <p>
+     * Set the address/port for secured CoAP Server (Using DTLS).
+     * <p>
+     * 
+     * By default a wildcard address and the default CoAPs port(5684) is used.
+     * 
+     * @param hostname The address to bind. If null wildcard address is used.
+     * @param port A valid port value is between 0 and 65535. A port number of zero will let the system pick up an
+     *        ephemeral port in a bind operation.
```
sophokles73: `@throws IllegalArgumentException`?
`@return`?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
```diff
@@ -85,61 +120,133 @@ public LeshanServerBuilder setLocalSecureAddress(String hostname, int port) {
         return this;
     }
 
+    /**
+     * <p>
+     * Set the address for secured CoAP Server (Using DTLS).
+     * </p>
+     * 
+     * By default a wildcard address and the default CoAP port(5684) is used.
```
sophokles73: `@param localSecureAddress`?
`@return`?

jvermillard:
Personally, I really hate the formalism of Javadoc, all the `@param` and `@return` are seen as mandatory by developers and most of the time are just noise stating the obvious so they can declare success and brag the javadoc is complete.

So I tend to prefer to have a doc stating the purpose of the class and how to use it than repeating endlessly that the builder setter @return the builder for chaining operation. It's just cognitive load on the reader.

If find golang approach on doc more pleasant to read:

```
// Fprint formats using the default formats for its operands and writes to w.
// Spaces are added between operands when neither is a string.
// It returns the number of bytes written and any write error encountered.
func Fprint(w io.Writer, a ...interface{}) (n int, err error) {
```
https://blog.golang.org/godoc-documenting-go-code

I'm going to introduce a builder for the bootstrap server so I prepare you to face my javadoc where I ignore tags for which I have nothing useful to say :)

sophokles73:
I guess everybody has an opinion :-) Anyway, I did not want to impose a particular style to be followed, I just wanted to point out that so far there doesn't seem to be any style being, well, defined at all. So, if the majority thinks that it is not worth following the recommendations for JavaDoc, I am ok with that as well. So, I guess this is then the (non-)definition of the JavaDoc style for leshan, assuming that the *French Connection* is assembled behind @jvermillard 's take on it ;-)

sbernard31:
Integrated in master (commit 82ba13ca7be3a584fe30c57dfdd0f95c1f5f534c)

PR #326, opened 2017-04-13
Rename serverPublicKeyOrId to serverPublicKey to align with spec

scop:
This is strictly speaking a backwards incompatible API change. Would you like me to add some deprecation/backwards compatibility code to handle it?


sbernard31:
No it's ok about API break for now. (Even if we try to limit that... maybe without success ^-^! )
Integrated in master (commit 33db7ab894751d040cdce40680c36278f280a701)

PR #335, opened 2017-04-28
Use ObservationUtil shallowClone() instead of serialize/parse.

boaks:


sophokles73:
LGTM

PR #350, opened 2017-06-12
Add a way to tweak DTLS configuration.

sbernard31:
Currently, there is no way to configure DTLS like changing the retransmissions timeout. This PR should fix that.

This is clearly not the more user-friendly way to do, but this is a first step.

### DIFF
--- a/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/LeshanBootstrapServerDemo.java
+++ b/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/LeshanBootstrapServerDemo.java
@@ -20,7 +20,6 @@ package org.eclipse.leshan.server.bootstrap.demo;
 
 import java.io.File;
 import java.net.BindException;
-import java.net.InetSocketAddress;
 import java.util.List;
 
 import org.apache.commons.cli.CommandLine;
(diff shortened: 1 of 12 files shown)


sbernard31:
@boaks, @sophokles73, as you can see in this [PR](https://github.com/eclipse/leshan/pull/350/commits/98d2d90b71ce4f50df0934768f40d4d2df3615bf#diff-ab2b2333ae602c4ca955af3e2805c488R284), I didn't find an easy way to let Leshan users to tweak `DTLSConfig`.  They must initialize some stuff we should do for them (like create the `LwM2mPskStore`).

I would like to do better, like letting users set some basic stuff via the `DTLSconfigBuilder` (e.g. retransmission timeout) then complete what is missing in `LeshanBuilder` (e.g. PskStore). That's means that users should be able to create an incomplete DTLSConfig (which is not possible currently) or I need to know which fields is set in DTLSConfigBuilder (which is not possible too). Any Ideas ? Does it make sense for you to modify that in californium ?

boaks:
@sbernard31 

I didn't get it. Why not just using the `DtlsConnectorConfig.Builder` in the `LeshanServerBuilder`?
Should we add a `DtlsConnectorConfig.Builder` contructor with a `DtlsConnectorConfig` as parameter?

sbernard31:
That was my first idea but currently there is no way to create an inconsistent config because there is some checks on `DtlsConnectorConfig.build()`. E.g.  if user chose to use TLS_PSK_WITH_AES_128_CCM_8 cipher and does not set a pskstore, we will not be able to create the config to pass to the new constructor you want to introduce.

boaks:
And using the `DtlsConnectorConfig.Builder` itself as parameter?

sbernard31:
I think about that too, but this time we don't know what is set by the users and so what we should add :'(

boaks:
So your idea is to have a "incomplete configuration" and you add the missing stuff?
Or do you want to add "selected" stuff?

sbernard31:
I'm not sure to see the exactly what you mean by "selected stuff", but "having a incomplete configuration and adding the missing stuff" looks like what I have in mind.

boaks:
With "selected" I mean, e.g. "only PSKStore and/or certificates" related to the already provided cipher suites.
(Only a subset will be filled in, if missing).

sbernard31:
I probably need to only fill a subset. Does this nuance really matter ?

boaks:
In my opinion, this defines how the "user" will get it presented.

Should we go for getter in the Builder?

sbernard31:
Having getter in the DTLSConnectorBuilder could do the job.

boaks:
@sbernard31 

Please check, if

https://github.com/eclipse/californium/pull/320

would work.

sbernard31:
This would be better but I can not know if `maxConnections` or  `staleConnectionThreshold`
A flexible solution could be to make all Config parameters nullable and adding getter for all builder setter or even just a getter on the incomplete config.

This could allow to add missing parameters or even check if parameters seems good and raise warning if necessary.

boaks:
Next try.

sbernard31:
As https://github.com/eclipse/californium/pull/336 is accepted, this PR can be closed, because outclassed by #356.

PR #353, opened 2017-06-19
Reuse already unmodifiable map of additional registration attributes.

boaks:
Creating a "unmodifiable map" with each registration update will build a
very long list of unmodifiable map (see current implementation of
Collections.unmodifiableMap()). Therefore change the implementation to
create such a unmodifiable map only, if the map is not already
unmodifiable.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
@@ -12,6 +12,9 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
+ *     Achim Kraus (Bosch Software Innovations GmbH) - reuse map of
+ *                                                     additionalRegistrationAttributes
+ *                                                     if already unmodifiable.
  *******************************************************************************/
 package org.eclipse.leshan.server.registration;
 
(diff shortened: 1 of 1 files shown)


boaks:
An alternative approach would be to change the Builder instead of using a modifiable test.

sbernard31:
```java
if (additionalRegistrationAttributes == null || additionalRegistrationAttributes.isEmpty()) {
    this.additionalRegistrationAttributes = Collections.emptyMap();
} else {
    this.additionalRegistrationAttributes = Collections
        .unmodifiableMap(new HashMap<>(additionalRegistrationAttributes));
}
```
What  about that ?

boaks:
The empty part looks good, but why 

`Collections.unmodifiableMap(new HashMap<>(additionalRegistrationAttributes))`

?

sbernard31:
I thought that your problem was that after several updates we will face something like an UnmodifiableMap which encapsulates and UnmodifiableMap which encapulates and UnmodifiableMap which .... which encapsulate the "real" map. (only for in memory registrationStore by the way)

This `Collections.unmodifiableMap(**new HashMap**<>(additionalRegistrationAttributes))` will avoid that.

boaks:
Ah, OK! 
So, could/do you fix it in your way? I would appreciated that.

sbernard31:
Ok I will do that.

PR #354, opened 2017-06-19
Fix several issue about immutability

sbernard31:
The main problem is the memory leak #353 raised by @boaks and fixed by 639f5f1.


boaks:
Great!
I think it's worth to do the same for californium. Should I go for?

sbernard31:
> Should I go for?

If I well remember there is not so much immutable class in Californium, but you can have a look.

msangoi:
LGTM 👍

PR #355, opened 2017-06-20
Simple MessageIdProvider (memory consumption fix)

msangoi:
A basic MessageIdProvider implementation using a simple counter for each peer to get the next messageId.


sbernard31:
You should add it to `nonSecureEndpoint` endpoint too.

sbernard31:
👍

PR #363, opened 2017-07-04
Use NullMessageIdTracker from Cf instead of SimpleMessageIdProvider

sbernard31:
https://github.com/eclipse/californium/issues/323 is fixed we can remove SimpleMessageIdProvider introduce in #355.


leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
```diff
@@ -122,6 +122,7 @@ public LeshanClient build() {
         }
         if (networkConfig == null) {
             networkConfig = new NetworkConfig();
+            networkConfig.set(NetworkConfig.Keys.MID_TRACKER, "NULL");
```
msangoi: It would be nice to have the enum `org.eclipse.californium.core.network.InMemoryMessageIdProvider.TrackerMode` public.

msangoi:
LGTM 👍 
The tracker must be set for the bootstrap server as well. Probably part of another PR.

sbernard31:
Integrated in master

PR #375, opened 2017-08-29
Limit number of Threads used by Leshan client by default.

sbernard31:


leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
```diff
@@ -123,6 +123,16 @@ public LeshanClientBuilder setDtlsConfig(DtlsConnectorConfig.Builder config) {
         return this;
     }
 
+    public static NetworkConfig createDefaultNetworkConfig() {
+        NetworkConfig networkConfig = new NetworkConfig();
+        networkConfig.set(Keys.MID_TRACKER, "NULL");
+        networkConfig.set(Keys.MAX_ACTIVE_PEERS, 10);
+        networkConfig.set(Keys.MAX_MESSAGE_SIZE, 4096);
```
msangoi: MAX_RESOURCE_BODY_SIZE instead?
I don't know what is a "good" default value, so I would use the Californium default one.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
```diff
@@ -123,6 +127,16 @@ public LeshanClientBuilder setDtlsConfig(DtlsConnectorConfig.Builder config) {
         return this;
     }
 
+    public LeshanClientBuilder noUnsecureEndpoint() {
```
msangoi: disableSecureEndpoint() and disableUnsecureEndpoint()?
I think some Javadoc could be helpful here, to explain that localAddress/dtlsConfig may be ignored if the endpoints are disabled.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -236,6 +237,17 @@ public static void createAndStartClient(String endpoint, String localAddress, in
         List<LwM2mObjectEnabler> enablers = initializer.create(SECURITY, SERVER, DEVICE, LOCATION,
                 OBJECT_ID_TEMPERATURE_SENSOR);
 
+        // Create CoAP Config
+        NetworkConfig coapConfig;
```
msangoi: I think you should pass this coapConfig to the builder :)

sbernard31:
Integrated in master (commit 0a824be1d3bc8cf512e5acf3597a57f6bb505824,  c728b0f69925bf40eaaf9c3c1280d586a10e55f5, e3479f84dfc38f0b9e15e477863b15b0a70f5c34, 1eb11893abf7f61880d45d38d875f1bbffedbd54)

Issue #372, opened 2017-08-29
Allow a LeshanServer object to be created with an existing DTLSConnector


davidgraeff:
I'm in the situation that I have a ready configured DTLSConnector already available. The LeshanServer object does not allow to use this to create a secure endpoint. Is this on purpose to hide the implementation?

Why is the CoapServer exposed then via `getCoapServer`?

## Proposal

It would be very helpful if the `createUDPConnector(address, config)` would be a protected method and an analog proteced `createDTLSConnector(address, config)` is created and called by the constructor like in:

```
public LeshanServer(...) {
   ...
   if (securityStore != null) {
      DtlsConnector dtlsConnector = createDTLSConnector(localSecureAddress, networkConfig);
      ...
      secureEndpoint = new CoapEndpoint(dtlsConnector, networkConfig, observationService.getObservationStore(), sExchangeStore);
      ...
   }
   ...
}
```

Cheers, David

sbernard31:
> Why is the CoapServer exposed then via getCoapServer?

This is just to let some liberty to end user.

> The LeshanServer object does not allow to use this to create a secure endpoint. Is this on purpose to hide the implementation?

We don't exposed it until now because there was no needs expressed and no easy way to do that.

About your proposal,
This is a possibility but in that case users can not use the builder anymore.
Another solution could be to set a kind of factory

```java
builder.setEndpointFactory( new EndpointFactory() {
        public CoapEndpoint createUnSecureEndpoint
                 (InetSocketAddress address, NetworkConfig coapConfig, ObservationStore store) {
        }
        public CoapEndpoint createSecureEndpoint
                 (DtlsConnectorConfig dtlsConfig , NetworkConfig coapConfig, ObservationStore store) {
        }
}
```

WDYT ?

davidgraeff:
Yup that should do it. Are u doing it or should I create a pr?

sbernard31:
I will do it. I have several changes to do on the builder.

sbernard31:
I create a PR about that #377. Feedbacks is welcomed ;)

davidgraeff:
The proposed PR indeed solves this issue.

davidgraeff:
Thx for integrating this feature :)

PR #377, opened 2017-09-04
Add a way to use custom coap endpoint for LeshanServer

sbernard31:
It should solve #372.

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
@@ -22,10 +22,13 @@ import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.Arrays;
 
+import org.eclipse.californium.core.network.CoapEndpoint;
 import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.core.network.config.NetworkConfig.Keys;
+import org.eclipse.californium.scandium.DTLSConnector;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
 import org.eclipse.leshan.LwM2m;
+import org.eclipse.leshan.core.californium.EndpointFactory;
 import org.eclipse.leshan.core.node.LwM2mNode;
 import org.eclipse.leshan.core.node.codec.DefaultLwM2mNodeDecoder;
 import org.eclipse.leshan.core.node.codec.DefaultLwM2mNodeEncoder;
@@ -75,6 +78,8 @@ public class LeshanServerBuilder {
 
     private NetworkConfig coapConfig;
     private DtlsConnectorConfig.Builder dtlsConfigBuilder;
+
+    private EndpointFactory endpointFactory;
 
     /**
      * <p>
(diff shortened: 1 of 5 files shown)


davidgraeff:
Lgtm. Just some question: Is it actually necessary to provide all SSL configuration parameters to the builder, if the endpoint factory will return an own dtlsConnector? Is there a specific order for the builder methods now?

sbernard31:
There is no order. All is assembled in right order on `build()` method.
All the coap/dtls configuration done by the builder should be used when you create your endpoint in the factory. (I mean all parameters of this 2 methods)
```java
// For example 
CoapEndpoint createUnsecuredEndpoint(InetSocketAddress address, NetworkConfig coapConfig, ObservationStore store){
       return new CoapEndpoint(localAddress, coapConfig, store); 
}
CoapEndpoint createSecuredEndpoint(DtlsConnectorConfig dtlsConfig, NetworkConfig coapConfig,
            ObservationStore store) {
       return new CoapEndpoint(new DTLSConnector(dtlsConfig), coapConfig, store, null);
}
```
You're free to not using it but you must know what you are doing. E.g. create a "DTLS endpoint" without using our PskStore may break PSK authentication.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
```diff
@@ -157,7 +156,7 @@ protected Resource createRoot() {
                 return new RootResource();
             }
         };
-        nonSecureEndpoint = new CoapEndpoint(localAddress, coapConfig, this.observationService.getObservationStore());
+        nonSecureEndpoint = unsecuredEndpoint;
```
msangoi: you can rename it to unsecureEndpoint.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanBootstrapServer.java
```diff
@@ -47,28 +45,28 @@
...
         Validate.notNull(coapConfig, "coapConfig must not be null");
-        Validate.notNull(dtlsConfig, "dtlsConfig must not be null");
 
         this.bsStore = bsStore;
         this.bsSecurityStore = bsSecurityStore;
 
         // init CoAP server
         coapServer = new CoapServer(coapConfig);
-        nonSecureEndpoint = new CoapEndpoint(localAddress, coapConfig);
+        nonSecureEndpoint = unsecuredEndpoint;
```
msangoi: unsecuredEndpoint

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
```diff
@@ -269,6 +274,15 @@ public LeshanServerBuilder setDtlsConfig(DtlsConnectorConfig.Builder config) {
     }
 
     /**
+     * Used to create custom CoAP endpoint, this is only for advanced users. <br>
+     * DTLSConnector is expected for secured endpoint.
```
msangoi: We should explain here that localAddress/dtlsConfig are ignored if an endpoint factory is provided.

msangoi:
👍 LGTM

sbernard31:
Integrated in master (95566f295b516780ad8e71cecaa69e55db7a1909, 1f630498acdec9655fb2724436198103fe666dc0)

Same thing should be done for client but I need access to DTLSConnector and so Californium 2.0.0M5 which should be available soon.

davidgraeff:
Do you have an ETA for a (pre)release? We would need a version at some point for Eclipse Smarthome.

sbernard31:
We are waiting for Californium 2.0.0M5, once it will be available. We need to integrate some works based which depends of this 2.0.0M5 version in Leshan, then we will release a 1.0.0M4. I hope this will happen soon. (something like 1 week)

sbernard31:
@davidgraeff I totally forgot to tell you that the [1.0.0-M4](https://github.com/eclipse/leshan/releases/tag/leshan-1.0.0-M4) is out. It integrates this feature and some other new cool bugs ;)

davidgraeff:
Thank you :)

PR #381, opened 2017-09-07
Add a way to use custom coap endpoint for LeshanClient

sbernard31:
same idea than #377 for LeshanClient


sbernard31:
Integrated in master.

Issue #58, opened 2016-01-06
java.sql.Time is missing when running in a embedded JVM


msangoi:
I've tried to run to the Leshan client in a custom embedded JVM.
If the JVM was created with the 'compact1' profile (http://docs.oracle.com/javase/8/embedded/develop-apps-platforms/compact-profiles.htm), the java.sql.Time dependency is missing: 

```
Exception in thread "main" java.lang.NoClassDefFoundError: java/sql/Time
    at com.google.gson.Gson.<init>(Gson.java:230)
    at com.google.gson.GsonBuilder.create(GsonBuilder.java:545)
    at org.eclipse.leshan.core.model.ObjectLoader.<clinit>(ObjectLoader.java:46)
    at org.eclipse.leshan.client.resource.ObjectsInitializer.<init>(ObjectsInitializer.java:50)
    at org.eclipse.leshan.client.resource.ObjectsInitializer.<init>(ObjectsInitializer.java:45)
    at org.eclipse.leshan.client.example.LeshanClientExample.<init>(LeshanClientExample.java:82)
    at org.eclipse.leshan.client.example.LeshanClientExample.main(LeshanClientExample.java:68)
Caused by: java.lang.ClassNotFoundException: java.sql.Time
    at java.net.URLClassLoader.findClass(Unknown Source)
    at java.lang.ClassLoader.loadClass(Unknown Source)
    at sun.misc.Launcher$AppClassLoader.loadClass(Unknown Source)
    at java.lang.ClassLoader.loadClass(Unknown Source)
    ... 7 more
```

No problem when running in a 'compact2' JVM (adding JDBC package).
But it would be nice to be able to run the client in a minimal embedded JVM.

sbernard31:
We remove the Gson dependency from leshan #383, so no more dependency to `java.sql.Time`.

PR #383, opened 2017-09-12
Remove Gson dependency from Leshan-core.

sbernard31:
Currently we use **gson** in `leshan-core` and for demos (`server-demo` and `bsserver-demo`)
And we use **minimal-json** for `leshan-cluster`.

We should limit us to only one Json library.

**minimal-json** force to write a little more code but it is faster, smaller, more flexible, and does not have dependency to java.sql.time (see #58)
So I replace **gson** by **minimal-json** in `leshan-core` which is the most important part.

Demos will still use **gson** :'(


leshan-core/src/main/java/org/eclipse/leshan/core/model/ObjectLoader.java
```diff
@@ -92,11 +81,15 @@ public static ObjectModel loadDdfFile(InputStream input, String streamName) {
...
     public static List<ObjectModel> loadJsonStream(InputStream input) {
-        Reader reader = new InputStreamReader(input);
-        ObjectModel[] objectModels = GSON.fromJson(reader, ObjectModel[].class);
-        List<ObjectModel> models = new ArrayList<>();
-        Collections.addAll(models, objectModels);
-        return models;
+        try {
+            Reader reader = new InputStreamReader(input);
+            JsonValue json = Json.parse(reader);
+            List<ObjectModel> models = new ObjectModelSerDes().deserialize(json.asArray());
```
msangoi: You can remove 'models' variable.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
```diff
@@ -189,7 +189,7 @@ public void json_encode_device_object_instance() {
         b.append("{\"n\":\"9\",\"v\":100},");
         b.append("{\"n\":\"10\",\"v\":15},");
         b.append("{\"n\":\"11\",\"v\":0},");
-        b.append("{\"n\":\"13\",\"v\":1367491215},");
+        b.append("{\"n\":\"13\",\"v\":1.367491215E9},");
```
msangoi: Don't we want to keep the value without exponent?

leshan-core/src/main/java/org/eclipse/leshan/core/model/ObjectLoader.java
sbernard31: I will do that.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
sbernard31: I agree this is strange to encode a date with this notation... But I think this respect the specification as in JSON, there is only one format for number called "Number" :p (see https://json.org/)

**Why this behave like this :**
When we encode we use several format :
LwM2mNode > LWM2M JSON  > JSON > String

The resource 13 is a date but we loose this information during conversion.
Date ==(leshan)==> Number ==(leshan)==> Double ==(minimal-json)==> String

Minimal-json use [Double.toString](https://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#toString(double))

We can imagine to use a custom minimal-json JsonWriter (it sounds not so good)
or trying to no convert all Number in Double. (we need to add some instanceof in JsonArrayEntrySerDes.jSerialize)

I don't know if it worth it?

msangoi: Ok. indeed, the scientific notation is valid in Json. Let's keep it that way :)

PR #384, opened 2017-09-14
Need to set basename.

rokwoonkim:


sbernard31:
Thx @rokwoonkim.

About `baseName` : we didn't set it because if absent the value should be the request path.
This was mentioned in previous draft(OMA-TS-LightweightM2M-V1_0-20170126-D) of the specification : 
_"The Name attribute in of this array entry is a URI path relative to the Base Name; namely the Name attribute could simply be the full URI of a requested Resource Instance when Base Name is absent."_

But in the final 1.0.0 and 1.0.1 specification , I can not find this sentence anymore, so it seems you're right and `baseName` should be added.

However for baseTime, I'm not sure at all. Why do you want to set basetime ?

So if you agree let's focus on the baseName for now. Could you provide a PR with the basename modification (without basetime) and fix the JUnit tests too ?

The bad new is I need you to sign a CLA to accept your PRs. see [contribution guide](https://github.com/eclipse/leshan/blob/master/CONTRIBUTING.md). 😕

rokwoonkim:
Dear @sbernard31 

About basetime, it just written in TS docs 6.4.4 and i just set for it. but it was miss understand of current leshan's code state.
i will fix this and pull request again.

sbernard31:
@rokwoonkim, JFYI, you can keep the same PR and just git push --force on it.
So feel free to reopen this one or to recreate a new one as you prefer.

About basetime, the spec say : 
_"Note: According to [SENML] specification, time values (Base Time and Time) are represented in floating point as seconds, **a missing attribute is considered to have a value of zero**."_

And currently Leshan does not calculate a basetime it just use time in ressourceArray.

PR #390, opened 2017-09-26
Add a way to provide custom Request Sender

sbernard31:


sbernard31:
I'm still not sure this PR is really needed, so I close it for now.

If you think this PR make sense, do not hesitate to comment it.

PR #385, opened 2017-09-18
Setting Basename

rkimsb2:
Dear @sbernard31 

Sorry for promiscuous commits. i wasn't understand about ECA.

PR #394, opened 2017-09-27
ResponseCode.SERVICE_UNAVAILABLE added

bolddp:
We're working with the con-tiki stack on some clients, and it can respond with a 503 SERVICE UNAVAILABLE under some circumstances (e.g. when there are too many observes already).

This pull request adds support for this so that an Illegal argument exception isn't thrown.


bolddp:
Well, this was a new experience. I messed up with the signed-off-by feature, I'll create a new pull request later.

sbernard31:
@bolddp  if you want you can reuse the same branch and so the same PR.
 You just need to edit your commit comment, using git rebase interactive or git commit amend, then push force on the same branch.

For the CLA, see https://github.com/eclipse/leshan/blob/master/CONTRIBUTING.md

bolddp:
I'm seeing this as a lesson. It's my first Github pull request ever :) And I'm no GIT phantom either, so rebase is a command that I've never used before. But I will definitely check it out!

sbernard31:
Here some resource which could help https://github.com/eclipse/leshan/pull/385#issuecomment-330220547

PR #396, opened 2017-09-27
Added ResponseCode.SERVICE_UNAVAILABLE

bolddp:
Signed-off-by: Daniel Persson <bighatsmallfeet@gmail.com>

### DIFF
--- a/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/ResponseCodeUtil.java
+++ b/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/ResponseCodeUtil.java
@@ -1,15 +1,15 @@
 /*******************************************************************************
  * Copyright (c) 2017 Sierra Wireless and others.
- * 
+ *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v1.0 which accompany this distribution.
- * 
+ *
  * The Eclipse Public License is available at
  *    http://www.eclipse.org/legal/epl-v10.html
  * and the Eclipse Distribution License is available at
  *    http://www.eclipse.org/org/documents/edl-v10.html.
- * 
+ *
  * Contributors:
  *     Sierra Wireless - initial API and implementation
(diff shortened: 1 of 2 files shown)


sbernard31:
I look at the lwm2m spec and this is not a response code explicitly accepted.

Currently, we are pretty strict on that, but maybe it's not a good idea.

Before to make a decision, I would like to better understand your problem with the current version and what will change with this fix ? (are you using Leshan Client or Server ?)

bolddp:
One of our first implemented real-world Lwm2m clients are using the con-tiki stack, where the Observe pattern still needs some work to be less memory-consuming. You can set a constant that states how many simultaneous observes you accept as a client, and when the server sends a new observe request that will exceed that limit, the con-tiki stack responds with 503 Server unavailable. So the main usage is on the server side, but we're also interested in being able to simulate this behavior in our client simulator.

sbernard31:
I see with your change you should be able to create to simulate this 503 error at client side but at server side you will still have a InvalidResponseException because of [LwM2mResponseBuilder](https://github.com/eclipse/leshan/blob/master/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LwM2mResponseBuilder.java#L235).

Am I right ?

bolddp:
Yes, it looks like I missed that. I haven't yet had a chance to try it out fully, I'm still trying to figure out how I can allow our corporate leshan clone (with internal modifications related to the build environment etc.) to play nicely with changes that I think could be of value to the eclipse/leshan master. And I'm no expert on GIT, as you may have noticed from the earlier pull requests attempts...

But if you think that we should include graceful handling of 503, it looks like we need to update the LwM2mResponseBuilder as well.

sbernard31:
I rethink about that and the way we do in leshan is clearly to strict. I will search for another solution, I would like to handle all unknown code not only 503.
Once I have something "showable", I will shared it with you and you let me know if this fits your need.
Does it sounds good for you ?

bolddp:
Yes, I'd be glad to provide feedback.

sbernard31:
#399 in now integrated in master, so we can close this PR.

PR #399, opened 2017-10-04
Add some flexibility in the way we handle response code

sbernard31:
It is now possible to use response codes which are not explicitly defined in the lwm2m specification.

We are still strict on success response expected, if we don't receive the good one an `InvalidResponseException` is still raised.

But now any error codes (accepted by californium) is accepted.

Generally for interoperability, we aim **lenient on input, strict on output** but it seems that [nowadays this make no more consensus](https://softwareengineering.stackexchange.com/questions/152533/should-a-server-be-lenient-in-what-it-accepts-and-discard-faulty-input-silent).   

So I'm still asking myself if this tolerance at  leshan-server-demo is a good thing. Accepting silently not compliant response from the client is maybe not a good idea.

### DIFF
--- a/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/ResponseCodeUtil.java
+++ b/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/ResponseCodeUtil.java
@@ -28,11 +28,7 @@ public class ResponseCodeUtil {
 
     public static ResponseCode toLwM2mResponseCode(
             org.eclipse.californium.core.coap.CoAP.ResponseCode coapResponseCode) {
-        ResponseCode lwm2mResponseCode = ResponseCode.fromCode(toLwM2mCode(coapResponseCode));
-        if (lwm2mResponseCode == null)
-            throw new IllegalArgumentException("Invalid CoAP code for LWM2M response: " + coapResponseCode);
-
-        return lwm2mResponseCode;
+        return ResponseCode.fromCode(toLwM2mCode(coapResponseCode));
     }
 
     public static int toLwM2mCode(int coapCode) {
@@ -42,11 +38,7 @@ public class ResponseCodeUtil {
     }
 
     public static ResponseCode toLwM2mResponseCode(int coapCode) {
-        ResponseCode lwm2mResponseCode = ResponseCode.fromCode(toLwM2mCode(coapCode));
-        if (lwm2mResponseCode == null)
-            throw new IllegalArgumentException("Invalid CoAP code for LWM2M response: " + coapCode);
-
-        return lwm2mResponseCode;
+        return ResponseCode.fromCode(toLwM2mCode(coapCode));
     }
 
     public static int toCoapCode(int lwm2mCode) {
--- a/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/ResponseCodeUtil.java
+++ b/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/ResponseCodeUtil.java
@@ -26,6 +26,15 @@ public class ResponseCodeUtil {
         return coapResponseCode.codeClass * 100 + coapResponseCode.codeDetail;
     }
 
+    public static ResponseCode toLwM2mResponseCode(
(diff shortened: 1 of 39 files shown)


sbernard31:
@bolddp if you have time you could have a look at this. This is a more generic way to handle your needs (see #396)

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/ResponseCodeUtil.java
```diff
@@ -16,69 +16,47 @@
...
-        Validate.notNull(code);
+    public static ResponseCode toLwM2mResponseCode(int coapCode) {
+        return ResponseCode.fromCode(toLwM2mCode(coapCode));
+    }
+
+    public static int toCoapCode(int lwm2mCode) {
+        int codeClass = lwm2mCode / 100;
+        int codeDetail = lwm2mCode % 100;
+        if (codeClass > 7 || codeDetail > 31)
+            throw new IllegalArgumentException("Could not be translate in a valid CoAP code");
```
bolddp: How about "Could not be translated into a valid COAP code"?

leshan-core/src/main/java/org/eclipse/leshan/ResponseCode.java
```diff
@@ -15,48 +15,156 @@
...
-        case METHOD_NOT_ALLOWED:
-        case FORBIDDEN:
-        case NOT_FOUND:
-        case INTERNAL_SERVER_ERROR:
-        case UNSUPPORTED_CONTENT_FORMAT:
-        case NOT_ACCEPTABLE:
+        return isClientError() || isServerError();
+    }
+
+    public boolean isSucces() {
```
bolddp: isSuccess

bolddp:
If I understand it correctly, this change means that any error code that CoAP accepts is turned into an ResponseCode with code=(int) coapResponseCode and name="UNKNOWN", instead of throwing an exception? This seems like a good compromise.

Regarding the tolerance of the Leshan server, I'd like to take a more pragmatic stance. At least one of the Lwm2m client stacks out there (con-tiki) uses a code (503) that is not explicitly stated in the Lwm2m spec.  And I think they are using it in a very reasonable way, signalling that the observe functionality, that was working just fine on the last request, now is unavailable due to resource limitations.

If I would like to raise this issue with the designers of the Lwm2m spec, where is the correct place to contact them?

sbernard31:
Thx for your feedback :), I will fix that.

>If I understand it correctly, this change means that any error code that CoAP accepts is turned into an ResponseCode with code=(int) coapResponseCode and name="UNKNOWN", instead of throwing an exception?

Correct.

>it in a very reasonable way, signalling that the observe functionality, that was working just fine on the last request, now is unavailable due to resource limitations.

That makes sense. Another way which is explicitly explained in the [CoAP observe RFC 7641](https://tools.ietf.org/html/rfc7641#section-4.1), is to answer without observe option. 

_"A server that is unable or unwilling to add a new entry to the list of observers of a resource MAY silently ignore the registration request and process the GET request as usual.  The resulting response MUST NOT include an Observe Option, the absence of which signals to the client that it will not be notified of changes to the resource and, e.g., needs to poll the resource for its state instead."_ 

This is compliant with the LWM2M specification too.

>If I would like to raise this issue with the designers of the Lwm2m spec, where is the correct place to contact them?

You can try : https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues

sbernard31:
Ok I fix it + add some way to show not LWM2M compliant code is used at client side in leshan-server-demo.

(I didn't want to be responsible to the next interoperability mess like html in the 2000s 😅)

danielhqv:
Hi! Do you consider this finished, so it will be merged in the next release? I'm moving my changes to our fork, but instead of introducing SERVICE_UNAVAILABLE like I initially planned, I would like to merge this into our fork and use instead.

sbernard31:
For me this is finished. I'm waiting for @msangoi feedbacks before to merge in master.
It should be available for the next release.

danielhqv:
OK, great! I've merged it into my fork, will be able to test it tomorrow, hopefully.

PR #400, opened 2017-10-04
Added version and urn to ObjectModel

bolddp:
We are in need of versioning and URN info on object models, since we're building a client and server that needs to be able to handle multiple Lwm2m object versions simultaneously. This pull request just displays what we need in the ObjectModel class, this will of course have effects on other code as well, e.g. including version attribute in Link objects when building a LeshanClient etc.

I still haven't come very far in this. What are the implications? Is somebody else working on it?

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/model/ObjectModel.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/model/ObjectModel.java
@@ -31,23 +31,41 @@ public class ObjectModel {
 
     private static final Logger LOG = LoggerFactory.getLogger(ObjectModel.class);
 
+	private static final String DEFAULT_VERSION = "1.0";
+
     public final int id;
     public final String name;
+    public final String version;
+    public final String urn;
     public final String description;
     public final boolean multiple;
     public final boolean mandatory;
 
     public final Map<Integer, ResourceModel> resources; // resources by ID
 
+    @Deprecated
     public ObjectModel(int id, String name, String description, boolean multiple, boolean mandatory,
             ResourceModel... resources) {
         this(id, name, description, multiple, mandatory, Arrays.asList(resources));
     }
 
+    @Deprecated
     public ObjectModel(int id, String name, String description, boolean multiple, boolean mandatory,
+            Collection<ResourceModel> resources) {
+    	this(id, name, DEFAULT_VERSION, null, description, multiple, mandatory, resources);
+    }
+
+    public ObjectModel(int id, String name, String version, String urn, String description, boolean multiple, boolean mandatory,
+            ResourceModel... resources) {
+    	this(id, name, version, urn, description, multiple, mandatory, Arrays.asList(resources));
+    }
+    
+    public ObjectModel(int id, String name, String version, String urn, String description, boolean multiple, boolean mandatory,
             Collection<ResourceModel> resources) {
         this.id = id;
         this.name = name;
+        this.version = version;
+        this.urn = urn;
         this.description = description;
         this.multiple = multiple;
         this.mandatory = mandatory;


sbernard31:
My remarks : 
URN seems to hold version and ID. So one can be calculated from the other and vice-versa.
So I'm not sure we need to store URN attribute and so we don't need to pass it as argument.
We can eventually add a new constructor which take urn instead of version+id.

>  Is somebody else working on it?

As far as I know, nobody are working on that.

> What are the implications? 

LwM2mModel stores objects in a map (objectid -> objectmodel), if we want to support different version of the same object we need to change it with something like (urn->objectmodel)
And complete the API of LwM2mModel to request by version.

bolddp:
Maybe I've misinterpreted something, but I can't see why the ObjectModel class shouldn't reflect the XML Schema of the OMA Object description file as closely as possible. Why omit data that is available, it just might as well be parsed? And version and URN is crucial for our implementation: If the version info on the object level is missing AND it's an OMA object, you should rely on the reported Lwm2m version. How do you determine if it's an OMA object if not by the URN prefix? If the version info on a non-OMA object is missing, you can assume version 1.0.

Anyway, the project I'm involved in where leshan is used, is aiming for starting end-to-end tests in a few weeks, so I've spent the last few days getting multiple object versioning and handling work in our Lwm2m server from end-to-end, e.g. client simulator reading JSON input file with support for different object versions, client performs a correct registration with "ver" attribute on links, and server picks the correct object models based on the registration info. Server then outputs the parsed client objects correctly based on object versions.

This is up and running in our cloned Leshan project as of today (quite a lot of testing left, though). Unfortunately this project is on a corporate network, so I'm not sure I'll find a way to merge the changes directly into my Github leshan fork. And that is of course only meaningful to do if our way of adding this new functionality is considered useful by the community.

sbernard31:
@bolddp, I'm not against adding  support of object version and URN.

I was just saying that maybe we don't need a new attribute and adding a dynamic getter is enough.

The spec say : 
_"With Object versioning, the Object Identifier is not changed but a new URN identifying that particular version of the Object specification MUST be registered according to the following format **“urn:oma:lwm2m:{oma, ext, x}:ObjectID[:{version}]”** where ‘version’ is the Object Version as defined in the following section"_

As I understand URN get be guessed from version+ObjectID or version+objectId can be guessed from URN. ([oma/ext/x can be guessed from ObjectID range](http://www.openmobilealliance.org/wp/OMNA/LwM2M/LwM2MRegistry.html))

Does I missed something, I must confess that this is a part of the spec I don't explore so much ?

bolddp:
I just don't see the point in taking any shortcuts when it comes to letting the ObjectModel class represent the OMA Specification file.

If the input is inconsistent, e.g. the version saying "1.1" and the URN not containing any version info at the end, what should you do then? I'd trust the version attribute, I think, and log a warning about it.

Another problem (not directly related to this but to versioning) is that the version of an OMA object can never be computed unless the object has an explicit version set or you have access to the Lwm2m version. And since the ObjectModel object is passed around on its own, e.g. in the BaseObjectEnabler, you need an explicit version. We've solved this by always setting the ObjectModel.version (which I've added in our fork). This is an argument for implementing ObjectModel.lwm2mVersion as well, which is also a field that is available in the OMA Specification file and used for some objects in the OMA registry.

Anyway, I'm investegating a way for the company I'm currently working for (Husqvarna AB) to make a fork of leshan to their organization's Github repository, which we would use internally. Then it would be much easier to cherry-pick commits for any pull requests to be able to contribute back.

sbernard31:
> I just don't see the point in taking any shortcuts when it comes to letting the ObjectModel class represent the OMA Specification file.

In my mind an ObjectModel represent a valid lwm2m object model, the OMA specification file is just a way to serialize it.

> If the input is inconsistent, e.g. the version saying "1.1" and the URN not containing any version info at the end, what should you do then? I'd trust the version attribute, I think, and log a warning about it.

It sounds reasonable. I think this is up to the  [de-serializer](https://github.com/eclipse/leshan/blob/master/leshan-core/src/main/java/org/eclipse/leshan/core/model/DDFFileParser.java) to do that, what do you think ? 

> Another problem (not directly related to this but to versioning) is that the version of an OMA object can never be computed unless the object has an explicit version set or you have access to the Lwm2m version..

The v1.0 spec say _"the URN where the Object Version is absent: “urn:oma:lwm2m:{oma, ext, x}:ObjectID“ (and supposed to be 1.0))."_ 
So no explicit object version means 1.0, so no need to lwm2mVersion.
By the way I look at the last 1.1 draft and it's still true in this version.

> Anyway, I'm investegating a way for the company I'm currently working for (Husqvarna AB) to make a fork of leshan to their organization's Github repository, which we would use internally. Then it would be much easier to cherry-pick commits for any pull requests to be able to contribute back.

I'm glad to see you're motivate to contribute to Leshan :)

bolddp:
> The v1.0 spec say "the URN where the Object Version is absent: “urn:oma:lwm2m:{oma, ext, x}:ObjectID“ (and supposed to be 1.0))."
So no explicit object version means 1.0, so no need to lwm2mVersion.
By the way I look at the last 1.1 draft and it's still true in this version.

But if a registration states that its Lwm2m version is 2.0, then all OMA Object registration links that are missing version info is implicitly v2.0 as well. So there can be ObjectModel objects, both missing version info, that are in fact representing different versions of the same object. This is stated in section 6.2.3 of the spec.

sbernard31:
I don't see it like that.
e.g. `lwm2m=“2.0”,ep=nodename,</1/0>` and `</1/0>;ver= “2.0”` is the same object model. It's just 2 ways to serialize/deserialize it to/from `link-format`.

I just get the point of the lwm2mVersion :p. It means the minimal version of the enabler needed to support this object, not really link to the default value of the object version. As leshan only support v1.0 it seems reasonable to let this aside for now. We can eventually use it to not load a DDF file we can not support and log a warning/error ?

bolddp:
Let this aside for now? It's already been implemented once by me, since we need it to even be able to deploy our server in a few weeks. We have hardware that are going to connect and use different versions of our spec from the get-go. (OMA objects are at v1.0 but several of our custom objects are already at 2.0)

I just created this pull request to indicate the work that has been done once already (leaving out all the other file edits that this change creates), and to gauge the interest in getting it integrated into the official repo.

And if a pull request with one single file with these spec-compliant additions (no breaking changes, AFAIK) causes this much discussion, I'm already starting to question if it's worth it, to be honest.

sbernard31:
I'm sorry, but I do not understand your reaction... I don't understand what you expect to me ? Just merging your PRs without any remarks?

I think adding support of version object is valuable.
I write code hoping it can be useful. So I really want to find a way to satisfy users.

In this particular case, I just made a little remark : replacing URN attribute by a getter. Not an extravagant remarks : just avoid information duplication. I explained widely why I find this better. Maybe I'm wrong but it's up to you to convince me. FMPOV having an inconsistent model in the system could be error prone, and this is more flexible to let the DDF File loader handle inconsistent file. Again I can change my mind, maybe I missed something...

It seems you find this "process" too time consuming. Sorry about that, but I do not know how to do otherwise. If you chose to finally not contribute this because of lack of time, I understand, but I'm not sure handling an internal fork will be more efficient at long term. 
I hope all of this will not discourage you to contribute (maybe in another way less time consuming).

Good luck for your server deployment in the next weeks ;)

sbernard31:
I read this thread again and maybe we don't understand the spec in the same way :

> If the version info on the object level is missing AND it's an OMA object, you should rely on the reported Lwm2m version.

For me it's clear no version in URN means 1.0 (see https://github.com/eclipse/leshan/pull/400#issuecomment-334702834)

>  How do you determine if it's an OMA object if not by the URN prefix?  

Using object id range.  oma is 0-1023, ext is 2048-10240  and x is 10241-42768

bolddp:
OK, I didn't think about using the object id range for OMA object identification, that is also a possible solution.

But regarding the object versioning, me and the guy who writes our specifications on top of Lwm2m has interpreted the table in section 6.2.3 as I've previously stated: for OMA objects, if the object version info is missing, you cannot assume that it is version 1.0. Instead you need to use the Lwm2m version that was reported during registration.

And I'm not expecting any accepted pull requests without remarks,  but a lengthy discussion about whether to parse the URN for version info or read it from an existing property in the official definition files is not the correct thing to focus on, in my opinion. Should we try to be more clever than OMA in this case, but follow the spec meticulously in other cases? Why not go with the way I suggest for this minor detail and then focus on the real difficulties, like which links should have the version attribute attached and many more?

sbernard31:
> if the object version info is missing, you cannot assume that it is version 1.0, Instead you need to use the Lwm2m version that was reported during registration.

ok I understand the same thing. This has nothing to do with the URN, the default version 1.0 is about missing lwm2mVersion in URN. I suspect that you mixed up the 2 problems.

> Should we try to be more clever than OMA in this case, but follow the spec meticulously in other cases? 

What do you mean ? What is not clever ?

> Why not go with the way I suggest for this minor detail and then focus on the real difficulties

Doing severals PR helps to focus on details and generally it's the good way to do. 
If you want general/global feedback, describe your whole solution or contribute more code to this PR.
If you want to talk about "real difficulty", please expose it.

bolddp:
Closing this to create a new pull request from my organization's fork instead.

PR #402, opened 2017-10-09
Minor enhancement on RedisRegistrationStore

sbernard31:
Fix some potential race condition issue an add kind of clemency for expiration.

### DIFF
--- a/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
+++ b/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
@@ -292,37 +292,39 @@ public class RedisRegistrationStore implements CaliforniumRegistrationStore, Sta
     @Override
     public Deregistration removeRegistration(String registrationId) {
         try (Jedis j = pool.getResource()) {
-
-            byte[] regKey = toRegIdKey(registrationId);
-
-            // fetch the client ep by registration ID index
-            byte[] ep = j.get(regKey);
-            if (ep == null) {
-                return null;
-            }
-
-            byte[] data = j.get(toEndpointKey(ep));
-            if (data == null) {
-                return null;
-            }
-
-            Registration r = deserializeReg(data);
-            deleteRegistration(j, r);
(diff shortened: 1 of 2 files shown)


leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
```diff
@@ -62,6 +62,11 @@
  */
 public class RedisRegistrationStore implements CaliforniumRegistrationStore, Startable, Stoppable {
 
+    /** Default time in seconds between 2 cleaning tasks (used to remove expired registration). */
+    public static final long DEFAULT_CLEAN_PERIOD = 60;
+    /** Defaut Extra time for registration lifetime in seconds */
+    public static final long DEFAULT_CLEMENCY = 0;
```
msangoi: what do you think of renaming "clemency" to "grace period"? again a naming question ;)

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
```diff
@@ -262,7 +262,11 @@ public Date getLastUpdate() {
     }
 
     public boolean isAlive() {
-        return lastUpdate.getTime() + lifeTimeInSec * 1000 > System.currentTimeMillis();
+        return isAlive(0);
+    }
+
+    public boolean isAlive(long clemencyInSec) {
```
msangoi: I think this method needs a proper javadoc, in order to understand the meaning of this grace/clemency period.
And the default `isAlive()` as well.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
sbernard31: I will change it

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
sbernard31: I can try

PR #405, opened 2017-10-20
Enabled ability to force registration update

danielhqv:
For a LwM2m Server instance to be able to react on a Registration Update execute request, there needs to be a mechanism that can reschedule the update registration to occur immediately.

Signed-off-by: Daniel Persson <daniel.p.persson@husqvarnagroup.com>

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
@@ -214,4 +214,8 @@ public class LeshanClient implements LwM2mClient {
     public String getRegistrationId() {
         return engine.getRegistrationId();
     }
+    
+    public RegistrationEngine getRegistrationEngine() {
+        return engine;
+    }
 }
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
@@ -176,6 +176,10 @@ public class LeshanClient implements LwM2mClient {
         LOG.info("Leshan client destroyed.");
     }
 
+    public void triggerRegistrationUpdate() {
+        engine.triggerRegistrationUpdate();
+    }
+    
     @Override
     public Collection<LwM2mObjectEnabler> getObjectEnablers() {
         return Collections.unmodifiableCollection(new ArrayList<>(objectEnablers.values()));
@@ -213,9 +217,5 @@ public class LeshanClient implements LwM2mClient {
      */
     public String getRegistrationId() {
         return engine.getRegistrationId();
-    }
-    
-    public RegistrationEngine getRegistrationEngine() {
-        return engine;
     }
 }
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/RegistrationEngine.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/RegistrationEngine.java
@@ -383,10 +383,14 @@ public class RegistrationEngine {
         }
     }
 
-    public void forceUpdate() {
-        cancelUpdateTask(false);
-        LOG.info("Forcing update of registration...");
-        updateFuture = schedExecutor.schedule(new UpdateRegistrationTask(), 100, TimeUnit.MILLISECONDS);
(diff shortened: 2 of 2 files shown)


leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
```diff
@@ -214,4 +214,8 @@ public void removeObserver(LwM2mClientObserver observer) {
     public String getRegistrationId() {
         return engine.getRegistrationId();
     }
+    
+    public RegistrationEngine getRegistrationEngine() {
+        return engine;
```
sbernard31: Do you really need to access to RegistrationEngine ? if not, I will go for an update() or forceUpdate() method.
Does it make sense to you ?

Probably off-topic:  currently we support only one server but when we want support more I suppose we need to add a registrationID parameter to this method.

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/RegistrationEngine.java
```diff
@@ -383,6 +383,12 @@ public void destroy(boolean deregister) {
         }
     }
 
+    public void forceUpdate() {
+        cancelUpdateTask(false);
+        LOG.info("Forcing update of registration...");
+        updateFuture = schedExecutor.schedule(new UpdateRegistrationTask(), 100, TimeUnit.MILLISECONDS);
```
sbernard31: why using a delay ?

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/RegistrationEngine.java
```diff
@@ -383,6 +383,12 @@ public void destroy(boolean deregister) {
         }
     }
 
+    public void forceUpdate() {
```
sbernard31: you should add `if (started)`

sbernard31: Should probably be a synchronized method.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
bolddp: You mean update() or forceUpdate() on LeshanClient directly? That would be OK too. Regarding the method naming, maybe the name triggerUpdateRegistration() is clearer. Just "update()" is not as clear, and "forceUpdate()" is maybe too harsh.

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/RegistrationEngine.java
bolddp: The reasoning is that this is triggered by a server request, and a delay would decrease the risk of interfering with the response to that request. I'm no UDP expert, but a short delay seems less error prone. Maybe it should even be longer?

sbernard31: It seems pretty binded to your specific use case. I will go for a scheduler.submit() and so you could add the delay you want on your side (this will be more flexible for you)

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/RegistrationEngine.java
```diff
@@ -383,6 +383,12 @@ public void destroy(boolean deregister) {
...
+        cancelUpdateTask(false);
```
sbernard31: I think we need to interrupt the task. (false => true)
What do you think ?

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/RegistrationEngine.java
bolddp: Please clarify your last comment, I'm not sure what you mean.

sbernard31: Replace schedExecutor.schedule() by schedExecutor.submit()

bolddp: OK

sbernard31:
Integrated in master (commit cf933b630b02192da8b97d9aba72ad2fb1baae15)

PR #408, opened 2017-10-23
Object versioning

danielhqv:
Adds object version support for ObjectModel and LinkFormatHelper etc. No handling of URN in this PR, mainly due to time limitations in our project. Since we don't register our custom objects with OMA, any URN that we use would not be guaranteed to be unique anyway.


danielhqv:
Why is this validation still failing? I signed the last commit (by resetting and commit --amend), and I have signed a ECA with my email address...

danielhqv:
I just realized that object versioning is not reflected in the server-demo... I will add that to this pull request as well

danielhqv:
Closing this since the last commit is correctly signed, but still the checks are failing. Will add object version handling to server-demo on this branch and then open a new pull request

sbernard31:
Don't worry so much about ECA validation. I can check it manually, if it's ok I don't care about this automatic validation.

danielhqv:
Oh, I actually thought that it would stop you from merging the commit. But that's good to know. Give me a few more tries, and I just might get it right every time :)

PR #414, opened 2017-10-27
Add snapcraft build files

slimjim777:
Adding the build files to create a snap that can be installed on multiple platforms (see [snapcraft.io](http://snapcraft.io)). This will make Leshan simpler for users to try out. Once the snap is built and installed it provides three commands:

```
leshan.server
leshan.bootstrap
leshan.client
```
Which are the demo server, bootstrap server and client.

The snap can be built and published on the Ubuntu Snap Store, which will make it available to install on any system that has snapd installed via:
```
snap install leshan
```

Signed-off-by: James Jesudason <james.jesudason@gmail.com>

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,32 @@
+name: leshan
+version: '1.0.0'
+summary: Eclipse Leshan is an OMA Lightweight M2M server and client Java implementation
+description: |
+  Leshan provides libraries which help people to develop their own Lightweight M2M server and client.
+  The project also provides a client, a server and a bootstrap server demonstration as an example of 
+  the Leshan API and for testing purpose.
+
+grade: stable
+confinement: strict
+
+apps:
+  server:
+    command: bin/server.sh
+    plugs: [network-bind, network]
+  bootstrap:
+    command: bin/bootstrap.sh
+    plugs: [network-bind, network]
+  client:
+    command: bin/client.sh
+    plugs: [network]
+
+parts:
+  leshan:
+    plugin: maven
+    maven-options: [-Pall]
+    maven-targets: [leshan-server-demo, leshan-bsserver-demo, leshan-client-demo]
+  bin:
+    source: snap/bin
+    plugin: dump
+    organize:
+      "*": bin/


sbernard31:
Hi,
  Thx for this proposal.
  How this is simpler than ?

```
wget https://hudson.eclipse.org/leshan/job/leshan/lastSuccessfulBuild/artifact/leshan-server-demo.jar
java -jar ./leshan-server-demo.jar
```

slimjim777:
Good question! It provides a universal approach to installing apps on a number of platforms. Also, by publishing the snap to the Ubuntu Store it then increases the exposure of Leshan to any system that has snapd installed. This includes Ubuntu, Debian, Yocto, Fedora....

The snap package includes everything that the apps need to run (it includes the client, server and bootstrap server), with all the dependencies and paths already setup. Your example assumes that the java runtime is installed and configured and is a compatible version. The snap is a single, complete, sandboxed application that can be installed and used simply.

Also, maybe not relevant for Leshan, but the snapcraft build system builds for multiple architectures e.g. amd64 and arm. So that makes it simpler to install and try out on a device.

sbernard31:
I'm still counting pros and cons.

About exposure, I don't know I'm not sure it's a good idea to expose demo without explaining they are demo. Leshan is not an application, it's a library.

For us (I mean as committer) :
1) We need to give access to snapcraft.io to our github account 
   - not sure to know why it needs access to "Organizations and teams" (read only)
   - for "Repository webhooks and services" (read/write), I suppose this is for be able to rebuild on changes, but I'm not fan to give write access...
 2) We need to maintain the snap file (it seems not so hard)

 For users, 
 1) They need to install `snapd` then install the leshan snap which contains a full JVM (if I well understand) ... not sure this is better to just install a JVM (which is often already installed)
  2) About the compatible version of java, we are java7 compliant so I'm not worried about that.
  3) About multiple arch,  java seems to do the job.

I really think that snap and flatpack are great improvement for linux world, but in our case sounds not so useful.

slimjim777:
Whilst I understand that, as a project team, you see Leshan as a framework and not an application, but I think that you may be victims of you own success :) Whilst the server and bootstrap demos are only intended as a reference, I think they have become really good out-of-the-box implementations of a LwM2M server. I think that the demos are actually quite attractive and useful to IoT developers who are implementing LwM2M at a client level.

Just look at what the server and bootstrap demos provide:

- a working LwM2M server and bootstrap implementation
- web interface to interact with the clients (limited, but functional)
- an HTTP API to allow a management adapter to interact with the client
- ability to test out DTLS and bootstrapped credentials

An IoT developer that is looking to add LwM2M to a device may not be proficient in Java and it's eco-system. So having the ability to quickly install a Leshan server is quite attractive. Many of the developers will be developing on Ubuntu anyway, so snapd is already installed. So being able to install a Leshan snap is quite beneficial to them - they can spin up a Leshan server quickly and focus on the functionality of their LwM2M client application.

I understand the concerns about the github account for snapcraft.io. The main benefit there is that it will build for multiple architectures. It is possible to handle builds manually using the ```snapcraft``` command and ```snapcraft push``` to upload to the store.

Where Leshan as a framework is more interesting is for other projects such as Kura where they need a LwM2M protocol adapter. An interesting personal observation from EclipseCon was that it was on their roadmap, but it did not seem high on their priority list.

sophokles73:
Here are my two cents: I agree with @sbernard31 in that leshan is about providing libraries for building LWM2M based applications and that people who are interested in using it in that way will not have any problem with setting up an appropriate Java environment for it.
That said, I do see @slimjim777 's point in that leshan has become quite attractive as a sort of *reference* server implementation that people use to verify their clients against. For that matter, having a snap for setting up a server in such an easy way seems very worthwhile to me and may indeed foster adoption of leshan even further than what is already happening.
So, FMPOV we should discuss the implications of integrating the GitHub repo with snapcraft.io with the Eclipse webmasters and see where this leads us.

sbernard31:
> We should discuss the implications of integrating the GitHub repo with snapcraft.io with the Eclipse webmasters and see where this leads us.

Let's see what they think about that.

> For that matter, having a snap for setting up a server in such an easy way seems very worthwhile to me 

So it seems the main benefits, is "the easy way to get leshan-demo running", but again the current way is really easy ... just download a jar and launch a java -jar command : 
```
wget https://hudson.eclipse.org/leshan/job/leshan/lastSuccessfulBuild/artifact/leshan-server-demo.jar
java -jar ./leshan-server-demo.jar
```

sbernard31:
No more news about integrating snapcraft.io.
And to be honest, I'm still not really convinced by this idea.

PR #415, opened 2017-10-29
Added acknowledgement handling for long running timeouts

danielhqv:
By making the SyncRequestObserver also use the onAcknowledgement callback when determining timeout, a long running write to Firmware update / Package could be accomplished.

Is this is a viable way of solving the issue with reset of the timeout during block transfers?

Signed-off-by: Daniel Persson <daniel.p.persson@husqvarnagroup.com>

### DIFF
--- a/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/SyncRequestObserver.java
+++ b/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/SyncRequestObserver.java
@@ -34,6 +34,7 @@ public abstract class SyncRequestObserver<T extends LwM2mResponse> extends Abstr
     private CountDownLatch latch = new CountDownLatch(1);
     private AtomicReference<T> ref = new AtomicReference<>(null);
     private AtomicBoolean coapTimeout = new AtomicBoolean(false);
+    private AtomicBoolean acknowledged = new AtomicBoolean(false);
     private AtomicReference<RuntimeException> exception = new AtomicReference<>();
     private Long timeout;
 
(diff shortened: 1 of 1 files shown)


sbernard31:
I don't think this is a good idea, the timeout is about maximum time to wait before for a response.

I mean this could be something like 15 minutes or more if you think you will transfer something very large.

> Is this is a viable way of solving the issue with reset of the timeout during block transfers?

Could you describe your issue exactly ?

danielhqv:
This is based on our need to write several 100 k to the Firmware update / Package resource, an action that can take several minutes to perform. If it fails, and we assume that it will quite often, we want to abort, release resources and be able to retry as quickly as possible.

And for our use case, writing to the Package URI resource is not an option, unfortunately.

sbernard31:
Currently the way we go for Leshan is to hide a lot of CoAP stuff.
The API will just expose : 
- a LWM2M request to send
- a LWM2M response received
- a "LWM2M response timeout"  : the maximum time to wait for this response (this should be added in Leshan for async response too)

All about the CoAP layer is handle transparently but tweak-able with configuration. (block mode, retransmission, acknowledgement ...)

1) So the block mode is transparent, this means that block transfer will be automatically done using the californium properties (PREFERRED_BLOCK_SIZE, MAX_MESSAGE_SIZE, MAX_RESOURCE_BODY_SIZE) 

2) BLOCKWISE_STATUS_LIFETIME (not fully operational in californium version integrated in Leshan for now) which is the maximum time to wait between a block request and a block response (CoAP client side) or between a block response and a block request (CoAP server side)

In your case, you are using block 1 and you are interested by the CoAP client side (LWM2M server sending a write request )
So, you will send a write request with a "big" payload and give a "LWM2M response timeout".

The failure case are : 
1) one of the block request is never acknowledged (max retransmission attempt) => this should raise a timeout at Leshan side.
2) one of the block request is acknowledged but we don't get the response in  BLOCKWISE_STATUS_LIFETIME ms => this should raise a timeout at Leshan side.
3) Any error response received during block transfer => this should raise an error at Leshan side.
4) No response received in a "LWM2M response timeout" => this should raise a timeout.

Do you think this could fit your needs ?

danielhqv:
I'm not sure I follow you completely. Can I already achieve the same behavior by tweaking the configuration?

One problem is also that it's extremely hard to predict how long the request/response timeout will need to be, probably >10 minutes to be on the safe side. It's much easier to predict a reasonable time between blocks, as I see it?

And will the transfer abort earlier than the timeout I set if there is an error in the COAP layer?

I will test this further and see what I find.

sbernard31:
The behavior I described is what it is expected, not how it behave for now.
Some californium features I talked about is not yet integrated Leshan and there is still some bugs.

But as I understand, the way you should go when all of that will be operational will be : 
1) use a LWM2M response timeout to 10/15 minutes.
2) use a BLOCKWISE_STATUS_LIFETIME  to something like 2/3 minutes. (or maybe less)

sbernard31:
More details in #428

sbernard31:
As #428 is integrated in master and as we use californium 2.0.0-M6 (since #440 is integrated). I think we can close this PR.
Do not hesitate to reopen it, If I'm wrong.

Issue #418, opened 2017-11-02
trustedCertificates from LeshanServerBuilder doesn't allow subtypes of Certificate


AlexITC:
### Problem:
The field [LeshanServerBuilder#trustedCertificates](https://github.com/eclipse/leshan/blob/master/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java#L77) doesn't allow to specify subtypes of `Certificate`, I need to cast my types just to comply with value type.

### Solution
Modify the field to type to allow subtypes of Certificate, one way could be to use `List<? extends Certificate> trustedCertificates`.

We could keep backwards compatibility, in case this is viable, I can upload a pull request with the changes.

sbernard31:
it sounds a good idea.

I suppose if we change this for `trustedCertificates`, we also need to change it for `certificateChain`

I look at the underlying DTLS library (scandium) and it seems that only [X509Certificate (and subclasses)](https://github.com/eclipse/californium/blob/2.0.x/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java#L720) was supported. 

Maybe we should change `List<? extends X509Certificate> trustedCertificates.` to be consistent ?

AlexITC:
I didn't change the type of `trustedCertificates` to `X509Certificate` to not introduce breaking API changes.

sbernard31:
API break is not really an issue, if we really think the new one is better.

You previously talk about using `List` instead of `Array`. Is there any reason to finally use Array ?

I don't know what could be the best choice,
 - java keystore returns Arrays,
 - but often you wan to manipulate it, like in [demo](https://github.com/eclipse/leshan/blob/master/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java#L261) and so you go with List.

AlexITC:
The reason was to not break API compatibility, do you think it would be better to use the list?

I would be happy to switch it.

sophokles73:
Is there a reason why you would want to change the trust anchor (the argument for using a `List` instead of an array) once it has been established?
If at all, we should change it to a `Set`, I guess, since we do not need duplicates in it ...

sbernard31:
what do you mean by "once it has been established" ?

sophokles73:
[...]

PR #419, opened 2017-11-04
#418: Allow subtypes of Certificate and X509Certificate on LeshanServerBuilder

AlexITC:
Signed-off-by: Alexis Hernandez <alexis22229@gmail.com>

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
@@ -226,7 +226,7 @@ public class LeshanServerBuilder {
      * For RPK the public key will be extract from the first X509 certificate of the certificate chain. If you only need
      * RPK support, use {@link LeshanServerBuilder#setPublicKey(PublicKey)} instead.
      */
-    public LeshanServerBuilder setCertificateChain(X509Certificate[] certificateChain) {
+    public <T extends X509Certificate> LeshanServerBuilder setCertificateChain(T[] certificateChain) {
         this.certificateChain = certificateChain;
         return this;
     }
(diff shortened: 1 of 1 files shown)


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
```diff
@@ -226,15 +226,15 @@ public LeshanServerBuilder setPrivateKey(PrivateKey privateKey) {
      * For RPK the public key will be extract from the first X509 certificate of the certificate chain. If you only need
      * RPK support, use {@link LeshanServerBuilder#setPublicKey(PublicKey)} instead.
      */
-    public LeshanServerBuilder setCertificateChain(X509Certificate[] certificateChain) {
+    public <T extends X509Certificate> LeshanServerBuilder setCertificateChain(T[] certificateChain) {
```
sophokles73: pardon my ignorance but what is the difference between the two? What kind of certificates can I pass in now which I couldn't pass in before?

sbernard31: I thought that arrays were invariant like generics but [there are not](https://stackoverflow.com/questions/2745265/is-listdog-a-subclass-of-listanimal-why-arent-javas-generics-implicitly-p/46496077#46496077).
So good question @AlexITC  can you provide an example where this modification helps ?

AlexITC: oh, I forget to mention, I'm using Scala, Java arrays are handled as `Array<T>` collection, hence invariant.

sophokles73:
LGTM

sbernard31:
integrated in master (commit e54f0a368d2df10525bb75b9b902779c17917348)

PR #428, opened 2017-11-09
 Timeout is now mandatory or have a default value for sync and async request 

sbernard31:
## Response timeout
This timeout is the maximum time to wait for a response. 
Let's call this the **response timeout**. 
You could choose the timeout for each requests like this :
```java
leshanServer.send(destination, requestToSend, timeoutInMs);
```
The default value is 2 minutes which is a bit longer than the **CoAP timeout**.
**response timeout** is not the only way a request could timed out. Let's see the other-ones.

## CoAP timeout

In CoAP a request could be Confirmable, this means that's the peer receiving the request should acknowledged it. There is a mechanism of retransmission. If request is not acknowledged after all retransmissions. The request will timed-out too. 
See [RFC-7252-§4.2](https://tools.ietf.org/html/rfc7252#section-4.2) for more details.
Let's call this the **CoAP timeout**. 
This value can be set for all requests  via the coapConfig.
```java
NetworkConfig coapConfig = LeshanServerBuilder.createDefaultNetworkConfig();

// here is the default value
coapConfig.setInt(NetworkConfig.Keys.ACK_TIMEOUT, 2000); // in ms
coapConfig.setFloat(NetworkConfig.Keys.ACK_RANDOM_FACTOR, 1.5f);
coapConfig.setFloat(NetworkConfig.Keys.ACK_TIMEOUT_SCALE, 2f);
coapConfig.setInt(NetworkConfig.Keys.MAX_RETRANSMIT, 4);
// with this value, a none-acknowledged confirmable request should timeout between 62s and 93s.
// see MAX_TRANSMIT_WAIT in the RFC-7252

builder.setCoapConfig(coapConfig);
```
This acknowledgement just mean "I receive your request and I will respond soon", so a request can be acknowledged but we never receive a response. In this case the request will timed-out because of **response timeout**.

Note that in CoAP you can send acknowledgement and response in the same packet (called [piggyback](https://tools.ietf.org/html/rfc7252#section-5.2.1))

## Block timeout
Leshan use californium transparent blockwise. This mean that blockwise mode is automatically use when payload is too large. (see [RFC-7959](https://tools.ietf.org/html/rfc7959))
If any of the bloc request is not acknowledged in time the request will timed-out (see **CoAP timeout**)

The maximum time between "a block request and a block response" or between "a block response and a block request" is `BLOCKWISE_STATUS_LIFETIME` (5min by default).
If no request or response is received this expected time, the request will timed-out too.
Let's call this **Block timeout**
This value can be set for all block requests  via the coapConfig.
```java
NetworkConfig coapConfig = LeshanServerBuilder.createDefaultNetworkConfig();

coapConfig.setInt(NetworkConfig.Keys.BLOCKWISE_STATUS_LIFETIME, 5*60*1000); // in ms
builder.setCoapConfig(coapConfig);
```

# Be aware !
When a request timed out in any way all the internal data relative to this request will be released.
It's up to you to define the right value for all this timeout.
If you plan to send or receive large data using block mode the default 2minutes one is probably too short.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/LwM2mServer.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/LwM2mServer.java
@@ -112,6 +112,27 @@ public interface LwM2mServer {
             ResponseCallback<T> responseCallback, ErrorCallback errorCallback) throws CodecException;
 
     /**
+     * Sends a Lightweight M2M request asynchronously.
+     * 
+     * @param destination the remote client
+     * @param request the request to send to the client
+     * @param timeout the request timeout in millisecond
+     * @param responseCallback a callback called when a response is received (successful or error response)
+     * @param errorCallback a callback called when an error or exception occurred when response is received. It can be :
+     *        <ul>
+     *        <li>{@link RequestRejectedException} if the request is rejected by foreign peer.</li>
+     *        <li>{@link RequestCanceledException} if the request is cancelled.</li>
+     *        <li>{@link InvalidResponseException} if the response received is malformed.</li>
+     *        <li>{@link TimeoutException} if the CoAP timeout expires ( see
+     *        http://tools.ietf.org/html/rfc7252#section-4.2 ).</li>
+     *        <li>or any other RuntimeException for unexpected issue.
+     *        </ul>
+     * @throws CodecException if request payload can not be encoded.
+     */
+    <T extends LwM2mResponse> void send(Registration destination, DownlinkRequest<T> request, long timeout,
+            ResponseCallback<T> responseCallback, ErrorCallback errorCallback) throws CodecException;
+
+    /**
      * Get the registration service to access to registered clients. You can use this object for listening client
      * registration lifecycle.
      */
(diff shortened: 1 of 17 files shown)


sbernard31:
**Block timeout** is not fully functional in californium 2.0.0-M5 and so in leshan for now.
We should wait for californium 2.0.0-M6 (this PR https://github.com/eclipse/californium/pull/449 should help)

leshan-server-core/src/main/java/org/eclipse/leshan/server/LwM2mServer.java
```diff
@@ -111,6 +111,27 @@
...
+     *        <ul>
+     *        <li>{@link RequestRejectedException} if the request is rejected by foreign peer.</li>
+     *        <li>{@link RequestCanceledException} if the request is cancelled.</li>
+     *        <li>{@link InvalidResponseException} if the response received is malformed.</li>
+     *        <li>{@link TimeoutException} if the CoAP timeout expires ( see
+     *        http://tools.ietf.org/html/rfc7252#section-4.2 ).</li>
+     *        <li>or any other RuntimeException for unexpected issue.
+     *        </ul>
+     * @throws CodecException if request payload can not be encoded.
+     */
```
msangoi: 👍

PR #382, opened 2017-09-11
Adapt for EndpointContext.

boaks:
Adjust observe test to set destination context.
Intended to provide first experience on using an endpoint context.
(DON'T MERGE NOW :-) ) 

Signed-off-by: Achim Kraus <achim.kraus@bosch-si.com>

PR #440, opened 2017-11-27
Use new EndpointContext from Californium

### LINKED ISSUE #187: Requests initiated by LWM2M Server get sent to wrong LWM2M client with same IP:Port
When a request to a LWM2M client, e.g. read request, gets triggered, Leshan looks up the client with the desired endpoint name from the client registry and sends the request to the IP:Port saved there. When this IP:Port gets used by another client in the meantime, e.g. port re-usage of NAT device, the request gets sent to the wrong client. I think without usage of DTLS this behavior is just fine.

However, the same applies also with DTLS. Example:
-          client1 registers with Leshan using PSK1
-          client2 creates new DTLS session with Leshan/Scandium from same IP:Port but with PSK2 (client does not need to register at LWM2M level)
-          server sends read request to client1, however request arrives at client2 (as client2 has IP:Port of client1) and client2 can even decrypt it

One way to solve this would be to handover recipient principal somehow to scandium and scandium only sends the message when this is the same as the active DTLS session. Or even more strict handover the DTLS sessionId etc. of the last message received from the according client to scandium (like Scandium CorrelationContext but only the other way around).

See also mailing list discussion: http://dev.eclipse.org/mhonarc/lists/leshan-dev/msg00610.html

sbernard31:
This is largely based on what @boaks done in #382.

Main changes use Identity instead of just address/port to send request.
Thx to EndpointContext we can now ensure we send request to the peer we expect.

This should fix  #187.

~(should not be merged before californium release a new version, but can be reviewed ;))~


sbernard31:
I update the PR to use the 2.0.0-M6 version of californium.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/ObservationSerDes.java
```diff
@@ -61,8 +68,15 @@
     public static Observation deserialize(byte[] data) {
         JsonObject v = (JsonObject) Json.parse(new String(data));
 
+        EndpointContext endpointContext = EndpointContextSerDes.deserialize(v.get("peer").asObject());
```
msangoi: `v.get("peer")`: possible NPE for already serialized observations?

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/ObservationSerDes.java
```diff
@@ -75,7 +89,7 @@ public static Observation deserialize(byte[] data) {
         }
 
         // TODO handle security context
```
msangoi: remove TODO?

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/RegistrationSerDes.java
```diff
@@ -86,8 +85,7 @@ public static String sSerialize(Registration r) {
 
     public static Registration deserialize(JsonObject jObj) {
         Registration.Builder b = new Registration.Builder(jObj.getString("regId", null), jObj.getString("ep", null),
-                new InetSocketAddress(jObj.getString("address", null), jObj.getInt("port", 0)).getAddress(),
-                jObj.getInt("port", 0),
+                IdentitySerDes.deserialize(jObj.get("identity").asObject()),
```
msangoi: v.get("identity"): same NPE issue with already serialized observations?
We should handle the transition to build the identity from the "address" and "port" fields.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/RegistrationUpdateSerDes.java
```diff
@@ -84,8 +83,7 @@ public static RegistrationUpdate deserialize(byte[] data) throws UnknownHostExce
 
         // mandatory fields
         String regId = v.getString("regId", null);
-        InetAddress addr = InetAddress.getByName(v.getString("address", null));
-        int port = v.getInt("port", -1);
+        Identity identity = IdentitySerDes.deserialize(v.get("identity").asObject());
```
msangoi: Same transition issue as `RegistrationSerDes`

sbernard31:
For redis store compatibility break, I try to propose a kind of solution with #454

PR #450, opened 2017-12-14
Leshan client refactoring : use one CoAP endpoint by server

sbernard31:
This PR aims to prepare the ground to implement support of RPK, X509 and multi server at client side.

Currently in Californium/Scandium there is no way to use different RPK,  X509 certificate or cipher suite depending of the destination for the same CoapEndpoint/Connector. @boaks could you confirm ?

So I try a way where a CoapEndpoint is created for each server destination.
This PR still not support RPK, 509 and multi server but this change the code in way to make it possible.

@AlexITC, @danielhqv  as you played with LeshanClient. If you have time I would really appreciate your feedback about that. (@boaks also of course)

As I changed a lot of thing there is probably some new bugs :/.
I'm not totally satisfy, I feel there is still redundant concept (like Server/ServerInfo),  maybe we can simplify this part later.

The next step is probably to add RPK. (this is the most straightforward)
Then X509, but I'm a bit confuse by the specification, I'm waiting some clarification https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/257.
Multi-server is another story, I don't know if there is a needs for that ? supported it will change a bit more the API but this should be feasible now.


boaks:
@sbernard31 

Yes, currently the LWM2M device support in californium is limited to one LWM2M server :-).
Therefore I liked your 

https://github.com/eclipse/californium/issues/484

But I'm not sure, how many users use californium for LWM2M devices in a product and not for demo's/testing only.

boaks:
OK, I think, I mixed up the issues:

https://github.com/eclipse/californium/issues/484

is just about the "implementation" of trust.

So we need a issue for providing a credentials interface selecting the credentials via the address as in PskStore.

sbernard31:
> how many users use californium for LWM2M devices in a product and not for demo's/testing only.

I agree, I'm curious to know that. If I well understand @danielhqv use it for a simulator.

> So we need a issue for providing a credentials interface selecting the credentials via the address as in PskStore.

This should be credentials and ciphersuite. @boaks did you think this is quickly/easily feasible ? I don't know what is the best plan ? changing californium or changing leshan client  I do in this PR ?

boaks:
I already started to change californium (I guess, it must be done there) in the afternoon, but some other issues seems to be more important (right now).

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
```diff
@@ -0,0 +1,198 @@
...
+        // TODO we should probably clean cipherSuite depending on which kind of endpoint(psk,rpk,x509) we will create
+        // in case users choose specific cipher
+
+        return newBuilder;
+    }
+
+    @Override
+    public synchronized Collection<Server> createEndpoints(Collection<? extends ServerInfo> serverInfo) {
+        if (serverInfo == null || serverInfo.isEmpty())
+            return null;
```
AlexITC: wouldn't be safer to return an empty collection instead of null?

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
```diff
@@ -0,0 +1,198 @@
...
+
+            if (endpointFactory != null) {
+                currentEndpoint = endpointFactory.createSecuredEndpoint(newBuilder.build(), coapConfig, null);
+            } else {
+                currentEndpoint = new CoapEndpoint(new DTLSConnector(newBuilder.build()), coapConfig, null, null);
+            }
+            server = Identity.psk(serverInfo.getAddress(), serverInfo.pskId);
+        } else {
+            if (endpointFactory != null) {
+                currentEndpoint = endpointFactory.createUnsecuredEndpoint(localAddress, coapConfig, null);
```
AlexITC: what about passing the ObservationStore to both endpoints?

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
```diff
@@ -0,0 +1,198 @@
...
+import org.eclipse.leshan.SecurityMode;
+import org.eclipse.leshan.client.servers.EndpointsManager;
+import org.eclipse.leshan.client.servers.Server;
+import org.eclipse.leshan.client.servers.ServerInfo;
+import org.eclipse.leshan.core.californium.EndpointFactory;
+import org.eclipse.leshan.core.request.Identity;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CaliforniumEndpointsManager implements EndpointsManager {
```
AlexITC: What is the responsibility of this class? What I see is that it will create a single endpoint for several servers, am I right?

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
boaks: The `ObservationStore` is only required for the CoAP-client of such an observation. Though LWM2M TS doesn't define, that a LWM2M-device-client is allowed to establish an observation of a LWM2M-server resource, I think, the `ObservationStore` is not used at the LWM2M-device-client.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
sbernard31: I will do that.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
sbernard31: I agree with @boaks, I think `ObservationStore` is not needed at LWM2M-client side.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
sbernard31: Creating/Starting/Stopping endpoints. This class depends on californium.
The registration engine does not depend on it and so uses the EndpointsManager abstraction.

This class should create 1 endpoint by server, but for now we don't support several dm server.
So it create only 1 endpoint for the bootstrap server in a bootstrapsession.
Once the bootstrap session is finished we create a new endpoint for the DM server (the only one we support for now)

I tried to put the brick to support multi-server (maybe it's a mistake and its too soon to do that)

sbernard31:
> I already started to change californium (I guess, it must be done there) in the afternoon, but some other issues seems to be more important (right now).

@boaks  if you have something to show even if this is not finished, I'm interesting to see that.

boaks:
See first "skeleton":

https://github.com/eclipse/californium/tree/credentials_store

The idea is a interface, which fetches the credentials based on the peer's address. But currently I'm concentrate on

https://github.com/eclipse/californium/issues/500

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
AlexITC: oh, I got confused

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
AlexITC: Then, my understanding is that we now use a single client per endpoint, right?

sbernard31: This time that's me who is confused :p

At Leshan server side, nothing changes. (2 endpoints : 1 secured and 1 unsecured which allow to talk to any number of clients)

At Leshan client side, until now that was like as server side. But californium has limitation and does not allow several RPK or X509 certificate for the same coapEndpoint nor to change credential dynamically. So I tried this approach which recreates endpoint at bootstrap which will help to support X509 or RPK and  which could also help to support several dm server later. In that case the lwm2m client will have an endpoint per server.

But maybe https://github.com/eclipse/californium/tree/credentials_store will be a game-changer ...

sbernard31:
I rebase this PR on master and this would be the base for a new try to support RPK at client side (#538)

Tests seems pretty instable, I will investigate this.

@rkimsb2 you could finally be interested by this.

sbernard31:
I pushed some new modifications.
It seems the tests was really more stable now.
I will try to add RPK support tomorrow.

Issue #427, opened 2017-11-09
Why was Queue Mode removed from the Leshan Implementation?


carlosgp143:
Hi everybody,

I'm working in a queue mode implementation inside the Leshan Server Core, and I have just seen that there was an initial implementation of it, but it was completely removed in the commit [#2826841](https://github.com/eclipse/leshan/commit/28268410a0b88676ba8e52cf9c70fe90e4ac6cd0). I'm wondering what was the reason for this, because I couldn't found any information about it. 

Can someone please explain me what was the reason for doing this?

Thank you very much.

sbernard31:
We had a first implementation of queue mode in Leshan long time ago.
This implementation used an internal Queue to store pending request, but we found this not so flexible.
In the meantime, the LWM2M specification changed. OMA dropped this queue idea and replaced it by something more like an offline/online mode (without changing the name by the way).
So we removed this queue implementation from Leshan. (More details about that at #191)

A "work in progress PR" was proposed #265 with something simple and more in line with the new specification, but contributor seems to give up :'( ...

Currently, we don't think a queue mode is a priority as you can easily handle it just by using  RegistrationListener and/or ObservationListener.

But if you want to contribute Queue mode based on #265, you're welcome :)

joakimeriksson:
I have tried to find this specific change - where is the no-longer-queuing change? For easiest possible application development on the server side (assuming that the service / application should not know all details of the client / IoT device) it would make sense to have a higher abstraction than knowing the details of on/off state? Or? One possible approach for doing it could of course be to do a very lightweight component that binds in a way so that the application / service decides if it only needs to know on/off state or to use a more fully-fledge queue to collect the requests?! Did anyone ever evaluate the Q-mode mode of operation when it comes to power efficiency, etc?

sbernard31:
> I have tried to find this specific change - where is the no-longer-queuing change?

You should read the **§8.3 Queue mode operation.** of the specification  : 
_"As such, the Queue Mode offers functionality for a LwM2M Client to inform the LwM2M
Server that it may be disconnected for an extended period of time and also when it becomes reachable again."_
_" Despite the title of the functionality, i.e. Queue Mode, this specification does not mandate an implementation to use queues nor does it specify where such a queue would exist (or any details of such queuing mechanism)."_

> it would make sense to have a higher abstraction than knowing the details of on/off state?

Maybe, but we should first create this online/offline service and only  if this is really needed create an optional API build on this online/offline service.
[...]

PR #451, opened 2017-12-14
Add basic implementation of Queue Mode with awake/sleeping listeners

carlosgp143:
Here I present a new pull request for a simplified implementation of Queue Mode, based on #265. All the logic is done inside the Class LwM2mQueue, which is instantiated as an attribute inside the Registration class, and the client is tracked using the LwM2mClientStateListener.

As we discussed in #427, the client is considered in sleeping state after CoAP MAX_TRANSMIT_WAIT has expired, or when the client is not responding, which is considered when a request does not get any response after all the retransmissions.

In #265 the Online (Awake here) state is only notified when the client state was sleeping, but it think it should be notified always when an update message has arrived, because the server can know that the client is awake for sure at that moment, and it might have lost track of the client state and not changed it to Sleeping previously.


leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/LwM2mQueue.java
```diff
@@ -0,0 +1,208 @@
...
+                @Override
+                public void run() {
+                    if (!isClientSleeping()) {
+                        setSleeping();
+                    }
+                }
+            };
+            clientAwakeTimer.schedule(clientAwakeTask, clientAwakeTime);
+        }
+        timerSemaphore.release();
```
sbernard31: release should be in `finally` block.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/LwM2mQueue.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    /**
+     * Stop the timer that handles the client wait before sleep time.
+     */
+    private void stopClientAwakeTimer() {
+        try {
+            timerSemaphore.acquire();
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
sbernard31: ugh again ..

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
```diff
@@ -74,7 +75,58 @@
     private final String rootPath;
 
     private final Date lastUpdate;
+    
+    /** The LwM2m Queue Mode management object */
+    private LwM2mQueue queue;
```
sbernard31: Registration is immutable. This is a choice to better handle persistence and threadsafeness.
And as discussed in #265  I don't think presence state should be store in registration because we don't want to persist it or to share it between server instance(for cluster)

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LwM2mQueueModeRequestSender.java
```diff
@@ -0,0 +1,252 @@
...
+import org.eclipse.leshan.core.request.DownlinkRequest;
+import org.eclipse.leshan.core.response.ErrorCallback;
+import org.eclipse.leshan.core.response.LwM2mResponse;
+import org.eclipse.leshan.core.response.ResponseCallback;
+import org.eclipse.leshan.server.model.LwM2mModelProvider;
+import org.eclipse.leshan.server.registration.Registration;
+import org.eclipse.leshan.server.request.LwM2mRequestSender;
+import org.eclipse.leshan.util.Validate;
+
+public class LwM2mQueueModeRequestSender implements LwM2mRequestSender {
```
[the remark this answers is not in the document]
carlosgp143: The code is basically the same as the CaliforniumLwM2mRequestSender class but it needs to check if the client is sleeping to return without sending any request, and it needs to restart the client awake timer every time a new request is sent. 

I thought it was more clear to have two different RequestSenders, one for Queue-Mode and one for non Queue-Mode, but if you prefer both classes can be combined in just one.

leshan-server-core/src/main/java/org/eclipse/leshan/server/LwM2mServer.java
```diff
@@ -123,6 +124,12 @@
      */
     ObservationService getObservationService();
 
+    /**
+     * Get the queue mode service to access to the queue mode listeners. You can use this object for listening the
+     * client state.
+     */
+    QueueModeService getQueueModeService();
```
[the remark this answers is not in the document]
carlosgp143: I've changed it to PresenceService as in #265

carlosgp143:
Hi! 

Finally I've had some time to read @sbernard31 's feedback and make some changes. I've made a new commit with some of the minor ones, and after some discussion I can make a new commit with the more important ones.

Regarding:

> I also remarks there is formatting issue : did you configure eclipse correctly ?

I hope now everything is OK, otherwise I don't know what I'm doing wrong.

About:

> I don't see how the server can lost this state ? and if it does then it will consider client as offline, so new update will raise awake event. If we raise even each time we get an update I'm not sure to see the point of this event. It does nothing more than update registration event.

For example, if the server thinks that the client will stay awake for a long time, like `COAP_TRANSMIT_WAIT`, because it has that time as default, but the client stays awake for lower time because it wants to save more battery, then the server will loose track of the real state of the client. The state of the client remains `AWAKE` unless the server tries and fails in sending a request. Then if the sleeping time of the client is lower than `COAP_TRANSMIT_WAIT`, it will send a new update message, and the listeners won't be notified because the previous state of the client is `AWAKE` because it wasn't changed to `SLEEPING`. That's why I think it is necessary to notify the listeners with an `awake event` always when the update registration message arrives.

And the most important part:

> Registration is immutable. This is a choice to better handle persistence and threadsafeness.
And as discussed in #265 I don't think presence state should be store in registration because we don't want to persist it or to share it between server instance(for cluster mainly).

If the `LwM2mQueue object` that handles the state of the client and the client awake timer (and possibly in a near future the actual message queues) can't be inside the `Registration object`, then I think a good way to implement it is to have the `LwM2mQueue object` in a MAP inside the `PresenceServiceImp object` with the endpoint name as key.  Then, the endpoint name will link the `LwM2mQueue object` with the `Registration object`. Do you think this is a good implementation?

sbernard31:
Sry for the delay response.

**About UPDATE/AWAKE event :** 
1. If you raise this AWAKE event on each UPDATE why not just using the UPDATE event ?
2. The use case you expose sounds a bit strange to me : 
2.1. You say the client go to sleep sooner but wake up 1 or 2 minutes later, is it a real use case ?
2.2. From my point of view the AWAKE event  will be used to send a set of requests to a device, so in your case:
Either the device go to sleep before we send all requests and so we will know that the device is sleeping (a request sending will fail).
Or we send all the requests, the device is going to sleep, so in this case we don't know the device is sleeping while few minutes. The way the LWM2M spec is done we can be sure that the device is not here unless we failed to send request to it. But is it a problem ? 
For me AWAKE state means you can send request, SLEEP state means you must wait for AWAKE event. So if a new request is available while the client is consider as AWAKE we will send it immediately, if is not here we will detect that it is sleeping. If it is here we be able to send the request. (We will not know that it go to sleep during few minutes seconds but this is not an issue)

**About immutability:**
Creating a map in PresenceServiceImpl could be the right solution. This was done like this in #265

**How to organize ourself ?**
This PR is pretty big and we will go more and more to the incomplete solution proposed in #265.
So I would like to propose you another approach
1. I create a new branch `queue_mode` in which we will work about queue mode.
This branch could be used to integrate smaller PR about queue mode. Once queue mode is "correct" state we integrate it in master.
2. I propose to start from #265, do you see in this PR any issue or problem you would like to fix ?
If it seems ok for you. I can integrate it in `queue_mode` branch.
3. a PR to add a sender which support queue mode. (avoiding to much duplicated code, maybe using delegation)
4. a PR to add timeout feature (I mean going to SLEEP/OFFLINE after some minutes of inactivity, without semaphore unless you have good reason to use it ?)
5. Maybe a PR with a way to make this period of time different by clients/registration (This is what you would like to do right ?)

WDYT ?

carlosgp143:
Hi!

Thanks for your response :)

**About UPDATE/AWAKE event :**
1. For me it would be the same event but the UPDATE event is interesting for the `RegistrationListener` and the AWAKE event is interesting for the `PresenceListener`. But I think as Queue Mode is specified in LWM2M, both events are equal.

2. I'm gonna try to explain in a different way because I think we are misunderstanding:
2.1 Imagine that the Server thinks that the client is going to stay awake for 93 seconds (`COAP_TRANSMIT_WAIT`), but the client is an IoT device that wants to save energy and stays awake for 5 seconds, and sleeps for one minute. In that case, once is awake, the client will send an update message to notify the server. The serves changes the state from SLEEPING to AWAKE. Then, the client goes to sleep and after 5+60 seconds, it is awake again and sends another update message. At this moment, if the server hasn't try to send any request to the client, the state is still AWAKE. Then, this update message won't be notified to the `PresenceListeners` as an AWAKE event.
2.2 I was thinking in the special case that the server wants to send a set of request every time the AWAKE event is triggered, and no request at any other instant. Then, as I explained in 2.1, the AWAKE event can be missed at some points and the request won't be sent. Of course if the servers tries to make any request in the middle it will notice that the client is SLEEPING, but if not... 
So I think this case needs to be considered because for me is the most usual one when we are using queue mode. In the future if we introduce some way for the server to know what is the exact time that the client stays AWAKE, we can just trigger an AWAKE event if the client state is SLEEPING, but for the moment with `COAP_TRANSMIT_WAIT`, I think it should be triggered always when an update message arrives. 

**About immutability:**
Ok, I'll do that. Sorry I didn't notice that the `Registration object` is immutable.

**How to organize ourself ?**

If you think this is the best way to do it, I agree with that. 

#265 is a good initial point. There are a couple of things I don't agree with. First one is the observation listener that it introduces in order to change the state of the client when a Response arrives. I think that the way Queue Mode is specified now, the client state only depends on the Update messages. Second one is that the `ConcurrentMap` inside the `PresenceServiceImpl`should be change now to `<String, LwM2mQueue>` to introduce all the new features inside that object and not have only the `Presence` value, but this could also be done in the first smaller pull request that I'll do.
Also we need to reach an agreement about the Awake event. Now it is implemented as you say, but I would change it as I have explained. 

So do you want me to make an initial Pull Request to the `queue_mode` branch with this initial changes in #265 ? (after reaching an agreement in the AWAKE event). From that I can start introducing the smaller pull requests as you say, starting probably with the timeout feature.

sbernard31:
**About AWAKE event :**
I read your explanation and understand it well. It seems to me that in this case you aren't interested by AWAKE event but just by UPDATE event. This is not the same thing to me. I see update as a simple event and AWAKE notify for a state transition.
Previously in the specification that was : AWAKE event =  REGISTER or UPDATE or NOTIFICATION, now it seems this is more AWAKE = REGISTER or UPDATE.

> In the future if we introduce some way for the server to know what is the exact time that the client stays AWAKE, we can just trigger an AWAKE event if the client state is SLEEPING, but for the moment with COAP_TRANSMIT_WAIT, I think it should be triggered always when an update message arrives.

So we agree about destination, we just disagree about the transition state.
This is not so important to me, so let's go on your way for now.

**Details :**
I create the branch [queue_mode](https://github.com/eclipse/leshan/tree/queue_mode) starting from #265 rebased on master. When you push next PR, use this branch as "base".

>  starting probably with the timeout feature.

No problem about order :)

sbernard31:
I think we can close this PR.
If I'm wrong, do not hesitate to reopen it.

PR #454, opened 2018-01-05
Make redis data format compatibility break less painfull 

sbernard31:
In #440 we change the format using to serialize registration and observation in RedisRegistrationStore.

We already warn that RedisRegistrationStore is just an experimental works, so this kind of break would happen again.

I propose this PR to limit the migration painfulness.
The idea is that registration and observation are kind of "session" data. So for the rare case where this break occurred. Just ignoring old key could be acceptable.
That's why I introduce key prefix for RedisRegistrationStore in 2nd commit.
The first one add "optional" expiration for all redis key, this avoid to remove keys manually.

I'm aware this is not so great, but I think this could help.


ming2377:
Any chance this PR will be merged soon? We are using RedisRegistrationStore and we are hesitate to move to 1.0.0-M5 due to this issue. 

Can anyone please help explain the risk of moving from 1.0.0-M4 to 1.0.0-M5? We are using RedisRegistrationStore because we plan to scale-out to clustering DM nodes.

sbernard31:
As this PR seems to not needed by the community. I close it.

Feel free to revive it, if needed ;)

PR #455, opened 2018-01-12
Continue the Queue Mode implementation. Add client awake time feature.

carlosgp143:
This is the first of a set of pull request to introduce new features in the Queue Mode implementation, using #265 as starting point.

This first pull request makes two different things: make small changes to solve some problems that #265 has, and introduce the first feature: change the client state to SLEEPING after some time (`MAX_TRANSMIT_WAIT`).

**Main changes to 265**
1. First, I've removed the Observation Listener. Now the state of the client only depends on Registration and Update.
2. Perform all the implementation of the Queue Mode inside the `LwM2mQueue` class. In this class is where the new features can be added, for example the timer in this case. There is one object of this kind per client. As a result of this, there are some necessary changes inside the `PresenceService`.
3. Included in the `LeshanServer' the Queue Mode operation regarding request sending. If the client is sleeping, the 'send' methods return immediately and inform the user.

**New feature**
The feature added in this pull request is to set the client state to SLEEPING and notify the listeners when the time that the client states awake expires, which is `MAX_TRANSMIT_WAIT' as default. For it, a timer is started every time a new AWAKE event is triggered or a request is sent when the client is AWAKE, and when this timer expires, the state is changed.

Please note: the client state should change to SLEEPING also when a request is not responded by the client, but this feature will come with the new sender Pull Request.

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
@@ -12,6 +12,7 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
+ *     RISE SICS AB - added Queue Mode operation
  *******************************************************************************/
 package org.eclipse.leshan.server.californium.impl;
 
@@ -35,7 +36,6 @@ import org.eclipse.leshan.core.observation.Observation;
 import org.eclipse.leshan.core.request.DownlinkRequest;
 import org.eclipse.leshan.core.response.ErrorCallback;
 import org.eclipse.leshan.core.response.LwM2mResponse;
-import org.eclipse.leshan.core.response.ObserveResponse;
 import org.eclipse.leshan.core.response.ResponseCallback;
 import org.eclipse.leshan.server.Destroyable;
 import org.eclipse.leshan.server.LwM2mServer;
@@ -45,8 +45,8 @@ import org.eclipse.leshan.server.californium.CaliforniumRegistrationStore;
 import org.eclipse.leshan.server.californium.LeshanServerBuilder;
 import org.eclipse.leshan.server.impl.RegistrationServiceImpl;
 import org.eclipse.leshan.server.model.LwM2mModelProvider;
-import org.eclipse.leshan.server.observation.ObservationListener;
 import org.eclipse.leshan.server.observation.ObservationService;
+import org.eclipse.leshan.server.queue.LwM2mClientStateListener;
 import org.eclipse.leshan.server.queue.PresenceService;
 import org.eclipse.leshan.server.queue.PresenceServiceImpl;
 import org.eclipse.leshan.server.registration.Registration;
@@ -157,46 +157,7 @@ public class LeshanServer implements LwM2mServer {
         });
 
         // notify applications of LWM2M client coming online/offline
-        this.registrationService.addListener(new RegistrationListener() {
-
-            @Override
-            public void updated(RegistrationUpdate update, Registration updatedRegistration) {
-                presenceService.setOnline(updatedRegistration);
-            }
-
-            @Override
-            public void unregistered(Registration registration, Collection<Observation> observations) {
-                // Noop.
-
-            }
-
-            @Override
-            public void registered(Registration registration) {
-                presenceService.setOnline(registration);
-            }
-        });
-        this.observationService.addListener(new ObservationListener() {
-
-            @Override
-            public void onResponse(Observation observation, Registration registration, ObserveResponse response) {
-                presenceService.setOnline(registration);
-            }
-
-            @Override
-            public void onError(Observation observation, Registration registration, Exception error) {
-                presenceService.setOnline(registration);
-            }
-
-            @Override
-            public void newObservation(Observation observation, Registration registration) {
-                // Noop.
-            }
-
-            @Override
-            public void cancelled(Observation observation) {
-                // Noop.
-            }
-        });
+        this.registrationService.addListener(new LwM2mClientStateListener(this));
 
         // define a set of endpoints
         Set<Endpoint> endpoints = new HashSet<>();
@@ -321,24 +282,57 @@ public class LeshanServer implements LwM2mServer {
     @Override
     public <T extends LwM2mResponse> T send(Registration destination, DownlinkRequest<T> request)
             throws InterruptedException {
+        if (destination.usesQueueMode()) {
+            if (presenceService.isClientSleeping(destination)) {
+                LOG.info("The destination client is sleeping, request couldn't been sent.");
+                return null;
+            } else {
+                presenceService.getQueueObject(destination).startClientAwakeTimer();
+            }
+        }
         return requestSender.send(destination, request, DEFAULT_TIMEOUT);
+
     }
 
     @Override
     public <T extends LwM2mResponse> T send(Registration destination, DownlinkRequest<T> request, long timeout)
             throws InterruptedException {
+        if (destination.usesQueueMode()) {
+            if (presenceService.isClientSleeping(destination)) {
+                LOG.info("The destination client is sleeping, request couldn't been sent.");
+                return null;
+            } else {
+                presenceService.getQueueObject(destination).startClientAwakeTimer();
+            }
+        }
         return requestSender.send(destination, request, timeout);
     }
 
     @Override
     public <T extends LwM2mResponse> void send(Registration destination, DownlinkRequest<T> request,
             ResponseCallback<T> responseCallback, ErrorCallback errorCallback) {
+        if (destination.usesQueueMode()) {
+            if (presenceService.isClientSleeping(destination)) {
+                LOG.info("The destination client is sleeping, request couldn't been sent.");
+                return;
+            } else {
+                presenceService.getQueueObject(destination).startClientAwakeTimer();
+            }
+        }
         requestSender.send(destination, request, DEFAULT_TIMEOUT, responseCallback, errorCallback);
     }
 
     @Override
     public <T extends LwM2mResponse> void send(Registration destination, DownlinkRequest<T> request, long timeout,
             ResponseCallback<T> responseCallback, ErrorCallback errorCallback) {
+        if (destination.usesQueueMode()) {
+            if (presenceService.isClientSleeping(destination)) {
+                LOG.info("The destination client is sleeping, request couldn't been sent.");
+                return;
+            } else {
+                presenceService.getQueueObject(destination).startClientAwakeTimer();
+            }
+        }
         requestSender.send(destination, request, timeout, responseCallback, errorCallback);
     }
 
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
@@ -287,7 +287,7 @@ public class LeshanServer implements LwM2mServer {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return null;
             } else {
-                presenceService.getQueueObject(destination).startClientAwakeTimer();
+                presenceService.startClientAwakeTimer(destination);
             }
         }
         return requestSender.send(destination, request, DEFAULT_TIMEOUT);
@@ -302,7 +302,7 @@ public class LeshanServer implements LwM2mServer {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return null;
             } else {
-                presenceService.getQueueObject(destination).startClientAwakeTimer();
+                presenceService.startClientAwakeTimer(destination);
             }
         }
         return requestSender.send(destination, request, timeout);
@@ -316,7 +316,7 @@ public class LeshanServer implements LwM2mServer {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return;
             } else {
-                presenceService.getQueueObject(destination).startClientAwakeTimer();
+                presenceService.startClientAwakeTimer(destination);
             }
         }
         requestSender.send(destination, request, DEFAULT_TIMEOUT, responseCallback, errorCallback);
@@ -330,7 +330,7 @@ public class LeshanServer implements LwM2mServer {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return;
             } else {
-                presenceService.getQueueObject(destination).startClientAwakeTimer();
+                presenceService.startClientAwakeTimer(destination);
             }
         }
         requestSender.send(destination, request, timeout, responseCallback, errorCallback);
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
@@ -286,8 +286,6 @@ public class LeshanServer implements LwM2mServer {
             if (!presenceService.isClientAwake(destination)) {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return null;
-            } else {
-                presenceService.startClientAwakeTimer(destination);
             }
         }
         return requestSender.send(destination, request, DEFAULT_TIMEOUT);
@@ -301,8 +299,6 @@ public class LeshanServer implements LwM2mServer {
             if (!presenceService.isClientAwake(destination)) {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return null;
-            } else {
-                presenceService.startClientAwakeTimer(destination);
             }
         }
         return requestSender.send(destination, request, timeout);
@@ -315,8 +311,6 @@ public class LeshanServer implements LwM2mServer {
             if (!presenceService.isClientAwake(destination)) {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return;
-            } else {
-                presenceService.startClientAwakeTimer(destination);
             }
         }
         requestSender.send(destination, request, DEFAULT_TIMEOUT, responseCallback, errorCallback);
@@ -329,8 +323,6 @@ public class LeshanServer implements LwM2mServer {
             if (!presenceService.isClientAwake(destination)) {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return;
-            } else {
-                presenceService.startClientAwakeTimer(destination);
             }
         }
         requestSender.send(destination, request, timeout, responseCallback, errorCallback);
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
@@ -46,9 +46,9 @@ import org.eclipse.leshan.server.californium.LeshanServerBuilder;
 import org.eclipse.leshan.server.impl.RegistrationServiceImpl;
 import org.eclipse.leshan.server.model.LwM2mModelProvider;
 import org.eclipse.leshan.server.observation.ObservationService;
-import org.eclipse.leshan.server.queue.LwM2mClientStateListener;
 import org.eclipse.leshan.server.queue.PresenceService;
 import org.eclipse.leshan.server.queue.PresenceServiceImpl;
+import org.eclipse.leshan.server.queue.PresenceStateListener;
 import org.eclipse.leshan.server.registration.Registration;
 import org.eclipse.leshan.server.registration.RegistrationHandler;
 import org.eclipse.leshan.server.registration.RegistrationListener;
@@ -157,7 +157,7 @@ public class LeshanServer implements LwM2mServer {
         });
 
         // notify applications of LWM2M client coming online/offline
-        this.registrationService.addListener(new LwM2mClientStateListener(this));
+        this.registrationService.addListener(new PresenceStateListener(this.presenceService));
 
         // define a set of endpoints
         Set<Endpoint> endpoints = new HashSet<>();
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
@@ -283,7 +283,7 @@ public class LeshanServer implements LwM2mServer {
     public <T extends LwM2mResponse> T send(Registration destination, DownlinkRequest<T> request)
             throws InterruptedException {
         if (destination.usesQueueMode()) {
-            if (presenceService.isClientSleeping(destination)) {
+            if (!presenceService.isClientAwake(destination)) {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return null;
             } else {
@@ -298,7 +298,7 @@ public class LeshanServer implements LwM2mServer {
     public <T extends LwM2mResponse> T send(Registration destination, DownlinkRequest<T> request, long timeout)
             throws InterruptedException {
         if (destination.usesQueueMode()) {
-            if (presenceService.isClientSleeping(destination)) {
+            if (!presenceService.isClientAwake(destination)) {
                 LOG.info("The destination client is sleeping, request couldn't been sent.");
                 return null;
             } else {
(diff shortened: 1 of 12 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/LwM2mClientStateListener.java
```diff
@@ -0,0 +1,63 @@
...
+import org.eclipse.leshan.server.registration.RegistrationListener;
+import org.eclipse.leshan.server.registration.RegistrationUpdate;
+
+/**
+ * Listener that controls the state of the client (awake/sleeping) It is in charge of sending all the queued messages
+ * when the client is awake (has sent an update message), and controlling the time the client is awake before going to
+ * sleep.
+ */
+
+public class LwM2mClientStateListener implements RegistrationListener {
```
sbernard31: Naming : I would prefer something like PresenceStateListener

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/LwM2mClientStateListener.java
```diff
@@ -0,0 +1,63 @@
...
+
+    LwM2mServer server;
+    PresenceService presenceService;
+
+    public LwM2mClientStateListener(LwM2mServer server) {
+        this.server = server;
+        presenceService = server.getPresenceService();
```
sbernard31: It seems we don't need LwM2mServer as we only use `PresenceService`.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/LwM2mQueue.java
```diff
@@ -0,0 +1,200 @@
...
+
+import org.eclipse.leshan.server.registration.Registration;
+import org.eclipse.leshan.util.Validate;
+
+/**
+ * Class that contains all the necessary elements to handle the queue mode. Every registration object that uses Queue
+ * Mode has a LwM2mQueue object for handling it.
+ */
+
+public class LwM2mQueue {
```
sbernard31: Naming : I would prefer something like `PresenceState` or `PresenceStatus.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/LwM2mQueue.java
```diff
@@ -0,0 +1,200 @@
...
+
+    /* Elements to handle the time the client waits before going to sleep */
+    private Timer clientAwakeTimer;
+
+    private TimerTask clientAwakeTask;
+
+    private int clientAwakeTime;
+
+    /* Registration and Presence Service for notifying sleeping */
+    private Registration registration;
```
sbernard31: Registration should not be part of the state because object could change on update.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/LwM2mQueue.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+    }
+
+    /**
+     * Reads from the Californium.properties file the MAX_TRANSMIT_WAIT constant of the CoAP Protocol
+     * 
+     * @return MAX_TRANSMIT_WAIT
+     */
+    public int readCoapMaxTransmitWait() {
```
sbernard31: Californium.properties is californium stuff and `leshan-server-core` does not know Californium.
This introduce an implicit dependency. 

We should replace this by a parameter or a Constant.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/LwM2mQueue.java
```diff
@@ -0,0 +1,200 @@
...
+    private Timer clientAwakeTimer;
+
+    private TimerTask clientAwakeTask;
+
+    private int clientAwakeTime;
+
+    /* Registration and Presence Service for notifying sleeping */
+    private Registration registration;
+
+    private PresenceService presenceService;
```
sbernard31: We should not have a `PresenceService` attribute. This create a cyclic dependency which is often means bad design.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/Presence.java
```diff
@@ -20,7 +20,7 @@
  */
 public enum Presence {
     /** LWM2M Client is reachable and messages can be sent **/
-    ONLINE,
+    AWAKE,
     /** LWM2M Client is not reachable and no messages can be sent **/
-    OFFLINE
+    SLEEPING
```
sbernard31: Naming: I would prefer ONLINE/OFFLINE but as the spec use both words, I think both are acceptable.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -39,12 +40,34 @@
...
+     * @return true if the status is {@link Presence#SLEEPING}
+     */
+    boolean isClientSleeping(Registration registration);
+
+    /**
+     * Set the state of the client identified by registration as {@link Presence#AWAKE}
+     * 
+     * @param registration the client's registration object
+     */
+    public void setAwake(Registration registration);
```
sbernard31: I think this is internal API and we don't want to expose it.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -39,12 +40,34 @@
...
+
+    /**
+     * Notify the listeners that the client state changed to {@link Presence#SLEEPING}. The state changes is produced
+     * inside {@link LwM2mQueue} when the timer expires or when the client doesn't respond to a request.
+     * 
+     * @param registration the client's registration object
+     */
+    public void notifySleeping(Registration registration);
```
sbernard31: I think this is internal API and we don't want to expose it.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -39,12 +40,34 @@
...
+
+    /**
+     * Creates a new {@link LwM2mQueue} object associated with the client.
+     * 
+     * @param registration the client's registration object.
      */
-    boolean isOnline(Registration registration);
+    public void createQueueObject(Registration registration);
```
sbernard31: I think this is internal API and we don't want to expose it.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
```diff
@@ -47,54 +45,43 @@ public void removeListener(PresenceListener listener) {
...
-     * 
-     * @param registration lwm2m client registration data
-     */
-    public void setOffline(Registration registration) {
-        if (registration.usesQueueMode()
-                && transitState(registration.getEndpoint(), Presence.ONLINE, Presence.OFFLINE)) {
+    @Override
+    public void setAwake(Registration registration) {
+        if (registration.usesQueueMode()) {
+            clientStatusList.get(registration.getEndpoint()).setAwake();
```
sbernard31: queue could be null.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
```diff
@@ -400,6 +400,10 @@ public static void createAndStartServer(int webPort, String localAddress, int lo
             ServiceInfo coapSecureServiceInfo = ServiceInfo.create("_coaps._udp.local.", "leshan", secureLocalPort, "");
             jmdns.registerService(coapSecureServiceInfo);
         }
+        
+        /* Example of a listener that is notified every time the state of the client changes */
+        lwServer.getPresenceService().addListener(new ExamplePresenceListener("IPSO-objects-Firefly"));
+        lwServer.getPresenceService().addListener(new ExamplePresenceListener("hello"));
```
sbernard31: I suppose this should not be part of the PR ?

carlosgp143:
Thanks for your feedback @sbernard31 ! Don't worry about the naming, I understand it's important.

I can do a commit with the small changes you have commented first, and then try to think about how to move logic from `LwM2mQueue` to `PresenceService` to have a better implementation. It's totally ok for me if you also refactor the code and come with a different PR. 

> I suspect we need to better handle multi-threading.

What do you mean with this? I can also take a look to that if you explain it a bit deeper...

sbernard31:
> I can do a commit with the small changes you have commented first

Ok

>  then try to think about how to move logic from LwM2mQueue to PresenceService to have a better implementation

No problem, you can experiment.

> It's totally ok for me if you also refactor the code and come with a different PR.

Good to know :) (Sometime writing code is simpler than writing comments)

> What do you mean with this? I can also take a look to that if you explain it a bit deeper..

I feel that "state changing"+"timer task" should be done in a more atomic way but I did not think enough about that to give you a clear explanation.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
carlosgp143: How? When registration is created, the `LwM2mClientStateListener` creates a queue object inside the `PresenceServiceImpl` linked to the endpoint name, so it shouldn't be null... 

If you don't agree with that what would you do? Surround with try, catch or check first if it's not null and then call `setAwake()`?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
carlosgp143: Do you mean these methods shouldn't be public, or that they shouldn't be part of the interface at all?

carlosgp143:
Hi @sbernard31 !

I just made a new commit trying to move the logic from `PresenceStatus` (old `LwM2mQueue`) to `PresenceServiceImpl`.

The idea now is that `PresenceStatus` contains the state of the client, and the timer object, but not the methods to manage the timer. These methods have been moved to `PresenceServiceImpl`. My idea is to maintain this `PresenceStatus` object to put more Queue Mode features in the future, like for example the actual queues.

So now the `PresenceStateListener`, on every updated message, calls the `setAwake()` method of `PresenceServiceImpl` with the corresponding updated registration object as parameter. This object is used by `PresenceServiceImpl` to get the corresponding `PresenceStatus`object, change the state to AWAKE, get the timer and start it. I think this is also better for what you mentioned in:

> I feel that "state changing"+"timer task" should be done in a more atomic way but I did not think enough about that to give you a clear explanation." 

What do you think about this implementation?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
```diff
@@ -47,54 +47,119 @@ public void removeListener(PresenceListener listener) {
     }
 
     @Override
-    public boolean isOnline(Registration registration) {
-        return Presence.ONLINE.equals(clientStatus.get(registration.getEndpoint()));
+    public boolean isClientSleeping(Registration registration) {
+        PresenceStatus presenceStatus = clientStatusList.get(registration.getEndpoint());
```
sbernard31: This is a public API, so user could use it with a Registration which does not exist anymore, right?
So how does we handle this ?

Personally,  I will go for `isPresent` / `isAwake` / `isOnline` instead of `isClientSleeping`.
and `isPresent` means we have a status for this registration and the status is present.
(About naming I sill feel, that we talk more about presence than "awakeness")

I even ask myself if this could be better to implement sleeping state by "no entry in the map" (but this is maybe too soon to go for this solution)

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
```diff
@@ -47,54 +47,119 @@ public void removeListener(PresenceListener listener) {
...
+        clientStatusList.put(reg.getEndpoint(), new PresenceStatus(clientAwakeTime));
+    }
+
+    /**
+     * Returns the {@link PresenceStatus} object associated with the given endpoint name.
+     * 
+     * @param reg The client's registration object.
+     * @return The {@link PresenceStatus} object.
+     */
+    public PresenceStatus getPresenceStatusObject(Registration reg) {
```
sbernard31: I think this should not be public, right ?

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
```diff
@@ -47,54 +47,119 @@ public void removeListener(PresenceListener listener) {
...
+    public void createPresenceStatusObject(Registration reg) {
+        clientStatusList.put(reg.getEndpoint(), new PresenceStatus());
+    }
+
+    /**
+     * Creates a new {@link PresenceStatus} object associated with the client, with a specific awake time.
+     * 
+     * @param registration the client's registration object.
+     */
+    public void createPresenceStatusObject(Registration reg, int clientAwakeTime) {
```
sbernard31: I ask myself, if this is not better to create automatically the status when it does not exist on setAwake ?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
```diff
@@ -321,24 +282,57 @@ public LwM2mModelProvider getModelProvider() {
     @Override
     public <T extends LwM2mResponse> T send(Registration destination, DownlinkRequest<T> request)
             throws InterruptedException {
+        if (destination.usesQueueMode()) {
+            if (presenceService.isClientSleeping(destination)) {
+                LOG.info("The destination client is sleeping, request couldn't been sent.");
+                return null;
+            } else {
+                presenceService.startClientAwakeTimer(destination);
```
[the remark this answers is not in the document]
carlosgp143: A far as I have understood the LWM2M protocol, the client stays awake for `MAX_TRANXMIT_WAIT` after the last received request. So it will restart its internal timer on the reception of a request. So if we refresh the timer when we send the request, it will have a small drift with respect to the client's timer, due to the request transmit time. But if we restart it when we receive a response, it will have also a small drift due to the response transmit time, and the request processing time. So I think the question is what drift do we prefer? 

For me, it would be better to refresh it before sending the request, so that the server timer expires first and considers that the client is sleeping a little bit before it really is, rather than consider that it is sleeping a little bit after it really is.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
carlosgp143: I am using the Q-Mode with actual hardware IoT nodes, so for me it is more intuitive to use the AWAKE/SLEEPING naming. But if you really want it, I can change it to the ONLINE/OFFLINE one.

Apart from that I agree about "isPresent means we have a status for this registration and the status is present".

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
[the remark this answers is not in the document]
sbernard31: But this is more complicated than that because this is not just : 
```
Request ->  
<- Response  
```
In CoAP, there is request acknowledgement which could be separated of response and even worst you could imagine scenario with blockwise transfer that means 1 LWM2M request means several CoAP requests.
([How timeout is handled](https://github.com/eclipse/leshan/wiki/Request-Timeout) give an idea of the complexity)

In the spec, nothing say that the client should response in MAX_TRANSMIT_WAIT.
It must wait MAX_TRANSMIT_WAIT after the last request before to go to sleep but between a request and a response time could be longer.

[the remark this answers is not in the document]
carlosgp143: Ok I understand it, I've simplified it too much. Then we agree that the client will restart its timer when it gets a request.

But then when should the server restart this timer? I don't see any good solution because:
1. If restart when the coap request is ACK, wouldn't work for non-confirmable messages.
2. If restart when response, as you said the time between the request and the response could be much more larger than the actual awake time of the client.
So... any other idea?

carlosgp143:
New commit with some changes. You can check it! 

> Entry in PresenceServiceImpl map is never removed.

Now it is removed when the client is unregistered by the `PresenceStateListener`.

> We should probably enhance the threadsafeness of PresenceService but modification you have done will help a lot.

IF you give me some more tips about this I can tr to work on it...

> Not sure but Timer is a bit old school class maybe we should consider to use ScheduledThreadPoolExecutor instead.

It's true, I checked and they recommend to use ScheduledThreadPoolExecutor instead of Timer. So I changed the code in order to do that, it wasn't a big thing.

So I'm guessing that the only issues left are:
1. When to restart the timer?
2. What do we need to change to enhance the threadsafeness?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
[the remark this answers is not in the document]
sbernard31: FMPOV, we will have difficulty to exactly know when the device is here or not..

So we should keep in mind this will just be **an approximation**.
For me this approximation must be good enough to allow users to sends requests to the client and not so more.

So, 2 solutions : 
1. We set the client awake on Response. Setting awake means change state + relaunch timer.
=> This means, in some cases we could consider the client sleeping although it is awake, but once we receive the response we will consider it awake again and so could send the next requests.
2. We pause the timer before to send the request and reactive it when we receive the response.
=> So we will not go in sleeping mode waiting we receive the response but this could be long.

Personally, I will go the 1. approximation as this is the simplest one and we will see later if this approximation cause troubles.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceService.java
```diff
@@ -47,5 +47,5 @@
      * @param registration data to check
      * @return true if the status is {@link Presence#SLEEPING}
      */
```
sbernard31: javadoc is not OK anymore

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceStatus.java
```diff
@@ -28,7 +30,9 @@
     private Presence state;
 
     /* Elements to handle the time the client waits before going to sleep */
-    private Timer clientAwakeTimer;
+    ScheduledExecutorService clientAwakeExecutor = Executors.newSingleThreadScheduledExecutor();
```
sbernard31: Create 1 scheduler by device sounds so good.
This is more about creating 1 scheduler for all tasks in PresenceService.

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceStatus.java
```diff
@@ -70,8 +74,8 @@ public void setSleeping() {
      * 
      * @return true if client is sleeping
      */
```
sbernard31: java doc

sbernard31:
About Threadsafeness :

This is about being sure that we can not go in incoherent state, mainly because of race condition.

Often, the idea is to think about what should be done in an atomic way.
Then trying to use java concurrent tools to achieve this.

First, trying with simple tool like AtomicReference, ConcurentHashMap ...
If this is not enough, going with more complex code using  synchronized (Lock), ReadWriteLock ...

leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceStatus.java
carlosgp143: But then you need to specify the number of tasks when creating the sheduler, and we don't now how many clients will be registered at a time... that's why I took this approach.

sbernard31: This is not the number of tasks, this is more the number of workers.
A scheduler have 1 or severals workers/thread.
You can schedule more tasks than workers available.
The only risk is that task could be delay a bit.

carlosgp143: Ok, I've misunderstood it. So then with just one single thread executor we can handle all the timer tasks of the registered clients, right?

sbernard31: Yep

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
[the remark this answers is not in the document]
carlosgp143: I totally agree that this will be an approximation. I think the first option is also better because it feels more safe for me to consider that the client is sleeping when it is actually awake than the other way around.

So I guess this implementation needs to go in the next pull request where the Queue Mode sender is added so that we can add the timer restarting inside the onResponse of the message observer. So I think I'll just remove the restarting of the timer from here and we will discuss more about it in that PR, once this one is accepted. Do you agree with that?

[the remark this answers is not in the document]
sbernard31: It sounds good to me :)

carlosgp143:
I just made one commit changing to use just one Scheduled executor and to not restart the timer when a request is sent.

So I guess the only thing left in this PR is Threadsafeness. I only see that there can be problems when setting the client's state to awake or sleeping, and starting or stopping the timer. So... should we surround this actions with a Lock to ensure the threadsafeness? Otherwise I don't see anything else to do.

sbernard31:
It sounds good.

> I only see that there can be problems when setting the client's state to awake or sleeping, and starting or stopping the timer. So... should we surround this actions with a Lock to ensure the threadsafeness? Otherwise I don't see anything else to do.

I agree. You can make it on this PR or we can also do that at the end (because the current design should allow us to do that later easily)

carlosgp143:
Hi! I did it in this PR cause I think it's easier. 

So I guess this PR is ready to be merged into he queue_mode branch so that I move forward to the Queue Mode request sender... right? :-)

sbernard31:
I will do that tomorrow.
You can start to go with a queue mode request sender. (Maybe an idea to reuse code is to create a sender which use the Current sender implementation using delegation)

sbernard31:
Integrated in queue_mode branch.

carlosgp143:
Hi! Just a couple of questions, cause I'm a bit confused...

1. Why didn't you integrate the last commit also? The "Added lock for atomic setAwake+starTimer and setSleeping+stopTimer" that I did yesterday?
2. Why did you close this PR and not merge it into the queue_mode branch? Maybe I'm a bit unfamiliar with the usual PR flow... but I don't know the difference.

sbernard31:
1. because I just forgot it. Thx for you vigilance ! It is integrated now.
2. Just because I am used to integrate PR manually. :)

carlosgp143:
Hey!

One question that comes to my mind when I think in how to implement the new Queue Mode Request Sender for the next PR. You said:

> You can start to go with a queue mode request sender. (Maybe an idea to reuse code is to create a sender which use the Current sender implementation using delegation)

Which, if I've understood correctly, means that the new sender will have the CaliforniumRequestSender as an attribute and will use its methods for sending requests. But inside this methods the CaliforniumRequestSender adds the corresponding Message observers, which are the ones we are interested in changing, for restarting the timer.

So... I don't think we can reuse much code... unless we change a bit the code of CaliforniumRequestSender and introduce the message observers as a parameter for the send methods. Otherwise I think the new sender needs to copy paste some code and add its own observers, in addition to check the state of the client to send or not. 

What's your opinion about this?

sbernard31:
> Which, if I've understood correctly, means that the new sender will have the CaliforniumRequestSender as an attribute and will use its methods for sending requests. But inside this methods the CaliforniumRequestSender adds the corresponding Message observers, which are the ones we are interested in changing, for restarting the timer.

That's my idea.

>So... I don't think we can reuse much code... unless we change a bit the code of CaliforniumRequestSender and introduce the message observers as a parameter for the send methods. Otherwise I think the new sender needs to copy paste some code and add its own observers, in addition to check the state of the client to send or not.

I'm not sure to understand ?

What I have in mind should look like this : 
```java
 public <T extends LwM2mResponse> void send(final Registration destination, DownlinkRequest<T> request,
            final ResponseCallback<T> responseCallback, ErrorCallback errorCallback) {
       // Here do you specific test about state.
      // if you can send, use internal sender
        internalRequestSender.send(destination, request, new ResponseCallback<T>() {
            @Override
            public void onResponse(T response) {
                 // restart timer
                 // call the user callback.
                responseCallback.onResponse(response);
            }
        }, errorCallback);
     }
     // you probably need a custom error callback to support timeout and set the device sleeping.
     }
```

carlosgp143:
Ok! I was thinking in using the methods with the SyncRequestObserver, but you're right, using the asynchronous ones makes it much easier.

So there should be a new PR quite soon :-)

PR #461, opened 2018-01-26
LwM2m Request Sender for Queue Mode

carlosgp143:
Here comes a new PR to continue the implementation of Queue Mode :-)

In this case, a new request sender is created, that uses methods from the `CaliforniumRequestSender` forming a delegation pattern. The tasks that the Queue Mode Request sender does are:

1. Before sending, checks if the client is sleeping, in that case it returns immediately and warns the user.
2. Set the client to `AWAKE` (this restarts the timer internally) when a response arrives. 
3. Set the client to `SLEEPING` if a timeout occurs, which means that the client does not respond so it is sleeping.

For delegation, the asynchronous method of `CalifornoumRequestSender` is used for both synchronous and asynchronous metods of the `QueueModeRequestSender`. This allows to reuse the code, because for using the synchronous method, the message observer should be changed to overwrite onResponse() and onTimeout(). For doing it synchronous, a new `ResponseCallback` class is created with an internal latch that stops the execution until a response arrives or a timeout occurs.

I have also a question regarding the approximation we have been discussing in #455 about when to restart the timer.
1. Should we consider to restart the timer in `onResponse()` with a time slightly smaller? The client restarts its timer when it gets the request, so if the server restarts it on a response, there will be a time in the future when the server will consider that the client is awake when it's actually not. Maybe the request sender can measure the round trip time and subtract half of it to the timer, or something like that... 

Maybe that is too complicated and not so important, it's just an idea to improve the approximation.

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
@@ -87,6 +87,8 @@ public class LeshanServer implements LwM2mServer {
 
     private final LwM2mRequestSender requestSender;
 
+    private final LwM2mRequestSender queueModeRequestSender;
+
     private final RegistrationServiceImpl registrationService;
 
     private final ObservationServiceImpl observationService;
@@ -147,7 +149,11 @@ public class LeshanServer implements LwM2mServer {
             @Override
             public void unregistered(Registration registration, Collection<Observation> observations, boolean expired,
                     Registration newReg) {
-                requestSender.cancelPendingRequests(registration);
+                if (registration.usesQueueMode()) {
+                    queueModeRequestSender.cancelPendingRequests(registration);
+                } else {
+                    requestSender.cancelPendingRequests(registration);
+                }
             }
 
             @Override
@@ -194,6 +200,7 @@ public class LeshanServer implements LwM2mServer {
         // create sender
         requestSender = new CaliforniumLwM2mRequestSender(endpoints, this.observationService, modelProvider, encoder,
                 decoder);
+        queueModeRequestSender = new QueueModeLwM2mRequestSender(this.presenceService, this.requestSender);
     }
 
     @Override
@@ -283,10 +290,7 @@ public class LeshanServer implements LwM2mServer {
     public <T extends LwM2mResponse> T send(Registration destination, DownlinkRequest<T> request)
             throws InterruptedException {
         if (destination.usesQueueMode()) {
-            if (!presenceService.isClientAwake(destination)) {
-                LOG.info("The destination client is sleeping, request couldn't been sent.");
-                return null;
-            }
+            return queueModeRequestSender.send(destination, request, DEFAULT_TIMEOUT);
         }
         return requestSender.send(destination, request, DEFAULT_TIMEOUT);
 
@@ -296,10 +300,7 @@ public class LeshanServer implements LwM2mServer {
     public <T extends LwM2mResponse> T send(Registration destination, DownlinkRequest<T> request, long timeout)
             throws InterruptedException {
         if (destination.usesQueueMode()) {
-            if (!presenceService.isClientAwake(destination)) {
-                LOG.info("The destination client is sleeping, request couldn't been sent.");
-                return null;
-            }
+            return queueModeRequestSender.send(destination, request, timeout);
         }
         return requestSender.send(destination, request, timeout);
     }
@@ -308,10 +309,8 @@ public class LeshanServer implements LwM2mServer {
     public <T extends LwM2mResponse> void send(Registration destination, DownlinkRequest<T> request,
             ResponseCallback<T> responseCallback, ErrorCallback errorCallback) {
         if (destination.usesQueueMode()) {
-            if (!presenceService.isClientAwake(destination)) {
-                LOG.info("The destination client is sleeping, request couldn't been sent.");
-                return;
-            }
+            queueModeRequestSender.send(destination, request, DEFAULT_TIMEOUT, responseCallback, errorCallback);
+            return;
         }
         requestSender.send(destination, request, DEFAULT_TIMEOUT, responseCallback, errorCallback);
     }
@@ -320,10 +319,8 @@ public class LeshanServer implements LwM2mServer {
     public <T extends LwM2mResponse> void send(Registration destination, DownlinkRequest<T> request, long timeout,
             ResponseCallback<T> responseCallback, ErrorCallback errorCallback) {
         if (destination.usesQueueMode()) {
-            if (!presenceService.isClientAwake(destination)) {
-                LOG.info("The destination client is sleeping, request couldn't been sent.");
-                return;
-            }
+            queueModeRequestSender.send(destination, request, timeout, responseCallback, errorCallback);
+            return;
         }
         requestSender.send(destination, request, timeout, responseCallback, errorCallback);
     }
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
@@ -87,8 +87,6 @@ public class LeshanServer implements LwM2mServer {
 
     private final LwM2mRequestSender requestSender;
 
-    private final LwM2mRequestSender queueModeRequestSender;
-
     private final RegistrationServiceImpl registrationService;
 
     private final ObservationServiceImpl observationService;
@@ -149,11 +147,8 @@ public class LeshanServer implements LwM2mServer {
             @Override
             public void unregistered(Registration registration, Collection<Observation> observations, boolean expired,
                     Registration newReg) {
-                if (registration.usesQueueMode()) {
-                    queueModeRequestSender.cancelPendingRequests(registration);
-                } else {
-                    requestSender.cancelPendingRequests(registration);
-                }
+
+                requestSender.cancelPendingRequests(registration);
             }
 
             @Override
@@ -198,9 +193,8 @@ public class LeshanServer implements LwM2mServer {
         coapServer.add(rdResource);
 
         // create sender
-        requestSender = new CaliforniumLwM2mRequestSender(endpoints, this.observationService, modelProvider, encoder,
-                decoder);
-        queueModeRequestSender = new QueueModeLwM2mRequestSender(this.presenceService, this.requestSender);
+        requestSender = new QueueModeLwM2mRequestSender(this.presenceService,
+                new CaliforniumLwM2mRequestSender(endpoints, this.observationService, modelProvider, encoder, decoder));
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
```diff
@@ -87,6 +87,8 @@
 
     private final LwM2mRequestSender requestSender;
 
+    private final LwM2mRequestSender queueModeRequestSender;
```
sbernard31: I propose to have only 1 LwM2mRequestSender attribute in LeshanServer.

We keep the 2 implementations QueuemodeSender and "CaliforniumSender".

QueueModeSender will do something like this:

```java
if (!registration.usesQueueMode()) {
     delegate.send(....);
} else {
    // DO queueu mode sutff
```

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/QueueModeLwM2mRequestSender.java
```diff
@@ -0,0 +1,164 @@
...
+
+        // If the client is sleeping, warn the user and return
+        if (!presenceService.isClientAwake(destination)) {
+            LOG.info("The destination client is sleeping, request couldn't been sent.");
+            return null;
+        }
+
+        // Create a response callback to use the asynchronous send method in a synchronous way
+        ResponseCallbackQueueModeSynchronous<T> synchronousResponseCallback = new ResponseCallbackQueueModeSynchronous<T>(
+                presenceService, destination);
```
sbernard31: Why we don't use the  delegatedSender synchronous API ?
(I remember you already try to explain me that but I probably miss-understood)

carlosgp143: We cannot directly use the function `public <T extends LwM2mResponse> T send(final Registration destination, final DownlinkRequest<T> request, long timeout)` because it is inside this function where the message observer is added to the coap request. Then, we cannot change it to perform the required actions in the methods onResponse() and onTimeout() (or at least I don't see a way).

So this method could be change to take the SyncRequestObserver as a parameter, and create a new one that overrides the onResponse() and onTimeout() methods. Or on the other hand we can use the asynchronous method with the callback that I created to make it synchronous.

sbernard31: Why not something like this ? 
```java
T response = delegatedSender.send(destination, request, timeout);
        if (response != null) {
            presenceService.setAwake(destination);
        } else {
            presenceService.clientNotResponding(destination);
        }
        return response;
```

carlosgp143: You're totally right... the `waitForResponse()` will return when a response or a timeout occurs. 

Sometimes we try to do things much more complicated than they really are :-).

I'll change this two things from your comments and come with a new commit!

carlosgp143:
Hi! I made a new commit changing to only have one request sender in the LeshanServer and to use the synchronous API.

One question that came to my mind:
If the Request sender calls the `setAwake()` method on every response, to change the state to awake and restart the timer in case it has expired and the server has considered the state of the client as `SLEEPING`, the listeners will be notified with the `onAwake()` method on every response. If we imagine the typical Queue Mode scenario where the server waits for the client to be awake, and send a set of requests in that moment, it will do that all the time when a response arrives. 

So for me there are two different events: the first time when the client is awake and sends the update message, and the possible future times when the client has considered to be sleeping (and the listeners have been notified) but it wasn't. 

Should we trigger two different events, for example `onFirstAwake()` when the update message arrives, and `onAwake()` when a response arrives so that the listeners can know that the client is still awake? Any other idea?

sbernard31:
We go back to our disagreement about : **raising event each time we get an UPDATE** vs **raising event only if state changes**.

If we go with **raising event only if state changes** we don't face the issue you mentioned, right ?

FMPOV, the event Awake/sleeping should help to be able to know if a request sending could be sent now or delayed.

If you want to send a message each time the client say "Hello" (which seems to be the use case you have in mind) using UPDATE event sounds more appropriate to me.

This way we have 3 events, UPDATE, AWAKE or NEW RESPONSE (I talked about the ResponseCallback) which cover several use cases.

carlosgp143:
> If we go with raising event only if state changes we don't face the issue you mentioned, right ?

This will happen only if the response arrives before the timer has expired. Otherwise, it triggers the `setSeleeping()` with the timer (when the client is not really sleeping) and then when the response arrives, it triggers the `setAwake()`. So there will be state change in this situation, even when the client state hasn't change in reality. That's why I think we should differentiate between the normal AWAKE event, and the "fake" AWAKE event that is created by `onResponse()`, which does not mean that the state really changed, only that the server was considering  the client as SLEEPING when it was AWAKE in reality. For the SLEEPING event this can't be avoided, we can just trigger it on every timer expiration, because we don't know if a response will come in a future or not.

> FMPOV, the event Awake/sleeping should help to be able to know if a request sending could be sent now or delayed.
If you want to send a message each time the client say "Hello" (which seems to be the use case you have in mind) using UPDATE event sounds more appropriate to me.

I agree that for the use case when the client says "Hello" an UPDATE event can be used. But FMPOV this limits the usage of listeners. If they are only used to see if a request can be sent or needs to be delayed... that could be done just with the `QueueModeLwM2mRequestSender`, by trying to send and if the client is sleeping, delay the request. I think the listeners should be useful for other actions too, and that they should be able to differentiate between a real AWAKE event, and the "fake" one .

sbernard31:
I'm not really convinced. I still think that our implementation should focus on making possible the queue mode as it is described in the specification. That means helping to "send or delay response".

> that could be done just with the QueueModeLwM2mRequestSender, by trying to send and if the client is sleeping, delay the request. 

That could even be done without a special sender, I mean currently we are using Leshan in our production server and our devices are using QueueMode. And we are able to do that without PresenceService and QueueModeSender. Just using UPDATE event and request TIMEOUT. That's why this was not implement sooner because it's possible to leave without it.

So question is : "What is the benefit of a this QueueMode implementation ? 
- I think the main one is not trying to send a request if we know(believe) that device is not here.
- Avoid to handle 2 consecutive UPDATE event as only one WAKEUP (e.g. :  2 UPDATEs could happen if device raises a first UPDATE because of it wakes up and a second one because it supports a new object)
So I see that just as a kind of optimization.

I think trying to support any other hypothetic use case is too soon. I would prefer to deliver the minimal implementation and see what it behave in real world.

BUT, if you **still** think that being able to make difference between kind of AWAKE/SLEEPING event should be done **now**.
I prefer we consider to add a `cause` or boolean parameter to AWAKE and SLEEPING event.

carlosgp143:
I see your point... I agree that this implementation should respect the specification, but I don't completely agree about restricting it to this minimal implementation. Adding stuff that could be useful but still respect the specification sounds good to me. I'm currently running power consumption experiments with actual IoT devices that stay awake for very short times using queue mode to save as much energy as possible, and where I see that this kind of different events could be useful.

But anyway, if you prefer to have first a minimal implementation as you explained for now, that's also fine for me. We could keep adding features in the future.

Then, I agree with you in the fact that the listener should be only notified if the state has change. So I'll change that in the code and make a new commit in this PR. And that should be all here I think... do you agree?

sbernard31:
It sounds good to me :)

carlosgp143:
Done :-)

sbernard31:
Integrated in queue_mode branch.

What are the next step ?
I think we should find a better way than just logging when we try to send a request to a sleeping device.
Maybe using a new Exception ? `DeviceOfflineException`.(or another name with Sleep)

Did you think about something else ?

carlosgp143:
Nice :-)

I agree with you, that would be nice because now when you try to send a request using the web application it informs you that there is a timeout, which is not really nice. I'll do a pull request to throw another type of exception.

For this first version of the Queue Mode I cannot think in anything else for being implemented. FMPOV, after the new Exception it could be integrated in the master branch...

The next steps for me are adding more features to extend this implementation (version 2 or something), like actual queues to store the requests or a way to lower down the 93s of awake time (by creating a new resource in the client with this value or a new parameter in the query of the update message). All this is out of the specification, but could be useful for future developers (of course always giving the possibility to enable/disable this additional features), and maybe give feedback to OMA for future specifications of the Queue Mode.

sbernard31:
About a way to lower down the 93s, creating an object is probably more in the LWM2M specification spirit, but I think it would be maybe smarter to just add a new parameter at registration. This parameter will be accessible via `additionalRegistrationAttributes` in `Registration`. (This is already implemented).
A custom PresenceService implementation could use this parameter to know  how many time to wait before considering client is sleeping.

carlosgp143:
Yes, I agree that this is the most suitable solution, because then the server does not need to make any request in order to know it, but it will come in the registration/update message.

I will come soon with a pull request about the "PresenceSleeping" exception. 

How do you want to proceed? Should the "reduce 93 seconds" feature come as a PR to this branch also or do you want to integrate the basic Queue Mode implementation first and then make a new branch for additional features?

sbernard31:
I prefer to begin with a basic QueueMode implementation (I mean without the "reduce 93 seconds")

I think a bit more about next step.
1. We should add some unit tests in `leshan-integration-tests` about QueueMode.
2. We should make the 93s configurable. For this maybe  we could introduce a class which is responsible  to give the time to wait for a given registration. This class could be settable via `LeshanServerBuilder`. We could have a default implementation which always return 93s for all registration. Users could also set it via `LeshanServerBuilder` using another default value. Someone who want a more dynamic one, could implement its own version. (e.g. using `additionalRegistrationAttributes`)

Does it sounds good ?

carlosgp143:
I think is also better to have a basic implementation first. So for this version... any other steps apart from integrating #467 with the sleeping exception? The unit tests you mentioned should be included in this version before integrating?

1. What things should be tested? I'm not really familiar with these kind of tests... I guess I should look deeper in the already implemented `leshan-integration-tsts`. But I would appreciate some guidelines :-).
2. Let's see if I have understood correctly, you mean a new object as an attribute in the `LeshanServer` that the user can set using the `LeshanServerBuilder`. This object will return the time, so that `PresenceService` can use it to create the `PresenceStatus` object with it. To do it dynamically, this object instead of a time, informs the `PresenceService` to take the time from the `additionalRegistrationAttributes`. Is this what you meant? It sounds like a very nice idea to me. But this shouldn't be included in the first version, right?

sbernard31:
Before to integrate this is master : 

A. I would like to play a bit with it. (I mean run it and test behavior)
B. I will probably add a way to deactivate the new queue_mode to have the legacy behavior via LeshanBuilder.

About adding unit tests and the flexible way to handle "the time before to go to sleep", I have no strong opinion about integrated it in a first or in a second version, so it's up to you.

>1.What things should be tested?

Just main scenario, a client register with queue mode, server try to send a request success, we wait for sleeping event then we try to send a request it failed, it send an update we receive awake event, we send another update we don't receive awake event, something like this.
But now I think about that I feel  we need a way to configure a time to wait lesser than 93s to avoid to have tests which are too long.
So we should implement 2. before to be able to do 1.

2. That's what I mean.

By the way, I rebase queue_mode branch on master, so do not forget to fetch --force before to start to work on this again.

carlosgp143:
Sounds good! If you find any issues whiles testing let me know. Also having a way of deactivating the queue_mode sounds like a good option. Do you want to add it? I could also look into it if you prefer...

I'll move on to the "reduce 93s feature" and the integration tests, and I'll come back with a PR when the first version is ready and integrated in master :-)

PR #467, opened 2018-02-07
Throw ClientSleepingException when a request tries to be sent in this state

carlosgp143:
Here is a small pull request to throw a new type of exception when the client is sleeping, and not just warn the user using the logger.

The client servlet in the server demo code is also changed to inform the user with the correct message when this situation happens.

I'm not really familiar about how exceptions work in Leshan, but I think I have made all the necessary changes... If not please just inform me!

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/QueueModeLwM2mRequestSender.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/QueueModeLwM2mRequestSender.java
@@ -16,6 +16,7 @@
 package org.eclipse.leshan.server.californium.impl;
 
 import org.eclipse.leshan.core.request.DownlinkRequest;
+import org.eclipse.leshan.core.request.exception.ClientSleepingException;
 import org.eclipse.leshan.core.request.exception.TimeoutException;
 import org.eclipse.leshan.core.response.ErrorCallback;
 import org.eclipse.leshan.core.response.LwM2mResponse;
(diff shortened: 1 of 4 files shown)


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/QueueModeLwM2mRequestSender.java
```diff
@@ -94,8 +94,7 @@ public QueueModeLwM2mRequestSender(PresenceServiceImpl presenceService, LwM2mReq
 
         // If the client is sleeping, warn the user and return
         if (!presenceService.isClientAwake(destination)) {
-            LOG.info("The destination client is sleeping, request couldn't been sent.");
-            return;
+            throw new ClientSleepingException("The destination client is sleeping, request cannot be sent.");
```
sbernard31: I'm asking myself if we should return the exception via the ErrorCallback or like you are doing.

My first idea was to use ErrorCallback but maybe you're right and its better like this.

carlosgp143: FMPOV the `ErrorCallback` throw an exception that indicates the client is sleeping... so `PresenceService` sets the state, and then as soon as a new response tries to be sent, the sleeping exception is thrown immediately. 

It makes sense to me, but it is true that the `ErrorCallback` could also throw the sleeping exception in addition to execute `clientNotResponding` to inform the user directly here and not when a new request tries to be sent. 

So both options are valid for me, either how it is now, or throw the exception both at the begging if the client is sleeping and in the error callback. What do you think is best?

sbernard31: let's start like this and we will adjust this later if necessary I will integrate this in queue_mode branch

PR #468, opened 2018-02-09
Add way to deactivate PresenceService

sbernard31:
@carlosgp143 Can you review this PR.

By looking more deeper at the code, I think that we will be able to simplify PresenceService implementation but we will see later after the additions of tests and a flexible way to handle time to sleep.

I also feel that we should identity PresenceStatus by registration ID, but I need to think more about that.

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
@@ -86,6 +86,7 @@ public class LeshanServerBuilder {
 
     private boolean noSecuredEndpoint;
     private boolean noUnsecuredEndpoint;
+    private boolean noQueueMode = false;
 
     /**
      * <p>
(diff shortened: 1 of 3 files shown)


carlosgp143:
Hi! I think everything is fine :-) this could be included and it's very easy to deactivate the support manually.

> By looking more deeper at the code, I think that we will be able to simplify PresenceService implementation but we will see later after the additions of tests and a flexible way to handle time to sleep.

I tried to made it as simple as possible, but of course we can simplify it more taking feedback from the tests and so on.

> I also feel that we should identity PresenceStatus by registration ID, but I need to think more about that.

Isn't it enough with being linked to the "endpoint name" inside `PresenceServiceImpl` ?

sbernard31:
> Isn't it enough with being linked to the "endpoint name" inside PresenceServiceImpl ?

I'm a bit traumatized by race condition and I feel that we can face some issues with simultaneous deregister/register using endpoint instead of registrationid but we will see this in our next iteration.

I will integrate this PR in queue_mode branch. I will also rewrite the queue_mode branch history to make it clearer before to integrate it in master. @msangoi will maybe take a look at the beginning of the next week. So I will integrate it in master 2nd part of the next week.

carlosgp143:
Those are very good news! Looking forward to the integration :-)

Issue #473, opened 2018-02-16
Problem with log config files.


sbernard31:
Currently we face several issues with log config files.

We currently use `logback-test.xml` for tests in almost all modules and `logback.xml` for executable modules.
The way logback load is described [here](https://logback.qos.ch/manual/configuration.html#auto_configuration). In our case it search in classpath for `logback-test.xml` then `logback.xml`.

**Problems in development :** (using eclipse)
1.1) eclipse does not do the difference between `test` classpath and  `runtime` classpath like maven. That's means that if you have both `logback-test.xml` and `logback.xml`, the test one will always be used even if you does not run tests in eclipse.
1.2) if there is several `logback.xml`  or several  `logback-test.xml` files in classpath there is no way to control which one will be used.

**Problems at build :** (using maven)
2) Separation between `test` and `runtime` works well with maven, but if you create an uber jar (jar with dependencies) and there is several logback.xml files, actually we do not control the one we are using.

**My ideas :**
To solve the 1.1) and 1.2) the only solution I see is to move log config file out of all classpath and define specifically which config file to use.
Logback allow to use a system properties to define which config file to use instead of searching in class path.
```
java -Dlogback.configurationFile=/path/to/config.xml chapters.configuration.MyApp1
```
Eclipse users could set it but I would like a solution which works out of the box.
A solution could be to
- move `./src/main/resource/logback.xml` to "./logback-config.xml" (so not in classpath)
- define a static block for all executable module which [define the default value for this properties](https://logback.qos.ch/manual/configuration.html#configFileProperty) using :
```java
static {
    // must be set before the first call to  LoggerFactory.getLogger();
    System.setProperty("logback.configurationFile", "logback-config.xml");
}  
```
If we want to also use this for Junit, we probably need to do move `logback-test.xml` out of the classpath too (as module root folder for example)  and to add this static block for all test classes...
About maven, `logback-test.xml` is no more in the classpath, so we need to add :
```xml
<artifactId>maven-surefire-plugin</artifactId>
<configuration>
<systemPropertyVariables>
	<logback.configurationFile>logback-test.xml</logback.configurationFile>
</systemPropertyVariables>
...
```
For uber-jar, with this method we have no more default config in jar but put a logback-config.xml file in working directory is enough to use it. (no configuration needed)
So we can :
a) add the "logback-config.xml" in output build directory (`target` directory) near the uber-jar, using :
```xml
<build>
<resources>
	<resource>
	<directory>${project.basedir}</directory>
[...]

PR #475, opened 2018-02-19
Fix log configuration issues

sbernard31:
This PR aims to implements solution express in #473.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@ bin/
 *.jar
 *.war
 *.ear
+
+# log Files #
+**/src/main/**/logback*.xml
+**/src/test/**/logback*.xml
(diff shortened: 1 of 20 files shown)


sbernard31:
After discussion with @msangoi, we finally decide to not add 
```
static {
     System.setProperty("logback.configurationFile", "logback-config.xml");
}
```
for Junit Tests.

We just add comments in `logback-test.xml` encouraging using `-Dlogback.configurationFile=logback-test-.xml` or just but manually only 1 `logback-test.xml` file in your classpath which will be ignore by git.

---------------------------------

The other change, I do not put anymore logback-config.xml in build output directory (target), because of maven-eclipse-plugin which will create an ant task and an external builder just because of that ...

PR #476, opened 2018-02-22
Update to californium 2.0.0 M8.

boaks:
Adjust for modified ObservationStore API.
Remove/replace Key by Token.


boaks:
Adjust the stores for the new ObservationStore API, I got aware, that `tokensByRegId` didn't check, if the token is already in the list. Is this intended?

boaks:
```
                if (!tokensByRegId.containsKey(registrationId)) {
                    tokensByRegId.put(registrationId, new ArrayList<Token>());
                }
                tokensByRegId.get(registrationId).add(token);

```

the add in the last line is always executed.

sbernard31:
You're right we should check it, or use a Set instead of a List.

boaks:
See #477

PR #481, opened 2018-03-05
Update the additionalRegistrationParameters when update messages are received

carlosgp143:
Hi all!

Here I come with a PR to update the `additionalRegistrationParameters` of the registration object when they are included in the update message. This is, among other applications, useful to use with Queue Mode to update the client awake time.

I have basically treated the additional parameters as they are treated regarding the Registration, so the code is very similar to the one present in those classes. Now the Map containing this parameters has changed from the type `Collections.unmodifiableMap` to `Collections.synchronizedMap`. Then, when new parameters arrive in the update message, using the method `putAll()` the already present attributes are updated with the new values, and the new ones are added.

Regarding the serialization used in the cluster, I'm not very familiar with this, but I have implemented the same code as in the serialization of the registration object, and it seems to pass correctly the JUnit test. But please check this in detail.

All feedback is welcomed :-)

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/request/UpdateRequest.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/request/UpdateRequest.java
@@ -15,6 +15,10 @@
  *******************************************************************************/
 package org.eclipse.leshan.core.request;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
 import org.eclipse.leshan.Link;
 import org.eclipse.leshan.core.request.exception.InvalidRequestException;
 import org.eclipse.leshan.core.response.UpdateResponse;
@@ -30,6 +34,7 @@ public class UpdateRequest implements UplinkRequest<UpdateResponse> {
     private final BindingMode bindingMode;
     private final String registrationId;
     private final Link[] objectLinks;
+    private final Map<String, String> additionalAttributes;
 
     /**
      * Sets all fields.
@@ -42,7 +47,7 @@ public class UpdateRequest implements UplinkRequest<UpdateResponse> {
      * @exception InvalidRequestException if the registrationId is empty.
      */
     public UpdateRequest(String registrationId, Long lifetime, String smsNumber, BindingMode binding,
-            Link[] objectLinks) throws InvalidRequestException {
+            Link[] objectLinks, Map<String, String> additionalAttributes) throws InvalidRequestException {
 
         if (registrationId == null || registrationId.isEmpty())
             throw new InvalidRequestException("registrationId is mandatory");
@@ -52,6 +57,10 @@ public class UpdateRequest implements UplinkRequest<UpdateResponse> {
         this.lifeTimeInSec = lifetime;
         this.bindingMode = binding;
         this.smsNumber = smsNumber;
+        if (additionalAttributes == null)
+            this.additionalAttributes = Collections.emptyMap();
+        else
+            this.additionalAttributes = Collections.unmodifiableMap(new HashMap<>(additionalAttributes));
     }
 
     public String getRegistrationId() {
@@ -72,6 +81,10 @@ public class UpdateRequest implements UplinkRequest<UpdateResponse> {
 
     public BindingMode getBindingMode() {
         return bindingMode;
+    }
+
+    public Map<String, String> getAdditionalAttributes() {
+        return additionalAttributes;
     }
 
     @Override
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
@@ -209,6 +209,8 @@ public class RegisterResource extends CoapResource {
         String smsNumber = null;
         BindingMode binding = null;
         Link[] objectLinks = null;
+        Map<String, String> additionalParams = new HashMap<>();
+
         for (String param : request.getOptions().getUriQuery()) {
             if (param.startsWith(QUERY_PARAM_LIFETIME)) {
                 lifetime = Long.valueOf(param.substring(3));
@@ -216,12 +218,18 @@ public class RegisterResource extends CoapResource {
                 smsNumber = param.substring(4);
             } else if (param.startsWith(QUERY_PARAM_BINDING_MODE)) {
                 binding = BindingMode.valueOf(param.substring(2));
+            } else {
+                String[] tokens = param.split("\\=");
+                if (tokens != null && tokens.length == 2) {
+                    additionalParams.put(tokens[0], tokens[1]);
+                }
             }
         }
         if (request.getPayload() != null && request.getPayload().length > 0) {
             objectLinks = Link.parse(request.getPayload());
         }
-        UpdateRequest updateRequest = new UpdateRequest(registrationId, lifetime, smsNumber, binding, objectLinks);
+        UpdateRequest updateRequest = new UpdateRequest(registrationId, lifetime, smsNumber, binding, objectLinks,
+                additionalParams);
 
         // Handle request
         final SendableResponse<UpdateResponse> sendableResponse = registrationHandler.update(sender, updateRequest);
--- a/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/RegistrationUpdateSerDes.java
+++ b/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/RegistrationUpdateSerDes.java
@@ -69,6 +69,14 @@ public class RegistrationUpdateSerDes {
             o.add("objLink", links);
         }
 
+        if (u.getAdditionalAttributes() != null) {
+            JsonObject addAttr = Json.object();
+            for (Map.Entry<String, String> e : u.getAdditionalAttributes().entrySet()) {
+                addAttr.add(e.getKey(), e.getValue());
+            }
+            o.add("addAttr", addAttr);
+        }
+
         return o;
     }
 
@@ -126,6 +134,12 @@ public class RegistrationUpdateSerDes {
             }
         }
 
-        return new RegistrationUpdate(regId, identity, lifetime, sms, b, linkObjs);
+        Map<String, String> addAttr = new HashMap<>();
+        JsonObject o = (JsonObject) v.get("addAttr");
+        for (String k : o.names()) {
+            addAttr.put(k, o.getString(k, ""));
+        }
+
+        return new RegistrationUpdate(regId, identity, lifetime, sms, b, linkObjs, addAttr);
     }
 }
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationUpdate.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationUpdate.java
@@ -18,7 +18,10 @@ package org.eclipse.leshan.server.registration;
 
 import java.net.InetAddress;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.eclipse.leshan.Link;
 import org.eclipse.leshan.core.request.BindingMode;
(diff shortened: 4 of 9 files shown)


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
```diff
@@ -209,19 +209,27 @@ private void handleUpdate(CoapExchange exchange, Request request, String registr
...
         for (String param : request.getOptions().getUriQuery()) {
             if (param.startsWith(QUERY_PARAM_LIFETIME)) {
                 lifetime = Long.valueOf(param.substring(3));
             } else if (param.startsWith(QUERY_PARAM_SMS)) {
                 smsNumber = param.substring(4);
             } else if (param.startsWith(QUERY_PARAM_BINDING_MODE)) {
                 binding = BindingMode.valueOf(param.substring(2));
+            } else {
+                String[] tokens = param.split("\\=");
+                if (tokens != null && tokens.length == 2) {
```
AlexITC: why is this limited to one parameter?

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/RegistrationUpdateSerDes.java
```diff
@@ -126,6 +134,12 @@ public static RegistrationUpdate deserialize(byte[] data) throws UnknownHostExce
             }
         }
 
-        return new RegistrationUpdate(regId, identity, lifetime, sms, b, linkObjs);
+        Map<String, String> addAttr = new HashMap<>();
+        JsonObject o = (JsonObject) v.get("addAttr");
+        for (String k : o.names()) {
+            addAttr.put(k, o.getString(k, ""));
```
AlexITC: you are not retrieving the serialized values, just the keys.

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationUpdate.java
```diff
@@ -111,6 +122,17 @@ public BindingMode getBindingMode() {
         return objectLinks;
     }
 
+    public Map<String, String> getAdditionalAttributes() {
+        return additionalAttributes;
+    }
+
+    Map<String, String> updateAdditionalAttributes(Map<String, String> oldAdditionalAttributes) {
```
AlexITC: I would prefer to specify the method visibility.

AlexITC:
Honestly, I don't see the reason to use a mutable map instead of the immutable one, as I remember, each time a request is send, all additional attributes are sent and you should be able to use the immutable map.

leshan-core/src/main/java/org/eclipse/leshan/core/request/RegisterRequest.java
```diff
@@ -70,7 +70,7 @@ public RegisterRequest(String endpointName, Long lifetime, String lwVersion, Bin
         if (additionalAttributes == null)
             this.additionalAttributes = Collections.emptyMap();
         else
-            this.additionalAttributes = Collections.unmodifiableMap(new HashMap<>(additionalAttributes));
```
sbernard31: `RegisterRequest` should be immutable.

leshan-core/src/main/java/org/eclipse/leshan/core/request/UpdateRequest.java
```diff
@@ -52,6 +57,10 @@ public UpdateRequest(String registrationId, Long lifetime, String smsNumber, Bin
         this.lifeTimeInSec = lifetime;
         this.bindingMode = binding;
         this.smsNumber = smsNumber;
+        if (additionalAttributes == null)
+            this.additionalAttributes = Collections.emptyMap();
+        else
+            this.additionalAttributes = Collections.synchronizedMap(new HashMap<>(additionalAttributes));
```
sbernard31: `UpdateRequest` should be immutable.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
sbernard31: That was the only "kind" of parameters supported for register request. 
If we want to support more than that this should be done in another PR.

@AlexITC which kind of additional Params would you like to support ?

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
```diff
@@ -113,9 +113,9 @@ protected Registration(String id, String endpoint, Identity identity, String lwM
         if (additionalRegistrationAttributes == null || additionalRegistrationAttributes.isEmpty()) {
             this.additionalRegistrationAttributes = Collections.emptyMap();
         } else {
-            // We create a new HashMap to have a real immutable map and to avoid "unmodifiableMap" encapsulation.
+
             this.additionalRegistrationAttributes = Collections
-                    .unmodifiableMap(new HashMap<>(additionalRegistrationAttributes));
```
sbernard31: `Registration` should be immutable.

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationUpdate.java
```diff
@@ -48,6 +52,10 @@ public RegistrationUpdate(String registrationId, Identity identity, Long lifeTim
         this.smsNumber = smsNumber;
         this.bindingMode = bindingMode;
         this.objectLinks = objectLinks;
+        if (additionalAttributes == null)
+            this.additionalAttributes = Collections.emptyMap();
+        else
+            this.additionalAttributes = Collections.synchronizedMap(new HashMap<>(additionalAttributes));
```
sbernard31: `RegistrationUpdate` should be immutable

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
carlosgp143: What do you mean one parameter? the for loop adds all the additional parameters that come in the update message.

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationUpdate.java
```diff
@@ -63,6 +71,10 @@ public Registration update(Registration registration) {
         BindingMode bindingMode = this.bindingMode != null ? this.bindingMode : registration.getBindingMode();
         String smsNumber = this.smsNumber != null ? this.smsNumber : registration.getSmsNumber();
 
+        Map<String, String> additionalAttributes = this.additionalAttributes != null
```
sbernard31: `additionalAttributes` can not be null.

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationUpdate.java
```diff
@@ -111,6 +122,17 @@ public BindingMode getBindingMode() {
         return objectLinks;
     }
 
+    public Map<String, String> getAdditionalAttributes() {
+        return additionalAttributes;
+    }
+
+    Map<String, String> updateAdditionalAttributes(Map<String, String> oldAdditionalAttributes) {
+
+        // putAll method updates already present key values or add them if not present.
```
sbernard31: create a new map to keep immutability contract

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/RegistrationUpdateSerDes.java
carlosgp143: Are you sure? I think it uses the method `put(key,value)` parsing the string to get the value. I copied this code from the `RegistrationSerDes`.

carlosgp143:
I'm not very familiar of how the Map classes work in Java, but when I was implementing it if I tried to apply the methods `put` or `putAll` to the unmodifiableMap it throwed an `UnsupportedOperationException`. So I think you cannot use this methods with this kind of Maps, and that's why I changed it to a synchronized map.

leshan-core/src/main/java/org/eclipse/leshan/core/request/UpdateRequest.java
carlosgp143: I don't get it... which kind of Map should I use?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
AlexITC: my mistake, I just saw the loop.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/RegistrationUpdateSerDes.java
AlexITC: you are right, thanks.

AlexITC:
the reason is that one an unmodifiable map has been created, you are not allowed to modify it, you can create an immutable map from a mutable map, something like this: `Collections.unmodifiableMap(mutableMap);`

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
sbernard31: I thought that you was talking about something like "parameter without value".
Total misunderstanding :laughing:  !

leshan-core/src/main/java/org/eclipse/leshan/core/request/UpdateRequest.java
sbernard31: Collections.unmodifiableMap(new HashMap<>(additionalAttributes));

leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationUpdate.java
sbernard31: An immutable object is an object which can not be modified after its creation.

Generally, in Java it means all attributes of this object are final and "non modifiable".
In Java, you can create unmodifiable "view" of a  collection from a modifiable collection using `Collections.unmodifiablexxx`.

How do you do when you need to modify this kind of  object : you create a new one.
This is a bit less efficient about memory consumption/GC activity but this is more safe about thread and concurrent access.

So something like you propose is OK.
We could even imagine something a bit smarter in case of empty collection but this is probably premature optimization.

carlosgp143: Ok! Thank you for your explanation :-)

carlosgp143:
I made a new commit with the changes that you pointed in the immutable Map. 

Anything else to be done in the PR? @sbernard31 you mentioned to add a test... but what kind? Only to test the updated parameters? Like create a registration object with some random parameters, update them with a registration update, and check if they are the ones expected?

sbernard31:
It sounds good.
Except `Registration` is now modified only about formatting issue. I would prefer to not change it just for that.

About tests, something like that yes.

If you are (or want to be) familiar with git, could you squash all commits in one, else I could quickly do that before to merge.

carlosgp143:
It's true, `Registration.java` does not need any changes in this PR. 

I can add some tests in this PR before integrating it if you want, and yes I'll squash all commits in one, I think I know how to do it :-)

sbernard31:
> I can add some tests in this PR before integrating it if you want

Adding tests was a suggestion, I think this could be useful to ensure your modification behave as expected and this could be a way to familiar yourself with that. But do it only if you think this makes sense :).

carlosgp143:
Yep, I think it totally does and it was very quick. So I think everything is done now :-)

carlosgp143:
Perfect!

What should be the next step in the Queue Mode implementation? Integration tests? Adding the queue for the requests?

sbernard31:
I would prefer to go with integration tests.

carlosgp143:
Sorry to discuss this here but I don't know where...

I have taken a look on the integration tests and regarding the client behavior how should we proceed? Now it sends an update message every lifetime-10%... which is not what it should do in Queue Mode, it should send an update message when it is awake. But tuning the Lifetime parameter I think we can have the expected behavior in the client. So my question is basically if I can do the tests like this or if I should make small changes in the client's implementation (basically in the `LeshanClient` and `RegistrationEngine` classes). I would go for the first approach to make it simpler and not spend too much time on this.

sbernard31:
I you want better control on update event you can increase life time for Queue mode test.
You can also use `LeshanClient.triggerRegistrationUpdate()`

PR #487, opened 2018-03-13
Queue Mode Integration Tests

carlosgp143:
In this PR I have implemented some integration tests for the Queue Mode. For it, I have created the class `QueueModeIntegrationTestHelper` which is similar to the already present `IntegrationTestHelper`, but creates a client that registers with binding mode UQ and also triggers an update according to its sleeping time (and not its lifetime). This is done with a timer that it's started when `onSleeping()` is called. The class also starts a server with the corresponding `StaticClientAwakeTimerProvider`.

Then, using this class I've created four different tests:

- Normal awake/sleeping behaviour
- Check that if an update message arrives when the client is already awake, only one notification is done. Here I realized that the awake timer was restarted on every update message before, without checking if the state has actually changed. I have updated this wrong behavior.
- Check that if a response timeout occurs, the client state changes to sleeping
- Check that the server can make requests when the client is awake.

All feedback on this implementation is welcomed, cause as I've already mentioned, I'm not really familiar with these integration tests. Also feel free to suggest more test, and even implement them.

### DIFF
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
@@ -1,0 +1,350 @@
+/*******************************************************************************
+ * Copyright (c) 2017 RISE SICS AB.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *     RISE SICS AB - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.leshan.integration.tests;
+
+import static org.junit.Assert.*;
+
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.util.Collection;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.eclipse.leshan.LwM2mId;
+import org.eclipse.leshan.client.californium.LeshanClient;
+import org.eclipse.leshan.client.californium.LeshanClientBuilder;
+import org.eclipse.leshan.client.object.Device;
+import org.eclipse.leshan.client.object.Security;
+import org.eclipse.leshan.client.object.Server;
+import org.eclipse.leshan.client.resource.LwM2mObjectEnabler;
+import org.eclipse.leshan.client.resource.ObjectsInitializer;
+import org.eclipse.leshan.core.model.LwM2mModel;
+import org.eclipse.leshan.core.model.ObjectLoader;
+import org.eclipse.leshan.core.model.ObjectModel;
+import org.eclipse.leshan.core.model.ResourceModel;
+import org.eclipse.leshan.core.model.ResourceModel.Operations;
+import org.eclipse.leshan.core.model.ResourceModel.Type;
+import org.eclipse.leshan.core.observation.Observation;
+import org.eclipse.leshan.core.request.BindingMode;
+import org.eclipse.leshan.core.response.ExecuteResponse;
+import org.eclipse.leshan.core.response.LwM2mResponse;
+import org.eclipse.leshan.server.californium.LeshanServerBuilder;
+import org.eclipse.leshan.server.californium.impl.LeshanServer;
+import org.eclipse.leshan.server.impl.InMemorySecurityStore;
+import org.eclipse.leshan.server.impl.RegistrationServiceImpl;
+import org.eclipse.leshan.server.model.StaticModelProvider;
+import org.eclipse.leshan.server.queue.PresenceListener;
+import org.eclipse.leshan.server.queue.StaticClientAwakeTimeProvider;
+import org.eclipse.leshan.server.registration.Registration;
+import org.eclipse.leshan.server.registration.RegistrationListener;
+import org.eclipse.leshan.server.registration.RegistrationUpdate;
+
+/**
+ * Helper for running a server and executing a client against it.
+ * 
+ */
+public class QueueModeIntegrationTestHelper {
+    public static final Random r = new Random();
+
+    static final String MODEL_NUMBER = "IT-TEST-123";
+    public static final long SLEEPTIME = 10; // seconds
+    public static final long AWAKETIME = 2; // seconds
+    public static final long LIFETIME = 3600; // Updates are manually triggered with a timer
+
+    public static final int TEST_OBJECT_ID = 2000;
+    public static final int STRING_RESOURCE_ID = 0;
+    public static final int BOOLEAN_RESOURCE_ID = 1;
+    public static final int INTEGER_RESOURCE_ID = 2;
+    public static final int FLOAT_RESOURCE_ID = 3;
+    public static final int TIME_RESOURCE_ID = 4;
+    public static final int OPAQUE_RESOURCE_ID = 5;
+    public static final int OBJLNK_MULTI_INSTANCE_RESOURCE_ID = 6;
+    public static final int OBJLNK_SINGLE_INSTANCE_RESOURCE_ID = 7;
+
+    LeshanServer server;
+
+    LeshanClient client;
+    AtomicReference<String> currentEndpointIdentifier = new AtomicReference<String>();
+
+    CountDownLatch registerLatch;
+    Registration last_registration;
+    CountDownLatch deregisterLatch;
+    CountDownLatch updateLatch;
+    CountDownLatch awakeLatch;
+
+    private final ScheduledExecutorService sleepTimerExecutor = Executors.newSingleThreadScheduledExecutor();
+    private ScheduledFuture<?> sleepScheduledFuture;
+
+    int awakeNotifications = 0;
+
+    protected List<ObjectModel> createObjectModels() {
+        // load default object from the spec
+        List<ObjectModel> objectModels = ObjectLoader.loadDefault();
+        // define custom model for testing purpose
+        ResourceModel stringfield = new ResourceModel(STRING_RESOURCE_ID, "stringres", Operations.RW, false, false,
+                Type.STRING, null, null, null);
+        ResourceModel booleanfield = new ResourceModel(BOOLEAN_RESOURCE_ID, "booleanres", Operations.RW, false, false,
+                Type.BOOLEAN, null, null, null);
+        ResourceModel integerfield = new ResourceModel(INTEGER_RESOURCE_ID, "integerres", Operations.RW, false, false,
+                Type.INTEGER, null, null, null);
+        ResourceModel floatfield = new ResourceModel(FLOAT_RESOURCE_ID, "floatres", Operations.RW, false, false,
+                Type.FLOAT, null, null, null);
+        ResourceModel timefield = new ResourceModel(TIME_RESOURCE_ID, "timeres", Operations.RW, false, false, Type.TIME,
+                null, null, null);
+        ResourceModel opaquefield = new ResourceModel(OPAQUE_RESOURCE_ID, "opaque", Operations.RW, false, false,
+                Type.OPAQUE, null, null, null);
+        ResourceModel objlnkfield = new ResourceModel(OBJLNK_MULTI_INSTANCE_RESOURCE_ID, "objlnk", Operations.RW, true,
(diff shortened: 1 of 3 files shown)


leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
```diff
@@ -0,0 +1,350 @@
...
+                    awakeNotifications++;
+                }
+
+            }
+
+            @Override
+            public void onSleeping(Registration registration) {
+                awakeNotifications = 0;
+                awakeLatch.countDown();
+                scheduleUpdateAfterSleeping((int) SLEEPTIME);
```
sbernard31: I'm not sure this is really needed.
You can use directly triggerUpdate in tests? (this is maybe a way to speed up the tests)

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
```diff
@@ -0,0 +1,200 @@
...
+        queueModeHelper.client.start();
+        queueModeHelper.waitForRegistration(1);
+
+        // Check client is well registered and awake
+        queueModeHelper.assertClientRegisterered();
+        queueModeHelper.ensureClientAwake();
+
+        // Send a response with very short timeout so that exception is thrown
+        ReadResponse response = queueModeHelper.server.send(queueModeHelper.getCurrentRegistration(),
+                new ReadRequest(3, 0, 1), 1);
```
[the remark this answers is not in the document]
carlosgp143: Should we increase it then? What value is more suitable in your computer?

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
```diff
@@ -0,0 +1,350 @@
...
+/**
+ * Helper for running a server and executing a client against it.
+ * 
+ */
+public class QueueModeIntegrationTestHelper {
+    public static final Random r = new Random();
+
+    static final String MODEL_NUMBER = "IT-TEST-123";
+    public static final long SLEEPTIME = 10; // seconds
+    public static final long AWAKETIME = 2; // seconds
```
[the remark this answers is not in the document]
carlosgp143: Yes, sure. It doesn't really matter. Maybe 2 seconds sleep and 1 second awake or something like that?

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
carlosgp143: I was trying to make a more realistic tests, where the client actually "sleep" for some time. But yeah, I guess we can also trigger the update message, simulating a "0 sleep time", that will test the server functionality. Would you prefer that? 

We can also maintain this way, and the user can set the sleeptime to 0 if it wants this behavior. I would prefer this option.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
```diff
@@ -0,0 +1,350 @@
...
+import org.eclipse.leshan.server.queue.StaticClientAwakeTimeProvider;
+import org.eclipse.leshan.server.registration.Registration;
+import org.eclipse.leshan.server.registration.RegistrationListener;
+import org.eclipse.leshan.server.registration.RegistrationUpdate;
+
+/**
+ * Helper for running a server and executing a client against it.
+ * 
+ */
+public class QueueModeIntegrationTestHelper {
```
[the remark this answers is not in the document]
carlosgp143: Done in new commit :-)

carlosgp143:
Hi! This PR is a bit dead... so let's try to push it forward again :-)

I made a new commit using inheritance in the `QueueModeIntegrationTestHelper` and also reducing the sleep and awake times, with the possibility to set them even to 0 to reduce the time a lot. 

Anything else to be done?

sbernard31:
Sry, I cumulated some delay on Leshan. I will try to look at this ASAP. Same for  #484.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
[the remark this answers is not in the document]
sbernard31: If I well understand you choose a very small timeout. In fact the smaller one : 1 ms.
Hoping that the request will timeout before we get the response.

My computer seems too fast :p because this tests fail almost all the time, so I think we should find another strategy.

Maybe stoping the client without deregistration ?

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
sbernard31: It works better

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
```diff
@@ -0,0 +1,179 @@
...
+                scheduleUpdateAfterSleeping((int) sleeptime);
+            }
+
+        });
+        // monitor client registration
+        setupRegistrationMonitoring();
+    }
+
+    protected LeshanServerBuilder createServerBuilder(int clientAwakeTime) {
+        LeshanServerBuilder builder = new LeshanServerBuilder();
```
sbernard31: You could use super.createServerBuilder() to avoid some duplicate code.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
```diff
@@ -0,0 +1,213 @@
...
+import org.junit.Test;
+
+public class QueueModeTest {
+
+    protected QueueModeIntegrationTestHelper queueModeHelper = new QueueModeIntegrationTestHelper();
+
+    @Before
+    public void start() {
+        queueModeHelper.initialize();
+        queueModeHelper.createServer((int) queueModeHelper.awaketime * 1000);
```
sbernard31: I would store awaketime as attribute of this class and remove if from queueModelHelper

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
```diff
@@ -0,0 +1,213 @@
...
+        // Start it and wait for registration
+        queueModeHelper.client.start();
+        queueModeHelper.waitForRegistration(1);
+
+        // Check client is well registered
+        queueModeHelper.assertClientRegisterered();
+        assertArrayEquals(Link.parse("</>;rt=\"oma.lwm2m\",</1/0>,</2>,</3/0>,</2000/0>".getBytes()),
+                queueModeHelper.getCurrentRegistration().getObjectLinks());
+
+        // Check that client is awake and only one notification
```
sbernard31: > and only one notification

This comment seems to be not related to the code.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
```diff
@@ -0,0 +1,213 @@
...
+        queueModeHelper.ensureClientAwake();
+
+        // Wait for client awake time expiration (1% margin)
+        queueModeHelper.waitForAwakeTime(queueModeHelper.awaketime * 1010);
+
+        // Check that client is sleeping
+        queueModeHelper.ensureClientSleeping();
+
+        // Wait for update when waking up (1% margin)
```
sbernard31: I feel this could be clearer, if we trigger the update manually instead of the waitForUpdate.
This need to be aware and understand the schedule stuff in QueueModelHelper

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
```diff
@@ -0,0 +1,213 @@
...
+        if (queueModeHelper.sleeptime != 0) {
+            queueModeHelper.waitForUpdate(queueModeHelper.sleeptime * 1010);
```
sbernard31: /!\ WaitForUpdate timeout is in seconds not milliseconds

Check the code, it seems you are using a lot update timeout >= 1000 ?

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Helper for running a server and executing a client against it.
+ * 
+ */
+public class QueueModeIntegrationTestHelper extends IntegrationTestHelper {
+    public final long sleeptime = 0; // seconds
+    public final long awaketime = 1; // seconds
+    public static final long LIFETIME = 3600; // Updates are manually triggered with a timer
+    CountDownLatch awakeLatch;
```
sbernard31: Only if there are a good reason attribute should be private and final.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private final ScheduledExecutorService sleepTimerExecutor = Executors.newSingleThreadScheduledExecutor();
+    private ScheduledFuture<?> sleepScheduledFuture;
+
+    int awakeNotifications = 0;
```
sbernard31: As this attribute is access by several threads in not synchronized block it should be an AtomicInteger.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeIntegrationTestHelper.java
carlosgp143: In `IntegrationTestHelper` the latches are not private and final... that's why I kept it like this. But I can change it :-)

carlosgp143: I think it cannot be final as it is assigned again in `resetLatch()` and `resetLatch()`. Also, in `IntegrationTestHelper` the latches are not private and final...

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
carlosgp143: Ok, I'll change it and directly call `queueModeHelper.client.triggerRegistrationUpdate();` for waking up.

carlosgp143:
Hi! I corrected all the feedback from your previous review and also removed the scheduler for the sleep time and directly trigger the update, to make it simpler :-)

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+        // Wait for client awake time expiration (1% margin)
+        queueModeHelper.waitForAwakeTime(awaketime * 1010);
+
+        // Check that client is sleeping
+        queueModeHelper.ensureClientSleeping();
+
+        // Trigger update manually for waking up
+        queueModeHelper.client.triggerRegistrationUpdate();
+        queueModeHelper.waitForUpdate(1);
```
sbernard31: Should we reset latch here ?

carlosgp143: True, sorry I missed it

sbernard31:
We finally did it ! This is integrated in master (commit c8e15385d94266e106cd928b3425cc7fc4115a72)
Sry again for the delay :wink: 

I fixed the formatting of `QueueModeTest` (tab instead of spaces), you should check your formatter configuration.

carlosgp143:
Perfect! Thanks for your help.

So next step... add the queues to store requests when the client is sleeping?

sbernard31:
Right.
I'm still not sure of the interest of this.
You don't change your mind about it ?

If you still think this is worth it. Which kind of design do you have in mind ?
I see it as a kind of helper class on top on the current Leshan API.

carlosgp143:
I think it can be interesting for some use cases, like the one I'm using Q-Mode for. I agree that people can just put the Queues on top of this implementation using the listeners, but I think it is a better option to have them directly inside the core... and people do not need to worry about it. Of course, always keeping the possibility of enable/disable this feature.

In my current design I have the queues inside the `PresenceStatus` class, and the `QueueModeRequestSender` put the request in them if the state is sleeping. Then, `PresenceService` sends all the queued requests when the state changes to awake. The only problem is that this creates a double dependency between  `QueueModeRequestSender` and `PresenceService`. So maybe a better option is to have another object on top of them, directly inside the `LeshanServer` class that manages the queues...

sbernard31:
I see that even out side of LeshanServer, I mean on top of this.

PR #493, opened 2018-03-28
Expose the certificate chain in the sender Identity

msangoi:


AlexITC:
I feel that I have already merged something like this.

AlexITC:
Found it, it seems that I forget to upload the pull request ([commit](https://github.com/artikcloud/leshan/commit/b309dad5eed67f12bbdcf0a8e1a6a4b82ab2bf3c)).

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
```diff
@@ -43,10 +42,11 @@ public static Identity extractIdentity(EndpointContext context) {
...
                 PublicKey publicKey = ((RawPublicKeyIdentity) senderIdentity).getKey();
                 return Identity.rpk(peerAddress, publicKey);
-            } else if (senderIdentity instanceof X500Principal || senderIdentity instanceof X509CertPath) {
-                // Extract common name
-                String x509CommonName = extractCN(senderIdentity.getName());
+            } else if (senderIdentity instanceof X500Principal) {
+                String x509CommonName = X509Util.extractCN(senderIdentity.getName());
                 return Identity.x509(peerAddress, x509CommonName);
+            } else if (senderIdentity instanceof X509CertPath) {
+                return Identity.x509(peerAddress, ((X509CertPath) senderIdentity).getPath());
```
AlexITC: isn't useful to have the common name and the certificate chain when possible?

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/IdentitySerDes.java
```diff
@@ -52,6 +52,7 @@ public static JsonObject serialize(Identity identity) {
             o.set(KEY_RPK, Hex.encodeHexString(publicKey.getEncoded()));
         } else if (identity.isX509()) {
             o.set(KEY_CN, identity.getX509CommonName());
+            // we don't store the certificate chain if it is available in the identity
```
AlexITC: why not?

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
```diff
@@ -67,28 +67,15 @@ public static EndpointContext extractContext(Identity identity) {
             } else if (identity.isRPK()) {
                 peerIdentity = new RawPublicKeyIdentity(identity.getRawPublicKey());
             } else if (identity.isX509()) {
-                /* simplify distinguished name to CN= part */
-                peerIdentity = new X500Principal("CN=" + identity.getX509CommonName());
+                if (identity.getCertificates() != null) {
+                    peerIdentity = new X509CertPath(identity.getCertificates());
+                } else {
+                    /* simplify distinguished name to CN= part */
+                    peerIdentity = new X500Principal("CN=" + identity.getX509CommonName());
```
AlexITC: I would suggest to create a separate method in `X509Util` to create the CN part.

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
msangoi: The common name is extracted from the first certificate of the chain in the Identity class.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/IdentitySerDes.java
msangoi: I guess the certificate chain is needed only when the session is established, in order to have a custom verification of the client certificate. Once it is verified, it is probably not needed anymore and storing it is quite costly in terms of payload size.
Do you have a use case for storing it?

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
msangoi: Done.

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/serialization/IdentitySerDes.java
AlexITC: When I added the changes, I was serializing the certificate chain, honestly, I can't remember why.

I prefer to leave this question to @boaks or @sbernard31

boaks: When I remember right, the idea was to move the turst-validation of the cert-chain out of californium/scandium to make it more flexible, e.g. use a tenant provided CA. Because scandium is currently too limited in the credential management, that's not possible in DTLS. For TLS using a more sophisticated `TrustManager` or `X509ExtendedKeyManager` should be the better approach.

I'm not really sure, where @AlexITC want to add the code for the tenant aware CA cert-chain validation. I guess, when this information is provided, we see, if this is a good idea or if there are better ones.

AlexITC: Yes, I implemented that using a custom `Authorizer`, as I'm not really familiar with Leshan Cluster, I can't say if it is ok to not serialize the certificate chain if we depend on it.

My understanding is that even if we don't serialize it, as the information comes from every request, that should be ok.

sbernard31: The needs are to make the trust-validation more flexible.

In this PR, we go for moving the trust-validation out of Scandium to move it in Leshan `Authorizer`.

But after reflexion, I think there is some issue doing like this. `Authorizer` is only called on registration/update/deregistration, then the identity stored in registration is used to ensure we talk to the right device.

If we only store the CN, a malicious device which do a new handshake (without registration but with same address/port) using another certchain but with the same CN could usurp device identity. (It should be doable with a man in the middle attack)

Storing the certain chain could be a solution, but this will increase a lot the RegistrationStore size and this does solve the case where the certchain is invalidate via OCSP or CRL.

So maybe a better solution is to move this flexibility in scandium at Handshake level ?
I already experiment different way to add flexibility to scandium :
https://github.com/eclipse/californium/pull/533
https://github.com/eclipse/californium/pull/536
But those do not really aim this need so we should probably take this point into account too.

boaks: FMPOV:
We need to validate the trust either on handshake, or at least once after that.
One solution for an upper layer would be, to create a validated cache for such credentials and validate all receiving messages (not only the messages for the registration interface) against this cache, if it's already in, OK - trust, if not, try to validate and if trusted, add it to the cache on success.

@AlexITC 
Did you try to use the java `TrustManager` or `X509ExtendedTrustManager` to implement your tenant aware trusting? If that was successful, we may consider to change scandium accordingly.

msangoi: From this discussion, it seems that being able to introduce some custom validation logic in scandium would be a better solution. It would be properly managed by the DTLS layer for all handshakes, with or without a registration.
Here is a proposal: https://github.com/eclipse/californium/pull/602
Let me know what you think.

AlexITC: @boaks I used none of them but a custom microservice focused on that part.

boaks: > custom microservice

Smells like something potentially takes more time then the trust API assumes :-).

From my experience, the current credentials/trust APIs are intended to be used in a sync and Immediate way; they are not intended to delegated to a asynchronous, remote and delayed micro service. Such delays will tend to block the "limited number of threads" used by the systems. This may change, but I'm not aware, that such other solutions are ready.

Therefore the idea to just finish the handshake and delegate the trust verification to the next layer will enable that layer to decouple the "trust verification" from the threads in a asynchronous manner.

@msangoi 
Your scandium PR is very welcome, but, (as I suspect for my self and therefore asked about the usage of the TrustManager), will raise follow-up issues.

AlexITC: you are right, sadly, that's the only way we could accomplish our goals with Leshan, until now, we haven't had issues due to these blocking calls.

msangoi:
I close this PR because it is now possible to introduce a custom certificate validation through the DtlsConnectorConfig (see https://github.com/eclipse/californium/pull/602).

PR #512, opened 2018-05-30
Better handling of null value in SecurityCheck

### LINKED ISSUE #443: SecurityCheck#checkRpkIdentity could throw NullPointerException
The [SecurityCheck#checkRpkIdentity](https://github.com/eclipse/leshan/blob/master/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityCheck.java#L126) method could throw a `NullPointerException` when `publicKey` is null and debug log level is enabled.

sbernard31:
It should fix #443


sbernard31:
Integrated in master (commit d4639b81aa6b6fa6b792fa424dc9e72db2c0057c)

Issue #464, opened 2018-02-02
Question about RegistrationStore#getRegistrationByAdress


AlexITC:
As I understand, the `Identity` object was added to the `Registration` in order to differentiate clients that are connecting by the same IP address and port.

If this understanding is correct, shouldn't this method return a list instead of a single value?

Thanks.

sbernard31:
Good question! 

`RegistrationByAddress` is used when LWM2M Server will try to initiate a handshake using PSK. This is the case where the LWM2M server act as DTLS Client => this is strange use case but this could happen.

What happen exactly, when you send data via `DtlsConnector` to a peer for which you have not DTLS session? The `DtlsConnector` will try to initiate an Handshake. If you are a LWM2M server and you are using PSK you should guess which PSK to use to establish a handshake with the device. To do that the API of DtlsConnector will ask for the psk to use for the peer with this address. (see `LwM2mPskStore.getIdentity(InetSocketAddress)`)

Currently, Leshan server search in RegistrationStore to find which registration is used for this address and so find PSK to use.
But you're right, there are cases where you could have 2 registration with the same address. (Devices with dynamic IP e.g. in NAT environment + QueueMode). 
This does not means that there is 2 devices with the same IP (this is a use case we don't want to support)
This means that 1 device uses the old IP to another device which are sleeping(QueueMode) .

Currently our default `RegistrationStore` implementation just take the first registration found. **This is an issue.**

I think about 3 way to solve this issue without Californium modification : 
1. When a device register itself or update its registration. If IP is already in use, remove peer address from old registration. (So we will have Registration without ip address :/ )
2. RegistrationStore#getByAddress should return the more recent one. (using last update ?)
3. RegistrationStore#getByAddress should return a list and this is LwM2mPskStore which is responsible to take the more recent one (still using last update).

I don't know what is the best solution 2. or 3. seems better to me. 
2. can already be implemented (we should add some javadoc for implementers). 
3. need API modifications (but maybe a more simple javadoc)
2. is maybe more easy to implement with index because we just need to keep and index to the last registration which use this IP.

I hope my comment is almost understandable.

AlexITC:
Interesting, personally I would go with the 3rd point because that give the client (who uses the APi) to make the choice and we should not return too many results anyway. As you said 3 needs API modifications.

Thanks.

sbernard31:
What do you mean by "that give the client (who uses the APi) to make the choice" ?

AlexITC:
[...]

PR #517, opened 2018-05-31
Fix getRegistrationByAdress (use index and handle address conflict)

sbernard31:
See #464 for more details.

### DIFF
--- a/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
+++ b/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
@@ -70,8 +70,9 @@ public class RedisRegistrationStore implements CaliforniumRegistrationStore, Sta
     private static final Logger LOG = LoggerFactory.getLogger(RedisRegistrationStore.class);
 
     // Redis key prefixes
-    private static final String REG_EP = "REG:EP:";
-    private static final String REG_EP_REGID_IDX = "EP:REGID:"; // secondary index key (registration)
+    private static final String REG_EP = "REG:EP:"; // (Endpoint => Registration)
+    private static final String REG_EP_REGID_IDX = "EP:REGID:"; // secondary index key (Registration ID => Endpoint)
+    private static final String REG_EP_ADDR_IDX = "EP:ADDR:"; // secondary index key (Socket Address => Endpoint)
     private static final String LOCK_EP = "LOCK:EP:";
     private static final byte[] OBS_TKN = "OBS:TKN:".getBytes(UTF_8);
     private static final String OBS_TKNS_REGID_IDX = "TKNS:REGID:"; // secondary index (token list by registration)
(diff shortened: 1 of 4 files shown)


AlexITC:
Based on the question, what I see is that we are discarding registrations having the same ip address, shouldn't this cause problems with clients behind the same NAT?

sbernard31:
What do you mean "we are discarding registrations having the same ip address" ? 

Registration was not discarded but the index (addr => registration) keeps the most recent registration.

AlexITC:
From what I see this doesn't not resolve the actual question, in my opinion, the ideal scenario would be to return a list of registrations instead of a single one.

sbernard31:
Ok, let me resume that, in a leshan server context : 

[PskStore.getIdentity(InetSocketAddress inetAddress)](https://github.com/eclipse/californium/blob/2.0.x/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/PskStore.java#L76) is used when the server want to initiate a DTLS connection. This should return only 1 identity, this make sense as we want to talk to 1 specific device/client.

In Leshan the way to know the psk identity, is to search in RegistrationStore to find which registration is associated to this socket address. 

Theoretical there is only 1 device for 1 address, but with NAT a device D1 could get an address A then go to sleep, then a new device D2 will get this free address A. So in the store 2 devices with the same address A. But in fact there is only 1 device behind the A address :D2. D1 lost this address and we don't know its new address, we will know that next time it will update its registration.

So this makes senses to me to only keep tracks to the device which use the address A lastly.

> From what I see this doesn't not resolve the actual question,

I'm not sure to understand what is the question ? :)

>  the ideal scenario would be to return a list of registrations instead of a single one.

I do not understand what does it bring more, as pskStore must choose one of them to choose 1 identity.

AlexITC:
Sounds reasonable, thanks!

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
```diff
@@ -139,15 +140,20 @@ public Deregistration addRegistration(Registration registration) {
...
+                // add registration: secondary indexes
+                byte[] regid_idx = toRegIdKey(registration.getId());
+                j.set(regid_idx, registration.getEndpoint().getBytes(UTF_8));
+                byte[] addr_idx = toRegAddrKey(registration.getSocketAddress());
+                j.set(addr_idx, registration.getEndpoint().getBytes(UTF_8));
 
                 if (old != null) {
                     Registration oldRegistration = deserializeReg(old);
                     // remove old secondary index
                     if (registration.getId() != oldRegistration.getId())
```
msangoi: Not related to this PR but this string equality looks buggy...

sbernard31: I will fix this in a separated commit

Issue #485, opened 2018-03-07
Add an easy way to send CoAP request from Leshan implementation.


sbernard31:
The idea is to reuse information about client registration at server side or server data in object model at client side to make easier to send CoAP request.

sbernard31:
This is done for server in #518.

PR #518, opened 2018-06-06
Add a new API to send CoAP request to a registered LWM2M client

sbernard31:
This aims to implements #485 for server side only.


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
```diff
@@ -343,6 +346,149 @@ public InetSocketAddress getSecuredAddress() {
         }
     }
 
+    /**
+     * A CoAP API, generally needed when you want to mixed LWM2M and CoAP protocol.
```
msangoi: typo "mix" / "procotolS"

PR #519, opened 2018-06-06
Enhance registration expiration of RedisRegistrationStore

### LINKED ISSUE #508: Performance issue about registration expiration of RedisRegistrationStore
In the RedisRegistrationStore.java file of cluster project,I find the Cleaner class.
In my test,while this running,the leshan server will become slower and slower,and sone new request  processed timeout.
Because there are 200000 registrations in the redis,but it only can read 50000-100000 registrations per second from redis.
I don't think it's a good idea to read all regionstrations for cleaning,do you thik so?
Maybe we can set the  expire time to every redis key?

sbernard31:
This aims to fix #508.

### DIFF
--- a/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
+++ b/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
@@ -29,6 +29,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
```diff
@@ -273,6 +273,14 @@ public Date getLastUpdate() {
         return lastUpdate;
     }
 
+    public long getExpirationTimeStamp() {
+        return lastUpdate.getTime() + lifeTimeInSec * 1000;
```
msangoi: return getExpirationTimeStamp(0L)

leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java
```diff
@@ -690,21 +711,17 @@ public void stop() {
...
-                    ScanResult<byte[]> res = j.scan(cursor.getBytes(), params);
-                    for (byte[] key : res.getResult()) {
-                        Registration r = deserializeReg(j.get(key));
-                        if (!r.isAlive(gracePeriod)) {
-                            Deregistration dereg = removeRegistration(j, r.getId(), true);
-                            if (dereg != null)
-                                expirationListener.registrationExpired(dereg.getRegistration(),
-                                        dereg.getObservations());
-                        }
+                Set<byte[]> endpoints_expired = j.zrangeByScore(EXP_EP, Double.NEGATIVE_INFINITY,
```
msangoi: snake case? 😮

PR #521, opened 2018-06-08
#503: Fix RequestSender to cancel blockwise request on deregistration

### LINKED ISSUE #503: Ongoing request is not canceled on deregistration
Looking into the code, there doesn't seem to be any exposed way to cancel an ongoing request?

Our scenario is that we're transferring a FOTA image, which may take several minutes, and sometimes we get a deregistration from the device during this, but the write request continues running in the background, eventually resulting in a success or timeout.

We would like to be able to cancel the ongoing request as soon as the deregistration occurs.

Thx!

sbernard31:
This aims to fix #503


sbernard31:
Integrated in master (commit a45b0932a3575eb0de10ee1e03b5ccaa3a478366)

PR #529, opened 2018-07-06
DeleteRequest should target object instance only.

rkimsb2:
And for DiscoverRequest, there's nothing different with current behavior but just fixed it for patterns of DownlinkRequest classes.

Signed-off-by: Rokwoon Kim <k862390@gmail.com>


sbernard31:
I integrated your PR in master (commit a7c958ab2dcad79db763e2d1fb9632604bcb30fe)
I fixed a test which didn't passed anymore with this fix but I ask myself if I should not just delete it...
Anyway thx again @rkimsb2 :+1:

Issue #435, opened 2017-11-15
Add RPK support to Leshan Client.


sbernard31:
Security parameters can change dynamically at bootstrap, so several solutions : 
1. add to scandium a kind of interface like PskStore  for RPK.
2. restart create a new DTLS connector if security parameters change.

sbernard31:
#546 implements this feature and is now integrated in master.

PR #538, opened 2018-07-16
Add Support of RPK at Client side

sbernard31:
This aims to implements #435.
This PR need the https://github.com/eclipse/californium/pull/536.

To use it with this PR.
```bash
# get californium credential_store branch
git clone git@github.com:eclipse/californium.git
cd /in/your/californium/directory
git checkout credential_store

# change version 
mvn versions:set -DnewVersion=2.0.0-credStore  

# install it
mvn clean install -DskipTests

# if you are using eclipse
cd /in/your/leshan/directory
mvn eclipse:eclipse
```


sbernard31:
@rkimsb2 you should be interested in this.

sbernard31:
It seems this is not the right way if we want to support co-located server. (see https://github.com/eclipse/californium/pull/536 for more details)
I don't know if this make sense to try to support this use case, but I will finally go back to the #450 way.

I don't close yet this PR, so if you have any feedback/thought about this you're welcome !

sbernard31:
Deprecated by #546

PR #546, opened 2018-07-27
Add Support of RPK at Client side 2

sbernard31:
This aims to implements #435.

This is based on #450 and so one socket/endpoint by server.

This is an alternative to #538.


sbernard31:
integrated in master.

PR #558, opened 2018-08-23
Cancellation of requests

danielhqv:
We have a problem with observation listeners that linger after a client has unregistered. Would the proposed change fix this, the comment directly above indicates that the intention was to also remove observations?

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/LeshanServer.java
@@ -166,6 +166,7 @@ public class LeshanServer implements LwM2mServer {
             public void unregistered(Registration registration, Collection<Observation> observations, boolean expired,
                     Registration newReg) {
                 requestSender.cancelPendingRequests(registration);
+                observationService.cancelObservations(registration);
             }
 
             @Override


sbernard31:
@danielhqv could you describe more precisely your problem ? Because at first sight, I feel that we don't really need to call `observationService.cancelObservations(registration)` in `unregistered`

The idea is when deregistration is done. RegistrationStore.removeRegistration() is called. This will remove registration and observation associated in an atomic way. The unregister event (from RegistrationListener) return the observation canceled at deregistration.

danielhqv:
It's just that I cannot find any code that removes observations from the ObservationServiceImpl, or the RegistrationStore for that matter. No cancel... methods related to observations ever seems to be called, except from some test classes and the server demo ClientServlet.

But I might have misunderstood how it works.

sbernard31:
Observations should be removed in RegistrationStore.

https://github.com/eclipse/leshan/blob/master/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/InMemoryRegistrationStore.java#L187

https://github.com/eclipse/leshan/blob/master/leshan-server-cluster/src/main/java/org/eclipse/leshan/server/cluster/RedisRegistrationStore.java#L337

This works because RegistrationStore implements `org.eclipse.californium.core.observe.ObservationStore`

sbernard31:
@danielhqv  should we close this PR ?

bolddp:
Yes, I think that's OK

Issue #597, opened 2018-10-17
Custom object creation


Yalayak:
Hi everyone, 
I'm a beginner so please bare with me. 
I would like to create a new custom object in Leshan. I would like to know if it is obligatory to only use the existing IPSO data models. Is it possible for me to create my own model and implement it? How can I make the server able to deal with the new objects/ressources?

sbernard31:
Hi, 
  You didn't search too much as less ;)  
  1 month ago, we get exactly the same question : https://github.com/eclipse/leshan/issues/570.

  But the good point is that your question motivate me to create documentation about that.
  See https://github.com/eclipse/leshan/wiki/Adding-new-objects.

  I also add this to the [F.A.Q.](https://github.com/eclipse/leshan/wiki/F.A.Q.#how-to-add-new-objects-).

  I hope this help.

(Please close this issue, if you get enough information)

Yalayak:
thank you very much for the help, I would like to know one more thing and excuse me if my question is dumb, how can I set different names for the different instances, meaning the names that will show up on the server? Thanks!

sbernard31:
In LWM2M, there is no name for instances. Instances is identified by number.

And so for `leshan-server-demo` there is no name to display.

If you create your own object, you can imagine that one of its resources is the name. And so a server could read this resource to get the name of the instance. But this will be only for your custom objects.

Yalayak:
Is there a way to associate a particular number identifier for a certain instance?

sbernard31:
If when the server create a new instance using CREATE request you can chose the Id.
If the server don't give any id, the client will generate it automatically. (currently you can not chose it)

At initialization, you can not chose it too. :disappointed:

Could you describe your use case, and we could think together how to modify the API to make it possible if needed?

Yalayak:
Ok thank you, 
So in my case I have some devices that are connected to a gateway and that I want to configure with Leshan. 
For each object I would like to have multiple instances, an instance for every device. I would like to provide each instance with a unique ID number so that in the server I can identify which instance is that of a certain device without adding a resource for this type of identification.

sbernard31:
How do you create this object instance using a `LwM2mInstanceEnablerFactory` ?

Yalayak:
Looking at the code of `LwM2mInstanceEnablerFactory` I thought I can't pass two different classes for instantiating each instance of an object. But it works apparently, I use this: 

`initializer.setInstancesForObject(OBJECT_ID, new class1(), new class2());`

[...]

PR #616, opened 2018-11-13
Add a way to make "atomic" operation (read/write) on LwM2mInstanceEnabler

sbernard31:
Experimental work to try to support https://github.com/eclipse/leshan/issues/597#issuecomment-437888248

>At the client side I need to read all resources of a particular instance and pass it to another layer. Now since resources ids are called one by one in switch case (write/read), its not possible to send complete object at once. Thats the main reason I wanted to read complete instance at once and pass it to other layer.

The new API in LwM2mInstanceEnabler

```java
/**
* Gets values of all readable resources of this instance.
* 
* @param identity the identity of the requester. This could be an internal call in this case
*        <code> identity == ServerIdentity.SYSTEM</code>.
* 
* @return a success response with an {@link LwM2mObjectInstance} as content or a failure response with optional
*         explanation message.
*/
ReadResponse read(ServerIdentity identity);
```

### DIFF
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Device.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Device.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.TimeZone;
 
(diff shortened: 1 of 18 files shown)


sbernard31:
I added the write and observe operation.

Issue #535, opened 2018-07-13
Support of object versioning at server side.


sbernard31:
The support of object version was added by #438, but to be really useful at server side we need a kind of `ObjectModelRepository` which could contains several version of the same object.

sbernard31:
This is done with #657 which is now integrated in master.

PR #657, opened 2019-02-25
Add object versioning support at server side

sbernard31:
It aims to implement  #535.


sbernard31:
If you want exactly the same behavior as before, you need to use a StaticModelProvider : 
```java
builder.setObjectModelProvider(new StaticModelProvider(ObjectLoader.loadDefault()));
```

PR #659, opened 2019-03-04
SenML JSON encoder in leshan-core with unit tests.

cavenaghi9:
Contributing SenML JSON. Implementing SenML-JSON encoder in leshan-core inspired by LwM2mJson, with unit tests.


sbernard31:
1. I read the [RFC 8428](https://tools.ietf.org/html/rfc8428) and I think a more appropriate [terminology](https://tools.ietf.org/html/rfc8428#section-3) would be : 
`SenMLDataPoint` => `SenMLRecord`
`SensorMeasurementList` => `SenMLPack`

2. Is there any reason to not follow the `org.eclipse.leshan.json` way ?
I mean : 
`JsonRootObject` => `SenMLPack`
`JsonArrayEntry`  => `SenMLRecord`
`JsonRootObjectSerDes` => `SenMLPackSerDes` (with only the encoding part)
`JsonArrayEntrySerDes`  => `SenMLRecordSerDes` (with only the encoding part)
`LwM2mJson` => `SenMLJson`
and some tests like in `org.eclipse.leshan.json` ?
If there is no good reason, I would prefer to keep it for consistency.

3. I didn't get the point about SenMLabel and SenMLJsonLabel ?

4. I would really appreciate if we put LwM2mNodeSenJsonEncoder for another PR.

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPack.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * The class representing the SenML Pack of LwM2M
+ */
+public class SenMLPack {
+
+    private String baseName = null;
```
sbernard31: As I understand the specification `base fields` should be SenMLRecord attributes.
https://tools.ietf.org/html/rfc8428#section-2

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
```diff
@@ -0,0 +1,78 @@
...
+                switch (records.getType()) {
+                case FLOAT:
+                case INTEGER:
+                    jsonObj.add("v", records.getFloatValue().floatValue());
+                    break;
+                case BOOLEAN:
+                    jsonObj.add("vb", records.getBooleanValue());
+                    break;
+                case OBJLNK:
+                    jsonObj.add("vd", records.getObjectLinkValue());
```
sbernard31: objlink is `vlo`

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
```diff
@@ -0,0 +1,78 @@
...
+
+            if (records.getTime() != null) {
+                jsonObj.add("t", records.getTime());
+            }
+
+            Type type = records.getType();
+            if (type != null) {
+                switch (records.getType()) {
+                case FLOAT:
+                case INTEGER:
```
sbernard31: reading the spec I didn't see INTEGER type :
https://tools.ietf.org/html/rfc8428#section-4.3
http://www.openmobilealliance.org/release/LightweightM2M/V1_1-20180710-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1-20180710-A.html#7-4-4-0-744-JSON

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
```diff
@@ -0,0 +1,78 @@
...
+                    jsonObj.add("v", records.getFloatValue().floatValue());
+                    break;
+                case BOOLEAN:
+                    jsonObj.add("vb", records.getBooleanValue());
+                    break;
+                case OBJLNK:
+                    jsonObj.add("vd", records.getObjectLinkValue());
+                    break;
+                case OPAQUE:
+                    jsonObj.add("vd", Hex.encodeHexString(records.getOpaqueValue()));
```
sbernard31: opaque should be encoded in base64 :
_"Data Value is a base64-encoded string with the URL-safe alphabet as defined in Section 5 of [RFC4648], with padding omitted"_
(source : https://tools.ietf.org/html/rfc8428#section-4.3)

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLRecord.java
```diff
@@ -0,0 +1,208 @@
...
+        return null;
+    }
+
+    public void setTime(Long time) {
+        this.time = time;
+    }
+
+    public Long getTime() {
+        return time;
+    }
```
sbernard31: I didn't see any time value. 
Neither in https://tools.ietf.org/html/rfc8428#section-4.3 nor in http://www.openmobilealliance.org/release/LightweightM2M/V1_1-20180710-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1-20180710-A.html#7-4-4-0-744-JSON

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/DefaultLwM2mNodeEncoder.java
```diff
@@ -71,6 +72,9 @@ public DefaultLwM2mNodeEncoder(LwM2mValueConverter converter) {
         case ContentFormat.OLD_JSON_CODE:
             encoded = LwM2mNodeJsonEncoder.encode(node, path, model, converter);
             break;
+        case ContentFormat.SENML_JSON_CODE:
+            encoded = LwM2mNodeSenMLJsonEncoder.encodeToJson(node, path, model, converter);
+            break;
```
sbernard31: The code does not compile because of this. Just remove it for now.

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLRecord.java
```diff
@@ -0,0 +1,208 @@
...
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public Long getTimeValue() {
+        return timeValue;
+    }
+
+    public void setTimeValue(Long time) {
```
sbernard31: I didn't see any time value. 
Neither in https://tools.ietf.org/html/rfc8428#section-4.3 nor in http://www.openmobilealliance.org/release/LightweightM2M/V1_1-20180710-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1-20180710-A.html#7-4-4-0-744-JSON

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
cavenaghi9: By referring above link, my understanding float value attribute of SenML record can be used to represent Integer, Float or Time Resource data type. Time Resource data type, the value shall be the difference, measured in seconds(not milliseconds), between the current time and midnight, January 1, 1970 UTC. In addition, do you think value of time data type should avoid to use scientific notation?

sbernard31: I understand this like that : 
The record type is FLOAT/Number.
This float can be used to store Integer, Float or Time, but at SenML level we have not this information.

At LWM2M level, thanks to LWM2M model we will be able to decode the float in float/integer or time.

sbernard31: > In addition, do you think value of time data type should avoid to use scientific notation?

I don't think so. This is just notation, I can see any good reason to forbid it for now. Do you ?

cavenaghi9: Agree.

sbernard31:
:+1: 

I squash all the commit in one and fix a little formatting issue.
This is now integrated in branch 2.0.x (commit 77abfe9)

This is the first commit of the 2.0.x branches. Thx @cavenaghi9  !

Next step ?

PR #674, opened 2019-04-04
Initial support for OSCORE security object

rikard-sics:
Beginning to add support for the OSCORE security object.

Signed-off-by: Rikard Höglund <rikard.hoglund@ri.se>

### DIFF
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/object/OSCORESecurity.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/object/OSCORESecurity.java
@@ -1,0 +1,109 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Sierra Wireless and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *     Sierra Wireless - initial API and implementation
+ *     Rikard Höglund (RISE SICS)
+ *
+ *******************************************************************************/
+package org.eclipse.leshan.client.object;
+
+import org.eclipse.leshan.client.request.ServerIdentity;
+import org.eclipse.leshan.client.resource.BaseInstanceEnabler;
+import org.eclipse.leshan.client.resource.LwM2mInstanceEnabler;
+import org.eclipse.leshan.core.node.LwM2mResource;
+import org.eclipse.leshan.core.response.ReadResponse;
+import org.eclipse.leshan.core.response.WriteResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A simple {@link LwM2mInstanceEnabler} for the OSCORE Security (0) object.
+ */
+public class OSCORESecurity extends BaseInstanceEnabler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(Security.class);
+
+    byte[] OSCORE_master_secret;
+    String OSCORE_sender_id;
+    String OSCORE_recipient_id;
+    int OSCORE_AEAD_algorithm;
+    int OSCORE_HMAC_algorithm;
+
+    public OSCORESecurity() {
+
+    }
+
+	public OSCORESecurity(byte[] OSCORE_master_secret, String OSCORE_sender_id, String OSCORE_recipient_id,
+			int OSCORE_AEAD_algorithm, int OSCORE_HMAC_algorithm) {
+		super();
+		this.OSCORE_master_secret = OSCORE_master_secret.clone();
+		this.OSCORE_sender_id = OSCORE_sender_id;
+		this.OSCORE_recipient_id = OSCORE_recipient_id;
+		this.OSCORE_AEAD_algorithm = OSCORE_AEAD_algorithm;
+		this.OSCORE_HMAC_algorithm = OSCORE_HMAC_algorithm;
+	}
+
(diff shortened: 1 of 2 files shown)


leshan-client-core/src/main/java/org/eclipse/leshan/client/object/OSCORESecurity.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+public class OSCORESecurity extends BaseInstanceEnabler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(Security.class);
+
+    byte[] OSCORE_master_secret;
+    String OSCORE_sender_id;
+    String OSCORE_recipient_id;
+    int OSCORE_AEAD_algorithm;
+    int OSCORE_HMAC_algorithm;
```
sbernard31: 1. Generally in Java, camelCase is prefered, so we do that in Leshan too.
One exception in Leshan, we are using snake case for test name.
So it should looks like `oscoreMasterSecret`.

2. Any reason do not use private access modifier ?

3. I know in the LWM2M specification,  resource name are prefixed by OSCORE and generally I prefer to keep as closest as possible as specification naming. But in this case I think it should be less verbose to just remove the `OSCORE` part in attribute name. So it looks like `masterSecret`, does it sound good to you ?

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/OSCORESecurity.java
```diff
@@ -0,0 +1,109 @@
...
+import org.eclipse.leshan.core.node.LwM2mResource;
+import org.eclipse.leshan.core.response.ReadResponse;
+import org.eclipse.leshan.core.response.WriteResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A simple {@link LwM2mInstanceEnabler} for the OSCORE Security (0) object.
+ */
+public class OSCORESecurity extends BaseInstanceEnabler {
```
sbernard31: The LWM2M object is named `LWM2M OSCORE`, should we name this java class `Oscore` ?

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/OSCORESecurity.java
```diff
@@ -0,0 +1,109 @@
...
+import org.eclipse.leshan.client.resource.BaseInstanceEnabler;
+import org.eclipse.leshan.client.resource.LwM2mInstanceEnabler;
+import org.eclipse.leshan.core.node.LwM2mResource;
+import org.eclipse.leshan.core.response.ReadResponse;
+import org.eclipse.leshan.core.response.WriteResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A simple {@link LwM2mInstanceEnabler} for the OSCORE Security (0) object.
```
sbernard31: LWM2M OSCORE object ID is 21, right ?

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/OSCORESecurity.java
```diff
@@ -0,0 +1,109 @@
...
+    public WriteResponse write(ServerIdentity identity, int resourceId, LwM2mResource value) {
+    	LOG.debug("Write on resource {}: {}", resourceId, value);
+    	// extend
+		return null;
+    }
+    
+    @Override
+    public ReadResponse read(ServerIdentity identity, int resourceid) {
+    	// extend
+		return null;
```
sbernard31: I suppose next step will be to implement this feature so it's not a big deal but just to let you know  the idea is more to return `ReadResponse.notFound()` or `ReadResponse.internalServerError("not implemented")` instead or return `null`

rikard-sics:
Thanks for the feedback on the code. I will start going through it and adjust the pull request accordingly.

rikard-sics:
Again, thank you for the feedback. I have now done 2 more commits where I believe I addressed all the points you listed.

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Oscore.java
```diff
@@ -0,0 +1,108 @@
...
+    @Override
+    public WriteResponse write(ServerIdentity identity, int resourceId, LwM2mResource value) {
+        LOG.debug("Write on resource {}: {}", resourceId, value);
+        // extend
+        return WriteResponse.notFound();
+    }
+
+    @Override
+    public ReadResponse read(ServerIdentity identity, int resourceid) {
+        // extend
```
sbernard31: Any reason why you add log in write and not in read ?

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Oscore.java
```diff
@@ -0,0 +1,108 @@
...
+        return ReadResponse.notFound();
+    }
+
+    public byte[] getOSCOREMasterSecret() {
```
sbernard31: As you renamed the attributes, you should rename getter, setter and setter parameters  too ?

rikard-sics: Yes, indeed I should do that.

sbernard31:
It sounds good !  :)

Do you want, I integrate this is 2.0.x branch and you provide a new PR ?

rikard-sics:
> It sounds good ! :)
> 
> Do you want, I integrate this is 2.0.x branch and you provide a new PR ?
> (I will squash all commits in one)

Yes, thanks. It would be great if you could do that. :)

sbernard31:
Integrated in 2.0.x (all squashed in one commit bb2f1e5)

Thx @rikard-sics

rikard-sics:
> Integrated in 2.0.x (all squashed in one commit [bb2f1e5](https://github.com/eclipse/leshan/commit/bb2f1e52cc8d35ad340cc1e828c58da861abd393))

Thanks a lot for all the feedback and assistance!

Issue #675, opened 2019-04-05
LwM2mNodeTextEncoder do not support OPAQUE resources


davideicardi:
The `LwM2mNodeTextEncoder` doesn't support encoding `OPAQUE` resources.
From specification it seems to be a supported scenario: http://www.openmobilealliance.org/release/LightweightM2M/V1_0_2-20180209-A/OMA-TS-LightweightM2M-V1_0_2-20180209-A.pdf

From APPENDIX C: 
> Opaque:  Represented as a Base64 encoding of the binary data [RFC4648]. For example, the sequence of bytes (in hex notation) {0x01, 0x02, 0x03, 0x04, 0x05} converts to the ASCII string “AQIDBAU=” in Base64 encoding.

It is by design or it is just a missing feature? Can I try to write a pull request?

thanks!

sbernard31:
We started to implement the specification when it was just a draft.
By the past, this was not allowed : http://www.openmobilealliance.org/release/LightweightM2M/V1_0-20141126-C/OMA-TS-LightweightM2M-V1_0-20141126-C.pdf

And it seems we missed this addition.

> Can I try to write a pull request? 

of course :)
We have a Base64 class which could help, see https://github.com/eclipse/leshan/blob/leshan-1.0.0-M10/leshan-core/src/main/java/org/eclipse/leshan/util/Base64.java

sbernard31:
This is now implemented in : https://github.com/eclipse/leshan/releases/tag/leshan-1.0.0-M11
Here an issue about potential enhancement and future behavior changing : https://github.com/eclipse/leshan/issues/678 

Thx @davideicardi !

PR #676, opened 2019-04-05
Added OPAQUE text encoding, using base64

davideicardi:
Signed-off-by: Davide Icardi <davide.icardi@gmail.com>

See #675


davideicardi:
Some doubts:
- Do we need to handle null values?
- It is correct to assume that the value is a `byte[]`?

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/text/LwM2mNodeTextEncoder.java
```diff
@@ -100,6 +101,10 @@ public void visit(LwM2mResource resource) {
                 ObjectLink objlnk = (ObjectLink) val;
                 strValue = String.valueOf(objlnk.getObjectId() + ":" + objlnk.getObjectInstanceId());
                 break;
+            case OPAQUE:
+                byte[] binaryValue = ((byte[])val);
```
sbernard31: formatting issue : no space between `(byte[])` and `val`

I can fix this myself, but If you plan to contribute often there is [an eclipse formatter file](https://github.com/eclipse/leshan/wiki/Code-&-design-guidelines#configure-your-ide).
I don't know if `Visual Studio Code` is able to handle it ?

If you don't, I can easily fix this before to integrate your commit, just let me know !

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
```diff
@@ -89,6 +89,15 @@ public void text_encode_multiple_instances() {
                 model);
     }
 
+    @Test
+    public void text_encode_opaque_as_base64_string() {
+        byte[] opaqueValue = new byte[] { 1, 2, 3, 4, 5 };
```
sbernard31: Just a detail, as the spec talk about hex value maybe we should use hex notation?
```java
byte[] opaqueValue = new byte[] { 0x1, 0x2, 0x3, 0x4, 0x5 };
```

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/text/LwM2mNodeTextEncoder.java
davideicardi: No problem, I will try without Eclipse. If you see other problems  just let me know.

davideicardi:
I will not have a look at the decoder. I think to add it here if for you is ok.

sbernard31:
> I think to add it here if for you is ok.

:+1:

davideicardi:
Added decoder and test. I have squashed all commits ...

Do you think that we need a try/catch inside the decoder?

sbernard31:
> Do you think that we need a try/catch inside the decoder?

I would say we should handle the case where the string is not a valid base64 payload and so raise a CodecException !

But the `Base64`  javadoc is not so verbose about that ...
(this is a copy paste from apache commons codec, we didn't want to depends to this jar only for 2 methods)

So reading the code, it seems that IllegalStateException could be raised ?
There is also a `isBase64` method ...
So several solutions : 
- if isBase64() ?
- try catch IllegalStateException ?
- try catch Exception ? (probably not a so good idea)

We should add some test with bad payload, we will see how base64 behave and you could try several solutions and choose the best one ?

davideicardi:
I have tried to write a test with an input that cause the `decode` to throw but without success. It seems that when an invalid characters is found it simply ignore it or trasform it to a `+`.

I have added a test to ensure this behaviour, see `text_decode_opaque_should_ignore_not_base64_chars`.

We can probably use `isBase64` and throw but not sure if it is a good idea, maybe the validation should be performed at an higher level...

In any case I'm open to suggestions, I will follow you ;-)

sbernard31:
> It seems that when an invalid characters is found it simply ignore it or transform it to a +

With this implementation, Invalid characters is ignored.
The "-" transform in "+" is about url safe variant. see [rfc4648#section-5](https://tools.ietf.org/html/rfc4648#section-5), by default this implementation is able to decode both standard and url safe alphabet. But we encode with standard alphabet.

As I was a bit lost with all of this ... I read the [RFC4648](https://tools.ietf.org/html/rfc4648).
And my understanding is that rfc recommends to : 
 - [NOT ignoring non alphabet characters](https://tools.ietf.org/html/rfc4648#section-3.3)
 - [Do not accept missing padding character](https://tools.ietf.org/html/rfc4648#section-3.2)
 - [be strict about canonical form](https://tools.ietf.org/html/rfc4648#section-3.5) meaning that padding bit must be 0 !

By the way the LWM2M specification does not talk about url safe alphabet.

So a perfect solution should be to have a default base64 decoder which :
 1. rejects url safe alphabet : `A-BB` should be rejected (the good one is `A+BB`)
 2. rejects invalid character : `'A+B;B!` should be rejected
 3. rejects  1 bit in  padding  :  `ABC=` should be rejected (the canonical form is  `ABA=`) 
 4. rejects missing padding character : `ABA` should be rejected (missing pad char `ABA=`)

by raising a CodecException.
And if user want a different behavior it could change it. (we will see later how to do that)

 1. It seems we can not do that with this implementation ...
 2. is doable with `isBase64` check...          
 3. it seems we can not do that with this implementation ...
 4. it seems we can not do that with this implementation ...

So, we can :
- go for a first shot implementation with just 2. and searching to improve this in a second time by searching a more appropriate implementation.
- or we search a good implementations for our needs now.

davideicardi:
Thanks! Good investigation!

If for you is ok I will proceed with:
> go for a first shot implementation with just 2. and searching to improve this in a second time by searching a more appropriate implementation.

I will first call `isBase64` and in case of false I will throw a `CodecException`.

Ok?

sbernard31:
> I will first call isBase64 and in case of false I will throw a CodecException.
Ok?

:+1:

davideicardi:
Ok done! let met know if you see problems ...

sbernard31:
Sounds good, I will integrate this in master. (I create a new issue about issue we discover here, see #678 )

sbernard31:
Thx @davideicardi  !

sbernard31:
We have a jenkins build : https://hudson.eclipse.org/leshan/job/leshan/, but jars are not deployed ... :disappointed: 

But [building jars locally is pretty easy](https://github.com/eclipse/leshan#compile-leshan--run-demos), doing this you will have a leshan-*-SNAPSHOT available in your local maven repository and so you could test it easily just adding SNAPSHOT version to you dependencies.

Last point, if you need a release just let me know :)

davideicardi:
The problem is that we have some build pipelines running with the Leshan dependency ... so building it locally can be a little difficult ... Taking it from a maven repo is easier...
If you are able to release a new version in one week or two I will appreciate, otherwise no problem, I will pubish it in some temporary repo 😉

sbernard31:
> If you are able to release a new version in one week or two I will appreciate

I will try, but maybe doing some validation before could be a good idea :thinking: 
(I would like to integrate this PR #657 too but I wait for a teammate review :) )

sbernard31:
@davideicardi did you made some tests ?
I plan to release a 1.0.0-M11 today ?

davideicardi:
Good news! I have used a local compiled version during these days and for now it is working fine. Green light for me!

sbernard31:
You probably don't care but finally #657 will not be integrated in this release. (next one)

davideicardi:
Yes, no problem for me.

I think that now we can close #675 issue right? we can leave #678 open for the enhancements ..

PR #677, opened 2019-04-08
LwM2mNodeSenMLEncoder for SenML-JSON in leshan-core with unit tests

cavenaghi9:
Signed-off-by: Boya Zhang <zhangboya@gmail.com>

LwM2mNodeSenMLEncoder for SenML-JSON in leshan-core with unit tests
- LwM2mNodeSenMLEncoder will use to encode LwM2M nodes with SenML-JSON and later SenML CBOR formats.
- New test cases in LwM2mNodeEncoderTest hasn't use DefaultLwM2mNodeEncoder, because of it doesn't include SenML-JSON yet, as will be updated in later PR.

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/senml/LwM2mNodeSenMLJsonEncoder.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/senml/LwM2mNodeSenMLJsonEncoder.java
@@ -39,8 +39,6 @@ import org.slf4j.LoggerFactory;
 public class LwM2mNodeSenMLJsonEncoder {
     private static final Logger LOG = LoggerFactory.getLogger(LwM2mNodeSenMLJsonEncoder.class);
 
-    private static final String EMPTY_RESOURCE_PATH = "";
-
     public static byte[] encode(LwM2mNode node, LwM2mPath path, LwM2mModel model, LwM2mValueConverter converter) {
         Validate.notNull(node);
         Validate.notNull(path);
@@ -63,7 +61,6 @@ public class LwM2mNodeSenMLJsonEncoder {
         private int objectId;
         private LwM2mModel model;
         private LwM2mPath requestPath;
-        private Long timestamp;
         private LwM2mValueConverter converter;
 
         // visitor output
@@ -77,11 +74,11 @@ public class LwM2mNodeSenMLJsonEncoder {
                 throw new CodecException("Invalid request path %s for object encoding", requestPath);
             }
 
-            // Create resources
+            // Create SenML records
             for (LwM2mObjectInstance instance : object.getInstances().values()) {
                 for (LwM2mResource resource : instance.getResources().values()) {
                     String prefixPath = Integer.toString(instance.getId()) + "/" + Integer.toString(resource.getId());
-                    lwM2mResourceToSenMLRecord(prefixPath, timestamp, resource);
+                    lwM2mResourceToSenMLRecord(prefixPath, resource);
                 }
             }
         }
@@ -99,8 +96,8 @@ public class LwM2mNodeSenMLJsonEncoder {
                 } else {
                     throw new CodecException("Invalid request path %s for instance encoding", requestPath);
                 }
-                // Create resources
-                lwM2mResourceToSenMLRecord(prefixPath, timestamp, resource);
+                // Create SenML records
+                lwM2mResourceToSenMLRecord(prefixPath, resource);
             }
         }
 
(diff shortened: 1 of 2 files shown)


leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/senml/LwM2mNodeSenMLJsonEncoder.java
```diff
@@ -0,0 +1,191 @@
...
+            }
+            
+            if(requestPath.isResource()) {
+                record.setBaseName(requestPath.toString());
+            }
+            else if(requestPath.isObjectInstance()) {
+                record.setName(resourcePath);
+                record.setTime(timestamp);
+            }
+
```
sbernard31: I suspect the way, setBaseName and Name is not correct ? What happens for `Object` path ?

A first minimal way could looks like this ?
```java
 // We set basename only on first record.
if (records.isEmpty()) {
    record.setBaseName(requestPath.toString() + "/");
}

// Set name if present
if (resourcePath != null && !resourcePath.isEmpty()) {
    record.setName(resourcePath);
}
```

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/senml/LwM2mNodeSenMLJsonEncoder.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        @Override
+        public void visit(LwM2mResource resource) {
+            LOG.trace("Encoding resource {} into SenML JSON", resource);
+            if (!requestPath.isResource()) {
+                throw new CodecException("Invalid request path %s for resource encoding", requestPath);
+            }
+
+            lwM2mResourceToSenMLRecord(EMPTY_RESOURCE_PATH, timestamp, resource);
```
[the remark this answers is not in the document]
sbernard31: :+1:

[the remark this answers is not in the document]
cavenaghi9: More complex that what I thought on yesterday, when I write unit test to cover object instance, single resource and multiple resource instance cases. Please kindly review, let me know if any further feedback.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
```diff
@@ -286,8 +286,10 @@ public void senml_json_encode_device_object_instance() {
         byte[] encoded = LwM2mNodeSenMLJsonEncoder.encode(oInstance, new LwM2mPath("/3/0"), model,
                 new DefaultLwM2mValueConverter());
 
+        System.out.println(new String(encoded));
```
sbernard31: debug sysout ?

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
```diff
@@ -316,5 +318,15 @@ public void senml_json_encode_resource() {
...
         Assert.assertEquals(expected, new String(encoded));
+
+        Map<Integer, Long> values = new HashMap<>();
+        values.put(0, 3800L);
+        values.put(1, 5000L);
+        oResource = LwM2mMultipleResource.newIntegerResource(7, values);
+        encoded = LwM2mNodeSenMLJsonEncoder.encode(oResource, new LwM2mPath("/3/0/7"), model,
+                new DefaultLwM2mValueConverter());
+
+        expected = "[{\"bn\":\"/3/0/7\",\"n\":\"0\",\"v\":3800},{\"n\":\"1\",\"v\":5000}]";
```
sbernard31: I think the test is incorrect as `bn` should end with "/". (or n should start with "/")
I prefer you make 2 tests 1 for single resource and 1 for multiple resource.

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/senml/LwM2mNodeSenMLJsonEncoder.java
```diff
@@ -111,54 +108,55 @@ public void visit(LwM2mResource resource) {
...
-            else if(requestPath.isObjectInstance()) {
-                record.setName(resourcePath);
-                record.setTime(timestamp);
+
+            // Compute baseName and name for SenML record
+            String bn = requestPath.toString();
+            String n = recordName == null ? "" : recordName;
+
+            if (records.isEmpty()) {
+                if (!requestPath.isResource()) {
```
sbernard31: I thing the right test is something like `!n.isEmpty()`

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
cavenaghi9: Will be removed in next commit.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
cavenaghi9: I checked [Table: 7.4.4.-5 JSON payload returned from example request to Device Object on Resource 0 of the LwM2M example client (Read /3/0/0)](http://www.openmobilealliance.org/release/LightweightM2M/V1_1-20180710-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1-20180710-A.html#7-4-4-0-744-JSON), bn doesn't end with "/", if record is resouce.

_[{"bn":"/3/0/0","vs":"Open Mobile Alliance}]_

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/senml/LwM2mNodeSenMLJsonEncoder.java
cavenaghi9: Not fully sure I got your point. Do you mean to enhance line 144 or combine 146 and 147? When encoding object instance, 'n' is resource ID.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
sbernard31: You're correct about this example but the reason is not because this is a resource but because there is no "n".

In fact, you could put anything in bn and n, the only condition is that concatenation of both should be the path of the resource !

[SenML RFC](https://tools.ietf.org/html/rfc8428#section-2) says : _"To keep the messages small, it does not make sense to repeat the "n"
   field in each SenML Record, so there is a concept of a Base Name,
   which is simply a string that is prepended to the Name field of all
   elements in that Record and any Records that follow it. "_

Or look at field description of [Table: 7.4.4.-1 JSON format and description](http://www.openmobilealliance.org/release/LightweightM2M/V1_1-20180710-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1-20180710-A.html#7-4-4-0-744-JSON)

You could also look at [LWM2M discussion](https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/380) if you are not convinced !

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/senml/LwM2mNodeSenMLJsonEncoder.java
sbernard31: This is for fixing, the point above https://github.com/eclipse/leshan/pull/677#discussion_r273973353

But we should agree on this point first :)

sbernard31: this is for replacing test line 147.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
cavenaghi9: Understood. Thanks! 😉.

sbernard31:
Integrated in 2.0.x (commit aa68bca)

PR #682, opened 2019-04-12
 SenML-CBOR encoder in leshan-core with unit tests

cavenaghi9:
SenML-CBOR encoder in leshan-core with unit tests


leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
```diff
@@ -126,4 +131,64 @@ public SenMLPack deserialize(JsonArray array) {
...
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+
+        try {
+            CBORGenerator generator = factory.createGenerator(out);
+            generator.writeStartArray();
+
+            for (SenMLRecord record : pack.getRecords()) {
+                generator.writeStartObject();
+
+                if (record.getBaseName() != null && record.getBaseName().length() > 0) {
```
sbernard31: Why not using `!record.getBaseName().isEmpty()` ?

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
```diff
@@ -126,4 +131,64 @@ public SenMLPack deserialize(JsonArray array) {
...
+                        break;
+                    }
+                }
+                generator.writeEndObject();
+            }
+
+            generator.writeEndArray();
+            generator.close();
+        } catch (Exception e) {
+            throw new CodecException(e.getLocalizedMessage(), e);
```
sbernard31: Instead of e.getLocalizedMessage() which is already available in e parameter. Should we add a more contextual  exception message, like : "Unable to serialize to Cbor." ?

leshan-core/src/test/java/org/eclipse/leshan/senml/SenMLSerializerTest.java
```diff
@@ -120,7 +124,22 @@ public void serialize_device_object() {
...
+    public void serialize_device_object_to_senml_cbor() {
+        byte[] cbor = SenMLCbor.toSenMLCbor(givenDeviceObjectInstance());
+        LOG.debug("CBOR: " + Hex.encodeHexString(cbor));
+
+        String expectedValue = "9fbf622d32652f332f302f613061306133744f70656e204d6f62696c6520416c6c69616e6365ffbf6"
+                + "13061316133764c69676874776569676874204d324d20436c69656e74ffbf61306132613369333435303030313233f"
+                + "fbf613063362f306132fa3f800000ffbf613063362f316132fa40a00000ffbf613063372f306132fa456d8000ffbf6"
+                + "13063372f316132fa459c4000ffbf613063382f306132fa42fa0000ffbf613063382f316132fa44610000ffbf61306"
+                + "1396132fa42c80000ffbf61306231306132fa41700000ffbf61306431312f306132fa00000000ffbf61306231336132"
+                + "fa4ea30485ffbf61306231346133662b30323a3030ffbf613062313561336155ffff";
```
sbernard31: This doesn't look like to the [CBOR payload example](http://www.openmobilealliance.org/release/LightweightM2M/V1_1-20180710-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1-20180710-A.html#7-4-5-0-745-CBOR) from the specification ? Does it should ?

cavenaghi9:
> I ask myself, if it makes sense to put cbor serialization code in SenMLPackSerDes class instead of a separated class?

If to give two seperated classes, one is named SenMLCborPackSerDes.java, another is named SenMLJsonPackSerDes.java, will it make sense?

sbernard31:
Yep, something like this.

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
cavenaghi9: Not all records have base name within same pack, so that only generate base name element when record contains base name and length great than 0.

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
cavenaghi9: In addition, do you think CodecException can be used here, because of CodecException is thrown when a problem occurs during LWM2M node encoding or decoding. Or to extend a new Exception class for SenML-CBOR, like TlvException and LwM2MJsonException.

leshan-core/src/test/java/org/eclipse/leshan/senml/SenMLSerializerTest.java
cavenaghi9: Good point, example in TS is {-2: "/3/0/", 0: "0", 3: "Open Mobile Alliance"}...., however generated CBOR notation in test case is {"-2": "/3/0/", 0: "0", 3: "Open Mobile Alliance"}. Field name shall be number type, instead of string type. I will check with Jackson CBOR library, to correct it.

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
sbernard31: I meant replacing `record.getBaseName().length() > 0` by `!record.getBaseName().isEmpty()`
(keep the  `record.getBaseName() != null`)

leshan-core/src/main/java/org/eclipse/leshan/senml/SenMLPackSerDes.java
sbernard31: Yep maybe this is even better. You can create a `SenMLCborException`.

leshan-core/src/test/java/org/eclipse/leshan/senml/SenMLCborSerializerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.eclipse.leshan.util.Hex;
+import org.junit.Assert;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SenMLCborSerializerTest extends AbstractSenMLTest {
+
+    private Logger LOG = LoggerFactory.getLogger(getClass());
```
sbernard31: Any reason to not use a static attribute for logger ?

leshan-core/src/test/java/org/eclipse/leshan/senml/SenMLCborSerializerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.eclipse.leshan.senml;
+
+import org.eclipse.leshan.util.Hex;
+import org.junit.Assert;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SenMLCborSerializerTest extends AbstractSenMLTest {
```
sbernard31: Generally, in this project we use "delegation" instead of "inheritance" for reuse code in tests.
But that's not a big deal, if let it like this :)

cavenaghi9:
> Sorry for the long delay...
> Never so closed to merge it ;)

Not so long time. I was taking break as well in past days. It is time to continue contribution.

leshan-core/src/test/java/org/eclipse/leshan/senml/SenMLCborSerializerTest.java
cavenaghi9: I will take care about it.

leshan-core/src/test/java/org/eclipse/leshan/senml/SenMLCborSerializerTest.java
cavenaghi9: After reviewing code, I found that LOGGER can be removed completely.

sbernard31:
Integrated in 2.0.x branch (all commits squashed in 876c1a2)

Thx @cavenaghi9 :+1:

PR #684, opened 2019-04-18
Initial support for OSCORE Context info on Leshan server web interface

rikard-sics:
The web interface of the Leshan server has been extended with some OSCORE-related fields. The idea is to continue extending this to enable specifying OSCORE Context information for clients via the web interface.

### DIFF
--- a/leshan-server-demo/src/main/resources/webapp/js/security-controllers.js
+++ b/leshan-server-demo/src/main/resources/webapp/js/security-controllers.js
@@ -99,6 +99,9 @@ angular.module('securityControllers', [])
                     var security = {endpoint: $scope.endpoint, psk : { identity : $scope.pskIdentity , key : $scope.pskValue}};
                 } else if($scope.securityMode == "rpk") {
                     var security = {endpoint: $scope.endpoint, rpk : { x : $scope.rpkXValue , y : $scope.rpkYValue, params : $scope.rpkParamsValue || $scope.defaultParams}};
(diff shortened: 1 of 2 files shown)


sbernard31:
Generally I start by implemented core feature, then add/use it in demo.

Do you think this method is appropriate for integrating OSCORE in Leshan ?

I didn't look deeper about OSCORE integration in LWM2M specification so maybe this does not make so much sense, but if this is the case I would prefer we proceed like this.

What about you ?

(By the way I will not be available next 2,5 week, so don't worried if I didn't reply)

rikard-sics:
> Generally I start by implemented core feature, then add/use it in demo.
> 
> Do you think this method is appropriate for integrating OSCORE in Leshan ?
> 
> I didn't look deeper about OSCORE integration in LWM2M specification so maybe this does not make so much sense, but if is does I would prefer we proceed like this.
> 
> What about you ?

I think that could be a feasible approach for this work also. I can work on creating somewhat bigger pull requests that add specific functionality and also a demonstration of that functionality being used elsewhere in the code.

For instance when it comes to this pull request it could be functionality to add OSCORE context information in the web interface of the Leshan server. Then also a demonstration of how that added context information will be parsed and finally used for communication between the client and server.

> (By the way I will not be available next 2,5 weeks, so don't worry if I don't reply)

Thanks for the heads up.

rikard-sics:
I think this pull request could be closed and I will proceed with creating a different one according to our discussion.

Issue #694, opened 2019-05-28
why IntegerResource value didn't use Integer as data type but Long? As result,when use diagnostic tools to verify, the type is not correct.


beclever:
when I verify the CBOR string :90a321642f332f3000613003744f70656e204d6f62696c6520416c6c69616e6365a200613103764c69676874776569676874204d324d20436c69656e74a20061320369333435303030313233a20061330363312e30a20063362f3002fa3f800000a20063362f3102fa40a00000a20063372f3002fa456d8000a20063372f3102fa459c4000a20063382f3002fa42fa0000a20063382f3102fa44610000a200613902fa42c80000a20062313002fa41700000a20062313102fa00000000a200623133021a5182428fa20062313403662b30323a3030a200623136036155

in [http://cbor.me/](url) ,the result is : 
[{-2: "/3/0", 0: "0", 3: "Open Mobile Alliance"}, {0: "1", 3: "Lightweight M2M Client"}, {0: "2", 3: "345000123"}, {0: "3", 3: "1.0"}, {0: "6/0", 2: 1.0}, {0: "6/1", 2: 5.0}, {0: "7/0", 2: 3800.0}, {0: "7/1", 2: 5000.0}, {0: "8/0", 2: 125.0}, {0: "8/1", 2: 900.0}, {0: "9", 2: 100.0}, {0: "10", 2: 15.0}, {0: "11", 2: 0.0}, {0: "13", 2: 1367491215}, {0: "14", 3: "+02:00"}, {0: "16", 3: "U"}]

I found there is primitive type in the long type, like 1.0 in  '{0: "6/0", 2: 1.0}'.

Does anybody have this problem?

sbernard31:
I suppose you are testing code from #682 ?

> I found there is primitive type in the long type, like 1.0 in '{0: "6/0", 2: 1.0}'.

 You mean you expect an integer value instead of a floating point one ? 

--------------------------

LWM2M 1.1 use SenML-CBOR representation.

The [RFC 8428](https://tools.ietf.org/html/rfc8428#section-4.3) (senML) says that  `value` label is of Type `Number`.

In CBOR, there is several way to encode `Number` but the specification allow to encode an integer with a floating point value. See [RFC 7049](https://tools.ietf.org/html/rfc7049#section-3.6)(CBOR) : 
> For the purposes of this specification, all number representations
   for the same numeric value are equivalent.  This means that an
   encoder can encode a floating-point value of 0.0 as the integer 0.
[...]

PR #708, opened 2019-06-26
SenML floating 

cavenaghi9:
#694


sbernard31:
This should be obsoleted by #922

PR #710, opened 2019-06-28
Do not do type conversion at encoding time by default anymore.

sbernard31:
Until now Leshan used by default a kind of magic converter.  
This converter was used `LwM2mNodeEncoder` to fix `LwM2mResource` which would used a different
`ResourceModel.Type` than the one defined in the `ResourceModel`.

This magic conversion was mainly used to correct value which came from the leshan-server-demo UI...
So this does not make so much sense to keep this behavior by default.

So now the default converter is a `NoValueConverter` which do not do any converstion but raise an exception, if type in resource does not match the model.

**If you want to keep the previous behavior** you need to duplication class from leshan-server-demo : `MagicLwM2mValueConverter` and create your server like this : 

```java
builder.setEncoder(new DefaultLwM2mNodeEncoder(new MagicLwM2mValueConverter()));
LeshanServer server = builder.build();

```


leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/NoValueConverter.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.eclipse.leshan.core.model.ResourceModel.Type;
+import org.eclipse.leshan.core.node.LwM2mPath;
+
+/**
+ * A {@link LwM2mValueConverter} which do no conversion but raise a {@link CodecException} if type is not the expected
+ * one.
+ *
+ */
+public class NoValueConverter implements LwM2mValueConverter {
```
msangoi: This `NoValue` name is bit confusing. It is more a `NoConversionConverter` 😄.
`CheckTypeConverter`?

sbernard31: I renamed it by `LwM2mValueChecker`

sbernard31:
I prefer to not keep in in `leshan-core` as it is only needed for `leshan-server-demo`

PR #713, opened 2019-07-04
Make bootstrap server more robust and customizable.

sbernard31:
1. Add more callback on BootstrapSessionManager for better monitoring.
2. Add a way to customize bootstrapHandler behavior on error.  
3. Improve BootstrapHandler to make it able to avoid to start 2 bootstrap session at the same time for the same device.  
4. Logs enhancement

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager.java
@@ -15,7 +15,6 @@
  *******************************************************************************/
 package org.eclipse.leshan.server.bootstrap;
 
-import org.eclipse.leshan.core.request.DownlinkRequest;
 import org.eclipse.leshan.core.request.Identity;
 import org.eclipse.leshan.core.request.LwM2mRequest;
 import org.eclipse.leshan.core.response.LwM2mResponse;
@@ -31,6 +30,29 @@ import org.eclipse.leshan.server.impl.DefaultBootstrapSessionManager;
  * @see BootstrapSession
  */
 public interface BootstrapSessionManager {
+
+    public enum BootstrapPolicy {
+        /**
+         * Continue bootstrap session ignoring failure.
+         */
+        CONTINUE,
+        /**
+         * Try to resend the failed request.
+         */
+        RETRY,
+        /**
+         * Retry the bootstrap session from the beginning generally.
+         */
+        RETRYALL,
+        /**
+         * Stop to try to write bootstrap config and directly send a "Bootstrap Finished" request.
+         */
+        SEND_FINISHED,
+        /**
+         * Stop the bootstrap session. The session is considered as failure.
+         */
+        STOP
+    }
 
     /**
      * Starts a bootstrapping session for an endpoint. In particular, this is responsible for authorizing the endpoint
@@ -55,16 +77,20 @@ public interface BootstrapSessionManager {
      * 
      * @param request The request for which we get a error response.
      * @param response The response received.
+     * 
+     * @return a {@link BootstrapPolicy} given the way to continue the bootstrap session.
      */
-    public void onResponseError(LwM2mRequest<? extends LwM2mResponse> request, LwM2mResponse response);
+    public BootstrapPolicy onResponseError(LwM2mRequest<? extends LwM2mResponse> request, LwM2mResponse response);
 
     /**
      * Called when a request failed to be sent.
      * 
      * @param request The request which failed to be sent.
      * @param exception The cause of the failure. Can be null.
+     * 
+     * @return a {@link BootstrapPolicy} given the way to continue the bootstrap session.
      */
-    public void onRequestFailure(LwM2mRequest<? extends LwM2mResponse> request, Throwable cause);
+    public BootstrapPolicy onRequestFailure(LwM2mRequest<? extends LwM2mResponse> request, Throwable cause);
 
     /**
      * Performs any housekeeping related to the successful ending of a Bootstrapping session.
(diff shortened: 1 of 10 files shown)


sbernard31:
In `DefaultBootstrapHandler`, I asked myself if the executor is really needed ? I would go for remove it. :thinking: 

WDYT ?

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager.java
```diff
@@ -42,6 +43,29 @@
      */
     public BootstrapSession begin(String endpoint, Identity clientIdentity);
 
+    /**
+     * Called when we receive a successful response to a request.
+     * 
+     * @param request The request for which we get a successfull response.
+     */
+    public void onSuccess(LwM2mRequest<? extends LwM2mResponse> request);
```
msangoi: I think the method names can be confusing. e.g. onSuccess() is not about the success of the bootstrap sequence. Or onFailure() and the exisiting failed() are pretty close. 
Maybe onResponseSuccess(), onResponseError(), onRequestFailure()?

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager.java
```diff
@@ -32,6 +31,29 @@
...
+         */
+        RETRYALL,
+        /**
+         * Stop to try to write bootstrap config and directly send a "Bootstrap Finished" request.
+         */
+        SEND_FINISHED,
+        /**
+         * Stop the bootstrap session. The session is considered as failure.
+         */
+        STOP
```
msangoi: FAIL?

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager.java
```diff
@@ -32,6 +31,29 @@
  */
 public interface BootstrapSessionManager {
 
+    public enum BootstrapAction {
```
msangoi: Not really an "Action". But I cannot propose a better name for now 🙂

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/DefaultBootstrapHandler.java
```diff
@@ -236,21 +234,19 @@ protected void writeSecurities(final BootstrapSession session, final BootstrapCo
...
                         afterWriteSecurities(session, cfg, securityInstancesToWrite, action);
                     }
                 }
             }, new ErrorCallback() {
                 @Override
                 public void onError(Exception e) {
-                    LOG.debug(String.format(
-                            "Error during bootstrap write of security instance %s to foreign peer [ep=%s,id=%s]",
-                            securityInstance, session.getEndpoint(), session.getIdentity()), e);
+                    LOG.debug("Error for {} while sending {} ", session, writeBootstrapRequest, e);
```
msangoi: I think the exception parameter is going to be ignored here. e will be considered as an argument to the string format.

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/DefaultBootstrapHandler.java
```diff
@@ -135,7 +135,12 @@ public BootstrapResponse bootstrap(Identity sender, BootstrapRequest request) {
         e.execute(new Runnable() {
             @Override
             public void run() {
-                startBootstrap(session, cfg);
+                try {
+                    startBootstrap(session, cfg);
+                } catch (RuntimeException e) {
+                    LOG.warn("Unexpected error at bootstrap start-up for {}", session, e);
```
msangoi: Same comment about exception as parameter.

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager.java
sbernard31: :+1:

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager.java
sbernard31: :thinking:  ok.

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager.java
sbernard31: You have all the week-end to think about that. :wink:

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/DefaultBootstrapHandler.java
sbernard31: I pretty sure it's OK, but I will test this.

sbernard31:
```java
public static void main(String[] args) {
	LOG.error("My log with params {} {}","param1","param2", new NullPointerException());
}
```
gives
```
17:18:26.659 [main] ERROR test.LOGGER - My log with params param1 param2
java.lang.NullPointerException: null
	at test.LOGGER.main(LOGGER.java:11)
```

So it looks OK

msangoi: ok 🤔 I was sure I've had this issue in the past. anyway, thanks for testing it.

sbernard31: Maybe it depends on the slf4j backend :thinking:
I only tested with logback.

sbernard31: (it's OK for slf4j-simple too)

leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager.java
sbernard31: Did you find a better name during your week-end :grin:  ?

sbernard31:
Any thought about the executor ?

Issue #437, opened 2017-11-15
"BootstrapConfig" should be a list of request


sbernard31:
Currently the bootstrap server is configurable by adding a bootstrapConfig for a given endpoint.

This bootstrap config is a list of `Security` objects and `Server` objects.
A bootstrap session will start by a `Delete` on `/` then a `Write` request for each `Security` objects then a `Write` request for each `Server` objects.
This is not so flexible.

The bootstrapConfig should rather be a list of requests (write,delete) to send.
This allow to : 
- support all objects,
- do not force to start with delete "/" (delete only object you need to delete),
- do request in different order.

But this will not cover cases where we need to make complex processing (using discover, using previous result to guess what to send ... ... ), I don't know what this might look like.

sbernard31:
After several thoughts and tries about making the `BootstrapConfig` more flexible we finally think this is maybe not a so good Idea.

We try to implement this in #883 and as explained at https://github.com/eclipse/leshan/pull/883#issuecomment-687043565 this bring new issues ...
This makes make us feel there is no way to provide a kind of Bootstrap Config which could supports correctly all use-cases.

Those makes us think that maybe we should keep `BootstrapConfig` which aims to : 
- be easy/straightforward to use for main use cases.
- but with limited feature.

In addition, the code should be refactored to make easier to integrate custom behavior and so support all cases. 

This means that recent addition of #883 was probably a bad idea and will be removed/replaced by new classes.

sbernard31:
#1027 is a new try to achieve this goals.

sbernard31:
I think #1027 achieve this goal. 
This new API will be available in Leshan v2.0.0 since M4.

PR #718, opened 2019-07-11
Initial support for OSCORE communication between client and server

rikard-sics:
This pull request adds early support for using OSCORE to communicate between a Leshan client and server. The server can register and perform its requests towards the server using OSCORE. Likewise the server can perform requests to the client using OSCORE.

The web interface of the server has been extended to allow adding clients that use OSCORE for security including their OSCORE security context information.

On the client side further additions are needed to allow setting OSCORE security context information on it. For instance via command line arguments.

### DIFF
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
@@ -12,6 +12,7 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
+ *     Rikard Höglund (RISE SICS) - Additions to support OSCORE
  *******************************************************************************/
 package org.eclipse.leshan.server.californium.impl;
 
@@ -31,6 +32,9 @@ import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.network.Exchange;
 import org.eclipse.californium.core.server.resources.CoapExchange;
 import org.eclipse.californium.core.server.resources.Resource;
+import org.eclipse.californium.oscore.HashMapCtxDB;
+import org.eclipse.californium.oscore.OSCoreCtx;
+import org.eclipse.californium.oscore.OSException;
 import org.eclipse.leshan.Link;
 import org.eclipse.leshan.core.request.BindingMode;
 import org.eclipse.leshan.core.request.DeregisterRequest;
@@ -81,6 +85,7 @@ public class RegisterResource extends CoapResource {
 
     @Override
     public void handleRequest(Exchange exchange) {
+
         try {
             super.handleRequest(exchange);
         } catch (InvalidRequestException e) {
@@ -140,6 +145,26 @@ public class RegisterResource extends CoapResource {
     }
 
     private void handleRegister(CoapExchange exchange, Request request) {
+
+        // Check if this incoming request is using OSCORE
+        if (exchange.advanced().getRequest().getOptions().getOscore() != null) {
+            LOG.info("Client registered using OSCORE");
+
+            // Update the URI of the associated OSCORE Context with the client's URI
+            // So the server can send requests to the client
+            HashMapCtxDB db = HashMapCtxDB.getInstance();
+            OSCoreCtx clientCtx = db.getContext(exchange.advanced().getCryptographicContextID());
+
+            try {
+                db.addContext(request.getScheme() + "://"
+                        + request.getSourceContext().getPeerAddress().getHostString().toString(), clientCtx);
+            } catch (OSException e) {
+                LOG.error("Failed to update OSCORE Context for registering client.", request, e);
+                e.printStackTrace();
+            }
+
+        }
+
         // Get identity
         // --------------------------------
         Identity sender = extractIdentity(request.getSourceContext());
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityInfo.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityInfo.java
@@ -12,12 +12,16 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
+ *     Rikard Höglund (RISE SICS) - Additions to support OSCORE
  *******************************************************************************/
 package org.eclipse.leshan.server.security;
 
 import java.io.Serializable;
 import java.security.PublicKey;
 import java.util.Arrays;
+
+import org.eclipse.californium.oscore.HashMapCtxDB;
+import org.eclipse.californium.oscore.OSCoreCtx;
 
 import org.eclipse.leshan.util.Validate;
 
@@ -29,6 +33,7 @@ import org.eclipse.leshan.util.Validate;
  * <li>Pre-Shared Key: an identity and a key are needed</li>
  * <li>Raw Public Key Certificate: a public key is needed</li>
  * <li>X509 Certificate: an X509 certificate is needed</li>
+ * <li>OSCORE: an OSCORE security context is needed</li>
  * </ul>
  */
 public class SecurityInfo implements Serializable {
@@ -46,14 +51,18 @@ public class SecurityInfo implements Serializable {
 
     private final boolean useX509Cert;
 
+    // OSCORE (FIXME: Save content properly information here. Must be serializable.)
+    boolean useOSCore;
+
     private SecurityInfo(String endpoint, String identity, byte[] preSharedKey, PublicKey rawPublicKey,
-            boolean useX509Cert) {
+            boolean useX509Cert, OSCoreCtx oscoreCtx) {
         Validate.notEmpty(endpoint);
         this.endpoint = endpoint;
         this.identity = identity;
         this.preSharedKey = preSharedKey;
         this.rawPublicKey = rawPublicKey;
         this.useX509Cert = useX509Cert;
+        this.useOSCore = oscoreCtx != null;
     }
 
     /**
@@ -62,7 +71,7 @@ public class SecurityInfo implements Serializable {
     public static SecurityInfo newPreSharedKeyInfo(String endpoint, String identity, byte[] preSharedKey) {
         Validate.notEmpty(identity);
         Validate.notNull(preSharedKey);
-        return new SecurityInfo(endpoint, identity, preSharedKey, null, false);
+        return new SecurityInfo(endpoint, identity, preSharedKey, null, false, null);
     }
 
     /**
@@ -70,14 +79,29 @@ public class SecurityInfo implements Serializable {
      */
     public static SecurityInfo newRawPublicKeyInfo(String endpoint, PublicKey rawPublicKey) {
         Validate.notNull(rawPublicKey);
-        return new SecurityInfo(endpoint, null, null, rawPublicKey, false);
+        return new SecurityInfo(endpoint, null, null, rawPublicKey, false, null);
     }
 
     /**
      * Construct a {@link SecurityInfo} when using DTLS with an X509 Certificate.
      */
     public static SecurityInfo newX509CertInfo(String endpoint) {
-        return new SecurityInfo(endpoint, null, null, null, true);
+        return new SecurityInfo(endpoint, null, null, null, true, null);
+    }
+
+    /**
+     * Construct a {@link SecurityInfo} when using OSCORE.
+     */
+    public static SecurityInfo newOSCoreInfo(String endpoint, String identity, OSCoreCtx oscoreCtx) {
+        Validate.notEmpty(identity);
+        Validate.notNull(identity);
+        Validate.notNull(oscoreCtx);
+
+        // Add the OSCORE Context to the context database
+        HashMapCtxDB db = HashMapCtxDB.getInstance();
+        db.addContext(oscoreCtx);
+
+        return new SecurityInfo(endpoint, identity, null, null, false, oscoreCtx);
     }
 
     public String getEndpoint() {
--- a/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
+++ b/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
@@ -43,6 +43,7 @@ import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.oscore.OSCoreCoapStackFactory;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig.Builder;
 import org.eclipse.californium.scandium.dtls.CertificateMessage;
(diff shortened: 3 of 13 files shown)


rikard-sics:
To test a scenario with the client and server using OSCORE perform the following steps:
1. Start the LeshanServerDemo. In its web interface add an OSCORE client with the following information:
Client endpoint: test
Master Secret: 11223344
Sender ID: AA
Recipient ID: BB

2. Start the LeshanClientDemo with argument "-oscore".

3. The client should now register using OSCORE (and in general perform requests to the server using OSCORE).

4. Likewise when performing requests for resources on the client from the server web interface OSCORE will be used.

5. It can be seen in Wireshark that the messages exchanged are CoAP but with option 9 and their payload is encrypted.

rikard-sics:
Some things to add in coming pull requests are the ability to indicate in command line parameters to the client what OSCORE security context information to use, and serialization of the OSCORE security context information.

sbernard31:
I will not be available the next 2 weeks but I look at this as soon as I'm back.

rikard-sics:
> I will not be available the next 2 weeks but I look at this as soon as I'm back.

Thanks for the update, sounds good.

sbernard31:
Ok I finally take time to look at :
 - [OSCORE](https://tools.ietf.org/html/draft-ietf-core-object-security-16)
 - [OSCORE in LWM2M 1.1](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html) 
 - and the current PR #718.

All of this is probably not perfectly clear for me (and I didn't follow PR/discussion about that at Californium), so maybe some of my remarks doesn't make sense :sweat_smile:

I will start by high-level remarks.

### 1) I'm a bit disturbed by the singleton `HashMapCtxMap
FMPOV Singleton pattern has to many drawback, we can discuss about that if needed.
I think the idea should be more to have a kind of dynamic store like [PskStore in scandium](https://github.com/eclipse/californium/blob/2.0.0-M16/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/PskStore.java).
I see you created a `OSCoreCtxDB` interface maybe you already have this in mind and singleton is just a first step. :thinking: 

1 of the benefits to define a store would be that we could have 1 implementation at server side based on Leshan `SecurityStore` and 1 implementation at client side which will search context data in client LWM2M OSCORE object.

(by the way, I'm a bit traumatized by memory leak issues and all those map in HashMapCtxMap scarred me a bit :grin:)

### 2) persistent and volatile data in OSCORE context
1 concern : We need to search in store for each request/response, with a persistent store this will maybe produce some performance issue.

There is probably some part of the context which must be persist (and  so go in a store)  and some other just keep in memory :thinking: 
But if we don't persist the whole context we need to handle  "losing Part of the Context State" and this sounds not so easy :sweat_smile: 

Anyway this is maybe premature concern.
### 3) OSCORE + DTLS
About CoAP/OSCORE, I understand that you can use OSCORE+DTLS at the same time.
At LWM2M level, I think this is possible to use both too. So OSCORE/DTLS config will not be exclusive.

### So what we do now ?
I feel there is a lot of work at Californium and Leshan side.
Generally, I prefer integrate small functional consistent PR but here I feel this will not be so easy as we don't really know where we go. (maybe I'm the only one a bit lost here :sweat_smile: )
From what I see, the current PR is a kind of POC, right ? So how would you like to proceed ?
Do you want I review this PR more in detail, then see it integrated in a Leshan branch or do you have another plan ?

1. we can create an oscore branch waiting we have minimum viable feature.
2. or directly push it in branch 2.0.0

Generally I prefer the 1. solution but as 2.0.0 is very new and pretty much dedicated for you(waiting I deliver a 1.0.0)  2. is fine for me too.

rikard-sics:
Thanks for the feedback. Some quick initial reactions from my side. Let me get back with a longer reply shortly.

> 1) I'm a bit disturbed by the singleton `HashMapCtxMap

Yes and similar feedback was given by the Californium developers. This was changed so that the HashMapCtxDB is no longer a singleton anymore. It was changed in Californium PR https://github.com/eclipse/californium/pull/1000


> 2) persistent and volatile data in OSCORE context
> But if we don't persist the whole context we need to handle "losing Part of the Context State" and this sounds not so easy sweat_smile

In fact the OSCORE RFC defines a way to securely re-generate a context if the mutable parts of it has been lost. See 
https://tools.ietf.org/html/rfc8613#appendix-B.2

I am also in the process of adding this functionality to Californium, having done 2 PRs regarding this:
https://github.com/eclipse/californium/pull/957
https://github.com/eclipse/californium/pull/983

rikard-sics:
Again, thanks for the feedback on the PR. Further replies from my side follow.

> ### 1) I'm a bit disturbed by the singleton `HashMapCtxMap
> FMPOV Singleton pattern has to many drawback, we can discuss about that if needed.
> I think the idea should be more to have a kind of dynamic store like [PskStore in scandium](https://github.com/eclipse/californium/blob/2.0.0-M16/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/PskStore.java).
> I see you created a `OSCoreCtxDB` interface maybe you already have this in mind and singleton is just a first step. 
> 
> 1 of the benefits to define a store would be that we could have 1 implementation at server side based on Leshan `SecurityStore` and 1 implementation at client side which will search context data in client LWM2M OSCORE object.
> (by the way, I'm a bit traumatized by memory leak issues and all those map in HashMapCtxMap scarred me a bit )

As I mentioned the HashMapCtxDB is no longer a singleton in the latest Californium code. It is now possible to indicate which context DB to use when building an endpoint. So for instance on the server the client->server endpoint accepting messages from the client can use one context DB. While the server->client endpoint when the server is making requests to the client can use a different context DB.

We could also explore alternative ways of storing the OSCORE contexts in ways that may fit better with the existing Leshan architecture. Although the current approach in Californium relies on the use of a HashMapCtxDB to hold the context information. Depending on if sending/receiving and if it is a request/response the correct context will be retrieved from the appropriate map in the context DB.


> ### 2) persistent and volatile data in OSCORE context
> 1 concern : We need to search in store for each request/response, with a persistent store this will maybe produce some performance issue.
> 
> There is probably some part of the context which must be persist (and so go in a store) and some other just keep in memory 
> But if we don't persist the whole context we need to handle "losing Part of the Context State" and this sounds not so easy 
> 
> Anyway this is maybe premature concern.

I think this is a fair concern. This should hopefully be resolved by the functionality from OSCORE to re-derive contexts if parts of them have been lost. As I mentioned that functionality is being added to Californium.


> ### 3) OSCORE + DTLS
> About CoAP/OSCORE, I understand that you can use OSCORE+DTLS at the same time.
> At LWM2M level, I think this is possible to use both too. So OSCORE/DTLS config will not be exclusive.

Yes, that is true. In this PR the configuration is exclusive but it should be more flexible to allow for both OSCORE and DTLS being used at the same time.

> ### So what we do now ?
> I feel there is a lot of work at Californium and Leshan side.
> Generally, I prefer integrate small functional consistent PR but here I feel this will not be so easy as we don't really know where we go. (maybe I'm the only one a bit lost here  )

Yes it may be a little tricky splitting the functionality into very small parts. I had some high level goals in mind including:

- OSCORE communication between client and server
- OSCORE communication between client and bootstrap server
- Client receiving OSCORE security information from the bootstrap server

Of course there are a lot of detailed concerns and questions on how the OSCORE code can fit best into the current way Leshan does things.

> From what I see, the current PR is a kind of POC, right ? So how would you like to proceed ?
> Do you want I review this PR more in detail, then see it integrated in a Leshan branch or do you have another plan ?
> 
> 1. we can create an oscore branch waiting we have minimum viable feature.
> 2. or directly push it in branch 2.0.0
> 
> Generally I prefer the 1. solution but as 2.0.0 is very new and pretty much dedicated for you(waiting I deliver a 1.0.0) 2. is fine for me too.

Yes, I would like it if this PR could be reviewed further and integrated in a Leshan branch. Creating a separate OSCORE branch is fine for me. From there OSCORE can then be integrated into the 2.0.x branch at a later stage.

sbernard31:
For this part :
> Client receiving OSCORE security information from the bootstrap server

Currently bootstrap config is not so inflexible, this feature #437 would help for writing OSCORE object during a bootstrap session.

#437 is not planned for Leshan 1.0.0, but it makes senses to add it to Leshan 2.0.0.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
```diff
@@ -126,6 +127,9 @@ public static void main(String[] args) {
         // Define options for command line tools
         Options options = new Options();
 
+        // Enable OSCORE stack (fine to do even when using DTLS or only CoAP)
+        OSCoreCoapStackFactory.useAsDefault();
```
sbernard31: In the end this should rather be done in LeshanServerBuilder than in Demo.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/servlet/json/SecuritySerializer.java
```diff
@@ -42,7 +42,9 @@ public JsonElement serialize(SecurityInfo src, Type typeOfSrc, JsonSerialization
         if (src.getIdentity() != null) {
             JsonObject psk = new JsonObject();
             psk.addProperty("identity", src.getIdentity());
-            psk.addProperty("key", Hex.encodeHexString(src.getPreSharedKey()));
+            if (src.getPreSharedKey() != null) {
+                psk.addProperty("key", Hex.encodeHexString(src.getPreSharedKey()));
+            }
```
sbernard31: This is out of scope of this PR. Please open another PR or issue if you think this is a bug.
But to be honest, I'm not sure we want to do that :thinking:

leshan-server-demo/src/main/resources/webapp/partials/security-list.html
```diff
@@ -101,6 +112,8 @@ <h4 class="modal-title">New security configuration</h4>
 								<option value="psk">Pre-Shared Key</option>
 								<option value="rpk">Raw Public Key (Elliptic Curves)</option>
 								<option value="x509">X.509 Certificate</option>
+								<!-- Drop-down option for OSCORE security context information -->
```
sbernard31: this comment in maybe a bit overkill :)

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/servlet/json/SecurityDeserializer.java
```diff
@@ -102,6 +109,77 @@ public SecurityInfo deserialize(JsonElement json, Type typeOfT, JsonDeserializat
                 info = SecurityInfo.newRawPublicKeyInfo(endpoint, key);
             } else if (x509 != null && x509.getAsBoolean()) {
                 info = SecurityInfo.newX509CertInfo(endpoint);
+            } else if (oscore != null) {
+                // OSCORE Deserialization
+
+                // Parse hexadecimal context parameters
+                byte[] masterSecret = Hex.decodeHex(oscore.get("masterSecret").getAsString().toCharArray());
+                byte[] senderId = Hex.decodeHex(oscore.get("senderId").getAsString().toCharArray());
+                byte[] recipientId = Hex.decodeHex(oscore.get("recipientId").getAsString().toCharArray());
```
sbernard31: should we raise JSON exception is this mandatory field are absent ?

leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityInfo.java
```diff
@@ -62,22 +71,37 @@ private SecurityInfo(String endpoint, String identity, byte[] preSharedKey, Publ
...
+     * Construct a {@link SecurityInfo} when using OSCORE.
+     */
+    public static SecurityInfo newOSCoreInfo(String endpoint, String identity, OSCoreCtx oscoreCtx) {
+        Validate.notEmpty(identity);
+        Validate.notNull(identity);
+        Validate.notNull(oscoreCtx);
+
+        // Add the OSCORE Context to the context database
+        HashMapCtxDB db = HashMapCtxDB.getInstance();
+        db.addContext(oscoreCtx);
```
sbernard31: Here this is a problem for me.
Create a new securityInfo should not modify the state of the application.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/RegisterResource.java
```diff
@@ -140,6 +145,26 @@ public void handleDELETE(CoapExchange exchange) {
...
+            try {
+                db.addContext(request.getScheme() + "://"
+                        + request.getSourceContext().getPeerAddress().getHostString().toString(), clientCtx);
+            } catch (OSException e) {
+                LOG.error("Failed to update OSCORE Context for registering client.", request, e);
+                e.printStackTrace();
+            }
+
+        }
+
```
sbernard31: This part is strange to me too but this is probably mainly because all of this is not clear to me. So we should probably discuss a bit about what could looks like the OSCORE integration in Leshan. (I will open an issue)

sbernard31: let's discuss about that here https://github.com/eclipse/leshan/issues/727

rikard-sics:
> Currently bootstrap config is not so inflexible, this feature #437 would help for writing OSCORE object during a bootstrap session.
> 
> #437 is not planned for Leshan 1.0.0, but it makes senses to add it to Leshan 2.0.0.

Yes, that looks useful. It should help in making the bootstrapping with OSCORE security information simpler.

> So let's go with a new oscore branch. ;)
> 
> As this is not the final step, it's ok for me if you don't take this remarks into account now. We can even integrate this PR like this. I feel the priority is more to explore the right way to integrate oscore in Leshan (see #725) ! (we could clean the code later).

Thanks for the detailed feedback. Yes it could make sense to simply integrate the code as-is into the oscore-branch for now (except for very critical things to change). 

I agree that exploring the right way to integrate things can be the best approach right now. We may have to re-design things in the OSCORE code as we go, so the code in this current PR may be changed significantly throughout this process.

I could open a new PR with this code towards an oscore branch. (Or perhaps re-point this PR to an oscore branch?) I could also then try to fix some of the smaller things from your feedback, but wait with bigger changes until after further discussion and redesigns.

sbernard31:
> I could open a new PR with this code towards an oscore branch. (Or perhaps re-point this PR to an oscore branch?) 

Ok I created the oscore branch and now this PR target it.

> I could also then try to fix some of the smaller things from your feedback, but wait with bigger changes until after further discussion and redesigns.

It's not clear to me  : do you want to add some modification to this PR or should I merge it like this ?

rikard-sics:
> It's not clear to me : do you want to add some modification to this PR or should I merge it like this ?

Could you merge it like this? Then I will proceed with the changes you mentioned in this thread and others at a later stage.

sbernard31:
Done :)

rikard-sics:
> Ok I created the oscore branch and now this PR target it.

Thanks!

> Done :)

Thanks a lot. Then perhaps my next step will be to start adding functionality on the client to take command line arguments with OSCORE security information instead of using the hardcoded material the client is using now. After that the next step can be to actually build an OSCORE object from the parameters.

sbernard31:
I think going from hardcoded value to command line arguments would not raise too many issue, so maybe this is not the higher priority. :thinking: 

Maybe you could keep the hardcoded one for now and focus more on : 
 - create the OSCORE object 
 - use it in leshan client demo (keeping hardcoded value but create oscore object)
 - modify ServersInfoExtractor and ServerInfo to support oscore.
 - use ServerInfo in CaliforniumEndpointsManager to create endpoint with oscore support.

rikard-sics:
> I think going from hardcoded value to command line arguments would not raise too many issue, so maybe this is not the higher priority. thinking

That is true, then I will rather proceed with the steps you outline.

PR #721, opened 2019-07-19
 Add BootstrapSession param to BootstrapConfigStore.get()

sbernard31:
It could make sense to store data in a session at `BootstrapSessionManager.begin(String, Identity)` to reuse it in `BootstrapConfigStore.get()`


leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapConfigStore.java
```diff
@@ -27,7 +27,8 @@
      * 
      * @param endpoint the endpoint of the device.
      * @param deviceIdentity the {@link Identity} the device.
```
msangoi: Should be renamed in the javadoc as well.

sbernard31: the parameter renaming was not intended. I revert it (sender => deviceIdentity)

PR #722, opened 2019-07-22
Add support for uploading file as a value for opaque resource

ukasiu:
I'd like to test Firmware Upgrade in push mode with Leshan and so I implement a way to upload a file and send it as an opaque value for resource.

Signed-off-by: Łukasz Gurdek <eclipse@ukasiu.pl>

### DIFF
--- a/leshan-server-demo/pom.xml
+++ b/leshan-server-demo/pom.xml
@@ -68,6 +68,11 @@ Contributors:
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.ooxi</groupId>
+            <artifactId>jdatauri</artifactId>
+            <version>1.0</version>
+        </dependency>
 
         <!-- runtime dependencies -->
        <dependency>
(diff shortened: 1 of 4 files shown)


sbernard31:
Hi @ukasiu, thx for this ! :pray: 

What is the benefits of using dataURI ? if we could I would prefer to not add a new format and a new dependency (jdatauri)

ukasiu:
Hi @sbernard31, 
`readAsDataURL` was to me the most straightforward method to convert file to base64 in the frontend. I could probably use `readAsBinaryString` and `btoa` and thus avoid `jdatauri` in backend. Would that be ok?

sbernard31:
I looked at this a more deeper and currently the way we expect to receive OPAQUE value is HEXA.
Look at : [DefaultLwM2mValueConverter](https://github.com/eclipse/leshan/blob/leshan-1.0.0-M11/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/DefaultLwM2mValueConverter.java#L134)

So if you sent the value in HEXA there is nothing to do at java side.

I agree that sending data in HEXA is not efficient. 
We started to use HEXA  as it sounds better for reability/debugging... (and as demo is just a demo we let aside the efficiency)

Do you think it's ok to use HEXA ? ([here](https://github.com/eclipse/leshan/blob/leshan-1.0.0-M11/leshan-server-demo/src/main/resources/webapp/js/security-controllers.js#L25) we already to something like this)

sbernard31:
@ukasiu, do you plan to work more on this PR ?

ukasiu:
Yes, just give me a week or two if possible

sbernard31:
@ukasiu, I re-implement this feature in #777.

Could you try if this works for you ! After your feedback I will integrate this in master.

sbernard31:
Without any news, I decide to integrate #777 in master.
As #777 obsoletes this PR, I close it.
Please, do no hesitate to share any feedback about.

PR #731, opened 2019-08-27
Update to Californium 2.0.0-M15 and fix for OSCORE Observe support

rikard-sics:
I was informed by a person we are collaborating with that the observe functionality was not working when using OSCORE in Leshan.

This PR fixes this by updating the Californium version to 2.0.0-M15 where observe support was added.

It also adds code to fix an issue where the initial observe request does not have a Token set in the code when using OSCORE in Leshan. However when looking in Wireshark the request has a Token set when actually being transmitted. This could potentially be because OSCORE re-creates the request and the old one is still used in Leshan somehow. This PR solves this issue but I need to understand the root cause better to create a more permanent fix.


sbernard31:
It sounds good to me.
mass conflict in perspective when I will to rebase 2.0.0 branch on master(1.0.0) :grin: 
(I will separate this in 2 commits the californium integration and the little oscore fix, so cherry picking would be easier)

sbernard31:
Ok I integrated this in oscore branch
- commit ee9deed3566eb451af590bbb3dd118d4ca258f84
- commit  9d9e42e8cadceb96becdf7f4e8c14ae8aac46ca1

About : 
```java
/*
 * Note: When using OSCORE and Observe the first coapRequest sent to register an observation can have
 * its Token missing here. Is this because OSCORE re-creates the request before sending? When looking in
 * Wireshark all messages have a Token as they should. The lines below fixes this by taking the Token
 * from the response that came to the request (since the request actually has a Token when going out the
 * response will have the same correct Token.
 * 
```
I understand that as a workaround but maybe I'm wrong.
Is it mandatory to create a new request ? I suppose it is. if I'm right,  is it possible to update the original request to set the token as the oscore one ? (I don't know if this is a good idea)


(I saw some formatting issue, you could [look at this](https://github.com/eclipse/leshan/wiki/Code-&-design-guidelines#configure-your-ide) to avoid that)

rikard-sics:
> It sounds good to me.
> mass conflict in perspective when I will to rebase 2.0.0 branch on master(1.0.0) 
> (I will separate this in 2 commits the californium integration and the little oscore fix, so cherry picking would be easier)

> Ok I integrated this in oscore branch
> 
> * commit [ee9deed](https://github.com/eclipse/leshan/commit/ee9deed3566eb451af590bbb3dd118d4ca258f84)
> * commit  [9d9e42e](https://github.com/eclipse/leshan/commit/9d9e42e8cadceb96becdf7f4e8c14ae8aac46ca1)

Thanks for separating the code and integrating it.

> I understand that as a workaround but maybe I'm wrong.
> Is it mandatory to create a new request ? I suppose it is. if I'm right, is it possible to update the original request to set the token as the oscore one ? (I don't know if this is a good idea)
> 
> (I saw some formatting issue, you could [look at this](https://github.com/eclipse/leshan/wiki/Code-&-design-guidelines#configure-your-ide) to avoid that)

It may be an option to change things in Californium to not create a new request. Another option is like you say to update the original request with the Token.

Issue #726, opened 2019-08-12
OSCORE over coap at client side


sbernard31:
Here we discuss about how integrate OSCORE over coap at client side.

Some ideas I have in mind :
The OSCORE context must be retrieve from OSCORE object so setting OSCORE context should looks like : 
```java
long oscoreInstanceId;
initializer.setInstancesForObject(21, new OSCORE(oscoreInstanceId, master, sid, rid, ....);
initializer.setInstancesForObject(SECURITY, Security.oscore(serverURI, 123, oscoreInstanceId));
initializer.setInstancesForObject(SERVER, new Server(123, 30, BindingMode.U, false));
```
Then in `CaliforniumEndpointsManager.createEndpoint()`, if the client should use oscore for the given server, a endpoint is created with support of oscore and with corresponding oscore context a bit like for PSK. (see https://github.com/eclipse/leshan/blob/leshan-1.0.0-M11/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java#L89) 

(@rikard-sics tell me if this is make sense as I don't know so well the OSCORE RFC I maybe missed something.)

rikard-sics:
Yes it makes sense to create the OSCORE context when the actual endpoint is created (using information from the OSCORE object). When the endpoint is being created the OSCORE context will be created and that endpoint will then use the OSCORE DB where this context is stored.

Every outgoing request sent with that endpoint should also have the OSCORE flag set. For incoming requests they are parsed with OSCORE automatically and the correct context will be found from the information in the OSCORE option. For incoming responses the correct OSCORE context is found from the CoAP Token (when sending a request the Token is saved associated to the context used).

Right now the OSCORE Object is not created on the client. Excluding the bootstrap server for now, the best solution would then be for the client to take OSCORE security information as command line parameters and create an OSCORE object from them?

sbernard31:
> Right now the OSCORE Object is not created on the client. Excluding the bootstrap server for now, the best solution would then be for the client to take OSCORE security information as command line parameters and create an OSCORE object from them?

Yep like PSK :-)

rikard-sics:
> Yep like PSK :-)

Okay, makes sense :)

rikard-sics:
> I will proceed according to the steps you outlined here:
> [#718 (comment)](https://github.com/eclipse/leshan/pull/718#issuecomment-521316754)

[...]

Issue #727, opened 2019-08-12
OSCORE over coap at server side


sbernard31:
Here we discuss about how integrate OSCORE over coap at server side.

Some ideas I have in mind :
A Leshan server user want to say that a device with the endpoint "mydevice" should communicate in OSCORE. So he creates a new SecurityInfo("mydevice", oscore indentifier ....) and I add it to the security store.

When a device use OSCORE, we could image that this is visible in endpoint context. From this we create the corresponding `identity`, then the authorizer is responsible to check if this device should connect itself using oscore and if it use the right oscore parameter (looking in security store).

For sending request we should also put pertinent data in endpointcontext like we do [here](https://github.com/eclipse/leshan/blob/leshan-1.0.0-M11/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CoapRequestBuilder.java#L197)

The idea is to make pretty much the same as we did for DTLS with PSK.

(@rikard-sics tell me if it makes senses)

rikard-sics:
> The idea is to make pretty much the same as we did for DTLS with PSK.

Yes I think the points you outline make a lot of sense. Especially trying to make the OSCORE functionality mirror how things are done for DTLS PSK.

rikard-sics:
I have now created a PR for setting an identity and matching the identity when using OSCORE here https://github.com/eclipse/leshan/pull/766

One thought I had was that long-term we could have something like the PskStore but for OSCORE contexts, rather than the OscoreHandler singleton-style used now.

rikard-sics:
I have now created a PR to add support for OSCORE communication to the bootstrap server also. See https://github.com/eclipse/leshan/pull/857

rikard-sics:
I have now created a PR https://github.com/eclipse/leshan/pull/919 to enable the client bootstrapping to receive OSCORE security material.

sbernard31:
With https://github.com/eclipse/leshan/pull/1277 integrated in `master` and missing work listed at https://github.com/eclipse/leshan/issues/725#issuecomment-1179021294.

Maybe we can close this one  too ? 
@rikard-sics WDYT ?

rikard-sics:
> Maybe we can close this one too ?
> @rikard-sics WDYT ?

Yeah, I think so. There are other issues we can continue the discussions in.

PR #736, opened 2019-09-12
Generate OSCORE object and use it for endpoint creation

rikard-sics:
Code to implement the points listed here:
https://github.com/eclipse/leshan/pull/718#issuecomment-521316754
- create the OSCORE object
- use it in leshan client demo (keeping hardcoded value but create oscore object)
- modify ServersInfoExtractor and ServerInfo to support oscore.
- use ServerInfo in CaliforniumEndpointsManager to create endpoint with oscore support.

The OSCORE object is generated and filled in the Leshan client demo. It will then be used to create a ServerInfo object using the ServersInfoExtractor. Next the ServerInfo is used in the CaliforniumEndpointsManager to generate endpoints towards servers that OSCORE should be used with.

See also issue regarding OSCORE Leshan client integration:
https://github.com/eclipse/leshan/issues/726


leshan-client-core/pom.xml
```diff
@@ -34,6 +35,12 @@ Contributors:
             <groupId>org.eclipse.leshan</groupId>
             <artifactId>leshan-core</artifactId>
         </dependency>
+	<!-- Dependency for OSCORE support -->
+        <dependency>
+                <groupId>org.eclipse.californium</groupId>
+                <artifactId>cf-oscore</artifactId>
+                <version>${californium.version}</version>
+        </dependency>
```
sbernard31: `leshan-client-code` must not depend to californium. (see https://github.com/eclipse/leshan/wiki/Modules#leshan-client-core)

rikard-sics: I see, thanks for the link. I will amend the code so leshan-client-core does not have any dependencies from Californium. I can probably keep it as an int instead of AlgorithmID.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -398,6 +372,10 @@ public static void createAndStartClient(String endpoint, String localAddress, in
                 initializer.setInstancesForObject(SECURITY, x509(serverURI, 123, clientCertificate.getEncoded(),
                         clientPrivateKey.getEncoded(), serverCertificate.getEncoded()));
                 initializer.setInstancesForObject(SERVER, new Server(123, 30, BindingMode.U, false));
+            } else if (useOSCore) {
+                initializer.setInstancesForObject(SECURITY, secOSCore(serverURI, 123));
+                initializer.setInstancesForObject(OSCORE, Oscore.set("11223344", "AA", "BB")); //Hardcoded values
+                initializer.setInstancesForObject(SERVER, new Server(123, 30, BindingMode.U, false));
```
sbernard31: Regarding your remarks here : 
> One issue I had was how exactly to link the Security and Oscore objects together. I now see your suggestion here: #726 (comment)

To be able to do that, you will probably need to add a new XML version of [security object](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#Table-E1-1-LwM2M-Object-LWM2M-Security-object-definition).

You should get it [here](http://www.openmobilealliance.org/wp/OMNA/LwM2M/LwM2MRegistry.html)

leshan-core/src/main/java/org/eclipse/leshan/SecurityMode.java
```diff
@@ -12,14 +12,15 @@
...
+ *     Rikard Höglund (RISE SICS) - Additions to support OSCORE
  *******************************************************************************/
 package org.eclipse.leshan;
 
 /**
  * The different DTLS security modes
  */
 public enum SecurityMode {
-    PSK(0), RPK(1), X509(2), NO_SEC(3);
+    PSK(0), RPK(1), X509(2), NO_SEC(3), OSCORE(4);
```
sbernard31: Regarding the [specification](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#Table-E1-1-LwM2M-Object-LWM2M-Security-object-definition), 4 is for _"Certificate mode with EST"_ and there is nothing for OSCORE :thinking: .

(It is not excluded there is issue in the specification but for now, I don't understand all of this enough to say if this is an issue :grin:)

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/ServersInfoExtractor.java
```diff
@@ -112,6 +122,16 @@ public static ServersInfo getInfo(Map<Integer, LwM2mObjectEnabler> objectEnabler
                         info.clientCertificate = getClientCertificate(security);
                         info.serverCertificate = getServerCertificate(security);
                         info.privateKey = getPrivateKey(security);
+                    } else if (info.secureMode == SecurityMode.OSCORE) {
```
sbernard31: as there is no OSCORE securityMode maybe we can just look if the resource OSCORE Security Mode(17) is set.

If this is set you get the right index for the FIXME below :)

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Oscore.java
```diff
@@ -32,78 +44,140 @@
...
+    
+    /**
+     * Returns a new OSCORE instance (less parameters and using default values).
+     */
+    public static Oscore set(String masterSecret, String senderId, String recipientId) {
+        int defaultAeadAlgorithm = 10; //AES_CCM_16_64_128
+        int defaultHmacAlgorithm = -10; //HKDF_HMAC_SHA_256
+        
+        return new Oscore(masterSecret, senderId, recipientId, defaultAeadAlgorithm, defaultHmacAlgorithm, "");
+    }
```
sbernard31: What is a benefits of both static "constructor" above ?

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Security.java
```diff
@@ -108,6 +109,14 @@ public static Security noSec(String serverUri, int shortServerId) {
         return new Security(serverUri, false, SecurityMode.NO_SEC.code, new byte[0], new byte[0], new byte[0],
                 shortServerId);
     }
+    
+    /**
+     * Returns a new security instance (OSCORE) for a device management server.
+     */
+    public static Security secOSCore(String serverUri, int shortServerId) {
```
sbernard31: here this would be more something like :
- oscore (or oscoreOnly)

and one day we will add something like this ? 
- pskWithOscore 
- rpkWithOscore
- x509WithOscore

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointFactory.java
```diff
@@ -28,6 +30,8 @@
 public interface EndpointFactory {
 
     CoapEndpoint createUnsecuredEndpoint(InetSocketAddress address, NetworkConfig coapConfig, ObservationStore store);
+    
+    CoapEndpoint createOSCoreEndpoint(InetSocketAddress address, NetworkConfig coapConfig, ObservationStore store, HashMapCtxDB db);
```
sbernard31: for now my feeling is that there no need to add createOSCoreEndpoint maybe just add the db parameter to both other methods (but as the californium API is not yet finished, I maybe missed something)

leshan-client-core/pom.xml
rikard-sics: This should now be taken care of in commit 8fc7bfcdb67d219f712c5f2a1536a2d4c692e860
I needed the hex2ByteArray method from StringUtil in Californium. For now I extracted and took the code from there. If that is not fine I can re-implement it.

sbernard31: maybe you could find what you need in `org.eclipse.leshan.util.Hex` ?

rikard-sics: > maybe you could find what you need in `org.eclipse.leshan.util.Hex` ?

Yes, I will check that.

rikard-sics:
> I suppose we should add a new attribute for OSCORE which could be used in addition of existing DTLS identity (in case of OSCORE over coaps) or without it (in case of OSCORE over coap).

Yes, that makes sense.

> In oscore what could be the value which could be used as an identifier ? I need to read the rfc again (intuitively I suppose that Sender ID or/and Recipient ID could be candidate?)

A combination of the sender ID, recipient ID and ID context should identify a pair of endpoints communicating using OSCORE.

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Security.java
rikard-sics: Yes, that makes sense. I can rename the current method to oscore/oscoreOnly. And then we proceed with adding methods later that combine DTLS+OSCORE.

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Oscore.java
sbernard31: You can create several constructors with different number of parameter.

(static constructor is a good way to name a constructor but the con is that you can not reused when you extend the class)

rikard-sics: > You can create several constructors with different number of parameter.
> 
> (static constructor is a good way to name a constructor but the con is that you can not reused when you extend the class)

Okay, so it would be better to have 1 or several normal constructors, but no static constructors for the Oscore class?

sbernard31: Yep, as it seems you don't really need to "name" your constructor, I think it's better to have several "normal" constructors.

rikard-sics: Removed the static "constructors" in commit dc37bd9c5a597cb41ca91238edbfe76665b72c76

leshan-client-core/pom.xml
rikard-sics: Update to use a method from the Leshan Hex class rather than the imported method from Californium in commit ebc728a69129645fe1347e312c1feb34ee4aae13

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Security.java
rikard-sics: Renamed the current method to oscoreOnly in commit dd58e43e1da396aa35c54a24b898752647c04bdd . In the future further methods can be added.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
rikard-sics: The security object xml has been updated and a link to the OSCORE object introduced in commit ecebdfa09ce8028fa02d0d26c448ce3029069479

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -373,8 +373,10 @@ public static void createAndStartClient(String endpoint, String localAddress, in
                         clientPrivateKey.getEncoded(), serverCertificate.getEncoded()));
                 initializer.setInstancesForObject(SERVER, new Server(123, 30, BindingMode.U, false));
             } else if (useOSCore) {
-                initializer.setInstancesForObject(SECURITY, oscoreOnly(serverURI, 123));
-                initializer.setInstancesForObject(OSCORE, new Oscore("11223344", "AA", "BB")); //Hardcoded values
+            	Oscore oscoreObject = new Oscore("11223344", "AA", "BB"); //Hardcoded values
+            	oscoreObject.setId(12345);
```
sbernard31: if you want you can create an Oscore constructor which take an instanceID in parameter

rikard-sics: Yes, I will do that instead.

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/ServersInfoExtractor.java
rikard-sics: Yes, I am no longer checking the security mode here to decide what to do but rather if the link from the Security object has been set. I am also using the link from the Security object to find the OSCORE object here. Done in commit 3e7db26f5131ad4ef4622ae864e69d5796b3eefa

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
rikard-sics: Now done in commit ae73ab25b69513f1b63b8bf4e2c9fbc4ca931a50

sbernard31:
You should have a look at [§5.5.5 - Endpoint Client Name](http://www.openmobilealliance.org/release/LightweightM2M/V1_1-20180710-A/HTML-Version/OMA-TS-LightweightM2M_Transport-V1_1-20180710-A.html#5-5-5-0-555-Endpoint-Client-Name) about you "identity" concern.

About that is sender ID / receiver ID available in californium endpoint context ?

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Oscore.java
```diff
@@ -64,8 +64,9 @@ public Oscore() {
     /**
      * Default constructor.
      */
-    public Oscore(String masterSecret, String senderId, String recipientId, int aeadAlgorithm, int hkdfAlgorithm, String masterSalt) {
-        this.masterSecret = masterSecret;
+    public Oscore(int instanceId, String masterSecret, String senderId, String recipientId, int aeadAlgorithm, int hkdfAlgorithm, String masterSalt) {
+        this.setId(instanceId);
```
sbernard31: clearly not a big deal but you can use `super(id)`

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/ServersInfoExtractor.java
```diff
@@ -110,7 +110,34 @@ public static ServersInfo getInfo(Map<Integer, LwM2mObjectEnabler> objectEnabler
...
-                    if (info.secureMode == SecurityMode.PSK) {
+
+                    // find instance id of the associated oscore object (if any)
+                    ObjectLink oscoreObjectLink = (ObjectLink) security.getResource(SEC_OSCORE_SECURITY_MODE).getValue();
+                    int oscoreObjectInstanceId = oscoreObjectLink.getObjectInstanceId();
+                    int oscoreObjectId = oscoreObjectLink.getObjectId();
+                    boolean useOscore = false;
+                    if(oscoreObjectId == OSCORE) {
+                    	useOscore = true;
+                    }
```
sbernard31: boolean useOscore = oscoreObjectId == OSCORE ?

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointFactory.java
rikard-sics: Yes I removed this new method and added the db parameter to the others instead. Done in commit 495d5019f750ffee544ce7a66bb27c270e372de7

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Oscore.java
rikard-sics: True, I will change that as part of the next commit.

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/ServersInfoExtractor.java
rikard-sics: Yeah, looks more compact and nicer. I will change it.

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/ServersInfoExtractor.java
```diff
@@ -110,7 +110,34 @@ public static ServersInfo getInfo(Map<Integer, LwM2mObjectEnabler> objectEnabler
...
+
+                    if(useOscore) {
+                    	// search corresponding oscore object
+                    	LwM2mObjectInstance oscoreInstance = null;
+                        for (LwM2mObjectInstance oscore : oscores.getInstances().values()) {
```
sbernard31: Not need to iterate, there is a getter by Id => `oscores.getInstance(id)`

rikard-sics:
> About that is sender ID / receiver ID available in californium endpoint context ?

Yes it is available but only from Californium 2.0.0-M17. Since the context DB is a singleton in the Californium we are using now the values could also be gotten from there (for now).

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/ServersInfoExtractor.java
rikard-sics: Yes, that is much nicer. Fixed in commit a255bd64d5334790597db3d9de6addcf81b78433

leshan-core/src/main/java/org/eclipse/leshan/SecurityMode.java
rikard-sics: I now removed the SecurityMode for OSCORE. Instead the OSCORE Security Mode link in the Security object will be used. Done in commit 3a4e777eb664aa126c9c4bae1eb1effba7eace0e

rikard-sics:
> You should have a look at [§5.5.5 - Endpoint Client Name](http://www.openmobilealliance.org/release/LightweightM2M/V1_1-20180710-A/HTML-Version/OMA-TS-LightweightM2M_Transport-V1_1-20180710-A.html#5-5-5-0-555-Endpoint-Client-Name) about you "identity" concern.
> 

I now updated the code so that the Sender ID of the client is equal to its name (endpoint). So if I start a client with name "c1" its sender ID will be 0x6331, and on the server side in the web interface the sender ID must be entered as "6331" (hexadecimal). Done in commit 401db2bfd59b9481cc5f452a524b7c52bdb84f19

rikard-sics:
For further work on the identity I would perhaps wait until after updating to Californium 2.0.0-M17. There the sender ID, recipient ID and ID context are conveniently available in the endpoint context.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java
```diff
@@ -97,36 +97,9 @@ public synchronized Server createEndpoint(ServerInfo serverInfo) {
...
-                        }
-                        return true;
-                    }
-                });
-                serverIdentity = Identity.rpk(serverInfo.getAddress(), expectedKey);
-            } else if (serverInfo.secureMode == SecurityMode.OSCORE) {
-                System.out.println("Adding OSCORE CTX " + serverInfo.getFullUri().toASCIIString());
+            // oscore only mode
+            if(serverInfo.useOscore) {
+            	System.out.println("Adding OSCORE CTX " + serverInfo.getFullUri().toASCIIString());
```
sbernard31: Please use LOG instead of sysout ? :)

rikard-sics: Yes, that was an old debug statement I should have fixed. Now done in commit a83918e1619c160747895ccc6e44731d1d90a6ed

rikard-sics:
I had to do one more commit regarding the registration, Identity and SecurityChecker. Previously I could not register an OSCORE client using the name I configured it with in the web interface of the server. Since it counted as not secure it but yet had a non-null securityInfo it wanted me to use DTLS. I fixed this in commit 40213f6bfc1fae4dd4c0d51eb3d48d17a1cfbd5a

The SecurityChecker should be expanded with checks to the identity when using OSCORE. Also the Identity class expanded to support OSCORE-related identities. SecurityInfo already has a useOSCore flag.

leshan-client-core/src/main/java/org/eclipse/leshan/client/servers/ServerInfo.java
```diff
@@ -65,7 +66,7 @@ public URI getFullUri() {
     }
 
     public boolean isSecure() {
-        return secureMode != SecurityMode.NO_SEC;
+        return secureMode != SecurityMode.NO_SEC || useOscore == true;
```
sbernard31: I know this is a bit strange but here I will not really change this.

Historically this was used to know if we should use unsecured(coap) endpoint or secured(coaps) endpoint.

So maybe the right name is "isCoaps" ?
But anyway I read the code again and maybe this does not make so much sense now, so it would be more wise to not touch it for now.

rikard-sics: Yes, fair enough, I do not need to change the functionality of isSecure. I see your point that it is more about coap vs. coaps (and OSCORE could apply to both cases). I changed this in commit 17098a79932f4c48aa608b7418edc3119a193be5

sbernard31:
About `CaliforniumEndpointsManager`, maybe I missed something but the idea is to begin by supporting Oscore without DTLS right ?

So I think the oscore stuff is not at "The Good Place".
this is the bad place!

> For further work on the identity I would perhaps wait until after updating to Californium 2.0.0-M17. There the sender ID, recipient ID and ID context are conveniently available in the endpoint context.

makes sense !

rikard-sics:
> About `CaliforniumEndpointsManager`, maybe I missed something but the idea is to begin by supporting Oscore without DTLS right ?

Yes, my intention was to start by supporting OSCORE without DTLS. So the good place would be outside of the `if(serverInfo.isSecure())`, like down by row 220?

sbernard31:
> So the good place would be outside of the if(serverInfo.isSecure()), like down by row 220?

yep :)

rikard-sics:
> yep :)

Okay, sounds good :)

I have now done this in commit 7116077e8e6ed4a1a8456b9040333ad01875b146

Perhaps at some later stage it could be extracted to a separate method since we may want to combine the OSCORE configuration of an endpoint with different DTLS modes in a convenient way and minimize code reuse.

sbernard31:
> Perhaps at some later stage it could be extracted to a separate method since we may want to combine the OSCORE configuration of an endpoint with different DTLS modes in a convenient way and minimize code reuse.

Correct, we will see when we will add this.


I think we are good, do you plan to add more stuff in this PR ? (maybe just fix formatting issue?)

rikard-sics:
> I think we are good, do you plan to add more stuff in this PR ? (maybe just fix formatting issue?)

Yes, fixing the formatting was the last thing I had planned. I will proceed with that.

rikard-sics:
> I think we are good, do you plan to add more stuff in this PR ? (maybe just fix formatting issue?)

I have now applied the Leshan Eclipse formatter to the parts of the code I have changed. Done in commit 0138768e1b906ad8f81f67389103266832360d66

sbernard31:
I squashed all your commits in one (e31012710a149b619852f6b741e8ea4da083f3e6) which is now integrated in oscore branch.

I tried a "mvn clean install" and that failed so I fixed issues, see commit cc6136a63f081cfa8f8280a6ef67ff53a84288ee
Do not hesitate to launch this often, tests is here to help you ;)

About formatting I find several more issues, see commit e2ae4bca7cffca3430876a732a793defebf1b22d.

Are you using eclipse ? Did you follow [this guide](https://github.com/eclipse/leshan/wiki/Code-&-design-guidelines#configure-your-ide). In theory "format on save" is activated per project, so it should be very difficult to write bad formatted code :thinking:.

What do you plan as next steps ? :)

rikard-sics:
> I squashed all your commits in one ([e310127](https://github.com/eclipse/leshan/commit/e31012710a149b619852f6b741e8ea4da083f3e6)) which is now integrated in oscore branch.

Thanks!

> I tried a "mvn clean install" and that failed so I fixed issues, see commit [cc6136a](https://github.com/eclipse/leshan/commit/cc6136a63f081cfa8f8280a6ef67ff53a84288ee)
> Do not hesitate to launch this often, tests is here to help you ;)

Yes, I was using Eclipse and not running the maven tests from the command line like that. I should have ran tests more frequently and will try to do so in the future.

> About formatting I find several more issues, see commit [e2ae4bc](https://github.com/eclipse/leshan/commit/e2ae4bca7cffca3430876a732a793defebf1b22d).
> 
> Are you using eclipse ? Did you follow [this guide](https://github.com/eclipse/leshan/wiki/Code-&-design-guidelines#configure-your-ide). In theory "format on save" is activated per project, so it should be very difficult to write bad formatted code .

Yes, I am using Eclipse. Unfortunately up until now I have not been using the "format on save" feature. But rather formatting manually with CTRL+SHIFT+F. I will start using "format on save" as you suggest.

> What do you plan as next steps ? :)

Some thoughts for next steps:

- Update code to Californium 2.0.0-M17
- Implement an Identity and related things in SecurityChecker for OSCORE
- Work on the server side issues according to discussion in https://github.com/eclipse/leshan/issues/727
- Add possibility of entering OSCORE context information as command line parameters for client

What do you think makes most sense to prioritize?

sbernard31:
This order makes sense to me :)

rikard-sics:
> This order makes sense to me :)

Okay, sounds good. I will proceed accordingly. :)

PR #749, opened 2019-10-03
Update to use Californium 2.0.0-M17

rikard-sics:
The OSCORE-related code has been updated to use Californium 2.0.0-M17. For now the OSCORE context database is still handled as a singleton. This will be updated in a coming PR.


rikard-sics:
Now the code has been updated to use Californium 2.0.0-M17. This will give access to more recent OSCORE features like OSCORE-related information (like sender/recipient ID) about incoming/outgoing CoAP messages.

In the recent versions of Californium the OSCORE context database is no longer a singleton. As we discussed before it is better if it is not a singleton in Leshan. For now I kept the OSCORE context database as as singleton and plan a separate PR to change this. I need to consider how I can make the context database accessible to all the places where it is needed.

sbernard31:
It sounds good. Integrated in `oscore` branch.

Issue #569, opened 2018-09-20
Use a more modern command line framework for demo.


sbernard31:
Currently we are using apache commons CLI for command line.
It is a bit old school. We could consider using something more modern like [picocli](https://github.com/remkop/picocli) with colors and autocompletion support.

Commons CLI 1.3.1 is ~53KB and Picocli 3.6.0 is ~237KB.
I don't know if there is lighter modern framework ?

sbernard31:
Eclipse console [does not support ANSI color code for now](https://bugs.eclipse.org/bugs/show_bug.cgi?id=112948) ...
Picocli have [some heuristics](https://github.com/remkop/picocli/issues/512#issuecomment-429170587) to check if color is supported : 
We should check if it behave correctly before to consider seriously picocli.

sbernard31:
Here is an example with CLI + interactive commands : https://github.com/sbernard31/dtls-server/tree/master/src/main/java/org/eclipse/californium/scandium/examples

sbernard31:
First try for **leshan-client-demo** show we have a cleaner code to manage CLI (than apache common-cli)

We also get a prettier usage help : 

![help_usage](https://user-images.githubusercontent.com/840294/127345965-4635237a-927c-44e3-b848-85d648328afa.png)

Error Handling is better too : 
![error](https://user-images.githubusercontent.com/840294/127346251-ffbd2e30-6a04-4e9e-b913-853e8aeba9d0.png)

sbernard31:
Another step, using Picocli + Jline2 to implement interactive console : #1064

sbernard31:
(Maybe one day we could use Jline3 : #1065)

sbernard31:
This is now integrated in master (#1059, #1071, #1077)

And #1079 is coming.

PR #755, opened 2019-10-10
Allow setting OSCORE security settings in client command line parameters

rikard-sics:
This PR adds support to the Leshan client for setting OSCORE security parameters like master secret and sender/recipient ID from the command line.


sbernard31:
Currently, we hack a little bit the `org.apache.commons.cli` to group option in help.
(Instead of doing this, we should probably use a more modern cli framework ... see #569 but not in the priority list for now)

The final result looks like this : 
```
usage: java -jar leshan-client-demo.jar [OPTION]

 -h,--help      Display help information.
 -n <arg>       Set the endpoint name of the Client.
                Default: the local hostname or 'LeshanClientDemo' if any.
 -b             If present use bootstrap.
 -lh <arg>      Set the local CoAP address of the Client.
                Default: any local address.
 -lp <arg>      Set the local CoAP port of the Client.
                Default: A valid port value is between 0 and 65535.
 -u <arg>       Set the LWM2M or Bootstrap server URL.
                Default: localhost:5683.
 -pos <arg>     Set the initial location (latitude, longitude) of the device to be
                reported by the Location object.
                Format: lat_float:long_float
 -sf <arg>      Scale factor to apply when shifting position.
                Default is 1.0.
                .
                .
                ================================[ PSK ]=================================
                | By default Leshan demo use non secure connection.                    |
                | To use PSK, -i and -p options should be used together.               |
 -i <arg>       Set the LWM2M or Bootstrap server PSK identity in ascii.
 -p <arg>       Set the LWM2M or Bootstrap server Pre-Shared-Key in hexa.
                .
                .
                ================================[ RPK ]=================================
                | By default Leshan demo use non secure connection.                    |
                | To use RPK, -cpubk -cprik -spubk options should be used together.    |
 -cpubk <arg>   The path to your client public key file.
                The public Key should be in SubjectPublicKeyInfo format (DER encoding).
 -cprik <arg>   The path to your client private key file.
                The private key should be in PKCS#8 format (DER encoding).
 -spubk <arg>   The path to your server public key file.
                The public Key should be in SubjectPublicKeyInfo format (DER encoding).
                .
                .
                ================================[X509]==================================
                | By default Leshan demo use non secure connection.                    |
                | To use X509, -ccert -cprik -scert options should be used together.   |
 -ccert <arg>   The path to your client certificate file.
                The certificate Common Name (CN) should generaly be equal to the client
                endpoint name (see -n option).
                The certificate should be in X509v3 format (DER encoding).
 -scert <arg>   The path to your server certificate file.
                The certificate should be in X509v3 format (DER encoding).
 -oscore        Use OSCORE for communication between client and server.
```

With this PR it looks like :
```
usage: java -jar leshan-client-demo.jar [OPTION]

 -h,--help             Display help information.
 -n <arg>              Set the endpoint name of the Client.
                       Default: the local hostname or 'LeshanClientDemo' if any.
 -b                    If present use bootstrap.
 -lh <arg>             Set the local CoAP address of the Client.
                       Default: any local address.
 -lp <arg>             Set the local CoAP port of the Client.
                       Default: A valid port value is between 0 and 65535.
 -u <arg>              Set the LWM2M or Bootstrap server URL.
                       Default: localhost:5683.
 -pos <arg>            Set the initial location (latitude, longitude) of the device to be
                       reported by the Location object.
                       Format: lat_float:long_float
 -sf <arg>             Scale factor to apply when shifting position.
                       Default is 1.0.
                       .
                       .
                       ================================[ PSK
                       ]=================================
                       | By default Leshan demo use non secure connection.
                       |
                       | To use PSK, -i and -p options should be used together.
                       |
                       -------------------------------------------------------------------
                       -----
 -i <arg>              Set the LWM2M or Bootstrap server PSK identity in ascii.
 -p <arg>              Set the LWM2M or Bootstrap server Pre-Shared-Key in hexa.
                       .
                       .
                       ================================[ RPK
                       ]=================================
                       | By default Leshan demo use non secure connection.
                       |
                       | To use RPK, -cpubk -cprik -spubk options should be used together.
                       |
                       | To get helps about files format and how to generate it, see :
                       |
                       | See
                       https://github.com/eclipse/leshan/wiki/Credential-files-format   |
                       -------------------------------------------------------------------
                       -----
 -cpubk <arg>          The path to your client public key file.
                       The public Key should be in SubjectPublicKeyInfo format (DER
                       encoding).
 -cprik <arg>          The path to your client private key file.
                       The private key should be in PKCS#8 format (DER encoding).
 -spubk <arg>          The path to your server public key file.
                       The public Key should be in SubjectPublicKeyInfo format (DER
                       encoding).
                       .
                       .
                       ================================[X509]=============================
                       =====
                       | By default Leshan demo use non secure connection.
                       |
                       | To use X509, -ccert -cprik -scert options should be used
                       together.   |
                       | To get helps about files format and how to generate it, see :
                       |
                       | See
                       https://github.com/eclipse/leshan/wiki/Credential-files-format   |
                       -------------------------------------------------------------------
                       -----
 -ccert <arg>          The path to your client certificate file.
                       The certificate Common Name (CN) should generaly be equal to the
                       client endpoint name (see -n option).
                       The certificate should be in X509v3 format (DER encoding).
 -scert <arg>          The path to your server certificate file.
                       The certificate should be in X509v3 format (DER encoding).
 -oscore               Use OSCORE for communication between client and server.
 -mastersecret <arg>   The OSCORE pre-shared key used between the Client and LwM2M
                       Server/Bootstrap-Server.
 -mastersalt <arg>     The OSCORE master salt used between the Client and LwM2M
                       Server/Bootstrap-Server.
 -idcontext <arg>      The OSCORE ID Context used between the Client and LwM2M
                       Server/Bootstrap-Server.
 -senderid <arg>       The OSCORE Sender ID used by the client to the LwM2M
                       Server/Bootstrap-Server.
 -recipientid <arg>    The OSCORE Recipient ID used by the client to the LwM2M
                       Server/Bootstrap-Server.
 -aead <arg>           The OSCORE AEAD algorithm used between the Client and LwM2M
                       Server/Bootstrap-Server.
 -hkdf <arg>           The OSCORE HKDF algorithm used between the Client and LwM2M
                       Server/Bootstrap-Server.
```

1. I would group the OSCORE option like we do for PSK/RPK/X509.
2. I should use smaller option name to avoid the presentation issue.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -135,6 +130,20 @@ public static void main(final String[] args) {
         options.addOption("scert", true,
                 "The path to your server certificate file.\n The certificate should be in X509v3 format (DER encoding).");
         options.addOption("oscore", false, "Use OSCORE for communication between client and server.");
```
sbernard31: Do we still need this global option `oscore` ? or maybe just the presence of new oscore option is enough ?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -312,20 +321,100 @@ public static void main(final String[] args) {
...
+
+            if (senderidStr == null) {
+                System.err.println("The OSCORE Sender ID must be indicated");
+                formatter.printHelp(USAGE, options);
+                return;
+            }
+
+            // Check that the Sender ID matches the client name
+            // http://www.openmobilealliance.org/release/LightweightM2M/V1_1-20180710-A/HTML-Version/OMA-TS-LightweightM2M_Transport-V1_1-20180710-A.html#5-5-5-0-555-Endpoint-Client-Name
+            if (!Arrays.equals(endpoint.getBytes(), Hex.decodeHex(senderidStr.toCharArray()))) {
```
sbernard31: If senderID must match the endpoint name, we can remove the senderID option and just document in help that endpoint client name will be used as senderID ?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -312,20 +321,100 @@ public static void main(final String[] args) {
...
+                return;
+            }
+
+            // Parse AEAD Algorithm (set default if not indicated)
+            if (aeadStr == null) {
+                aeadStr = "AES_CCM_16_64_128";
+            }
+
+            int aeadInt;
+            if (aeadStr.equals("AES_CCM_16_64_128") || aeadStr.equals("10")) {
```
sbernard31: if we support only 1 AEAD algorithmm we can remove this option ? and just document that only this algorithm is used.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -312,20 +321,100 @@ public static void main(final String[] args) {
...
+            }
+
+            // Parse HKDF Algorithm (set default if not indicated)
+            if (hkdfStr == null) {
+                hkdfStr = "HKDF_HMAC_SHA_256";
+            }
+
+            int hkdfInt;
+            if (hkdfStr.equals("HKDF_HMAC_SHA_256") || hkdfStr.equals("-10")) {
+                hkdfInt = -10;
```
sbernard31: same here ?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -312,20 +321,100 @@ public static void main(final String[] args) {
...
-        boolean useOSCore = false;
+        OSCoreSettings oscoreSettings = null;
         if (cl.hasOption("oscore")) {
             System.out.println("Using OSCORE");
-            useOSCore = true;
 
             HashMapCtxDB db = OscoreHandler.getContextDB();
             OSCoreCoapStackFactory.useAsDefault(db);
 
+            // Parse OSCORE related command line parameters
```
sbernard31: We should begin to check if configuration is consistent and all oscore mandatory option are present. (like we do for PSK/RPK/X509)

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -312,20 +321,100 @@ public static void main(final String[] args) {
...
+
+            String mastersecretStr = cl.getOptionValue("mastersecret");
+            String mastersaltStr = cl.getOptionValue("mastersalt");
+            String idcontextStr = cl.getOptionValue("idcontext");
+            String senderidStr = cl.getOptionValue("senderid");
+            String recipientidStr = cl.getOptionValue("recipientid");
+            String aeadStr = cl.getOptionValue("aead");
+            String hkdfStr = cl.getOptionValue("hkdf");
```
sbernard31: could you move the `String param = cl.getOptionValue("param") near their check ?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -466,4 +557,26 @@ public void run() {
             }
         }
     }
+
+    // Class for holding OSCORE related information
+    private static class OSCoreSettings {
+        public String masterSecret;
+        public String masterSalt;
+        public String idContext;
```
sbernard31: it sounds not to be used ?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -312,20 +321,100 @@ public static void main(final String[] args) {
                 return;
             }
         }
+
         // Set boolean controlling OSCORE usage
-        boolean useOSCore = false;
+        OSCoreSettings oscoreSettings = null;
         if (cl.hasOption("oscore")) {
             System.out.println("Using OSCORE");
```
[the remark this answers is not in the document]
rikard-sics: I have now removed that line and added some logging in commit 5d518fa945c946a784260e5a68bb693ccab7baf8

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
rikard-sics: I have now updated the parsing of the AEAD algorithm to parse it into a COSE AlgorithmID and not compare it to fixed values. My goal is to support multiple algorithms. Done in commit 76fa1268a8f8262eb39008693f37a75069173ba2

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
rikard-sics: Did the same thing as for the AEAD algorithm, the HKDF algorithm is now parsed into a COSE AlgorithmID and not compared it to fixed values. Done in commit 76fa1268a8f8262eb39008693f37a75069173ba2

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
rikard-sics: Yes, it is not used. It seems the OSCORE object as defined in LWM2M 1.1 does not support the ID Context parameter currently. This is something that the main OSCORE implementation uses and would be nice to have support for in LWM2M and Leshan long term. I have discussed this with Ericsson and we will see where things lead.

I removed this variable in commit 09a48e3d46c1b9234b84bf10df66f43c35a89fc1

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
rikard-sics: After discussions with Ericsson it may be better to remove this check. It is better if the Sender ID can be kept short since it will be included in every outgoing request. So having a long client endpoint name can then induce a lot of communication overhead. Thus it can be better to not require that they match.

I removed this check in commit 3892cb62eaadfb62acf3e7bcf1469952b05dae4b

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
rikard-sics: Done in commit 5d2aa3943fa4944954b1a2399f39a65a54da7e1e

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
rikard-sics: True, we do not need that. Instead we can check if the OSCORE master secret is set. I removed it in commit cf7aa439ab91a3b367945217ef38d849c402adc6

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
rikard-sics: Done in commit 0caaf4569fe0d6a380310ad2f9a4d87eb642f86d

rikard-sics:
> * I would group the OSCORE option like we do for PSK/RPK/X509.
> * I should use smaller option name to avoid the presentation issue.

This has now been done in commit 2074d38d0f60b697e047a9eb5387bc9068400ee3

sbernard31:
It sounds good. I will integrate it.

> After discussions with Ericsson it may be better to remove this check. It is better if the Sender ID can be kept short since it will be included in every outgoing request. So having a long client endpoint name can then induce a lot of communication overhead. Thus it can be better to not require that they match.

The specification clearly say this is possible but not mandatory, so its OK for me to remove this.
In another hand, this is just demo and so the constraint you add here is just for demo and not for core, so this should not impact Ericsson.

Another possibility would to use endpointname as default value when `sid` parameter is not used.

rikard-sics:
> The specification clearly say this is possible but not mandatory, so its OK for me to remove this.
> In another hand, this is just demo and so the constraint you add here is just for demo and not for core, so this should not impact Ericsson.

True, the specification says you may use as a hard check that they must match. I think for the demo and then for the server code it could be not mandatory to match, or we could implement that as a configurable option somewhere.

> Another possibility would to use endpointname as default value when `sid` parameter is not used.

Yes, that would be an option. Is it fine to leave this PR as is and in such case come back to that?

sbernard31:
> Is it fine to leave this PR as is and in such case come back to that?

Yes of course. Both are fine to me, what do you prefer ? (just let me know)

rikard-sics:
> Yes of course. Both are fine to me, what do you prefer ? (just let me know)

Yes I would prefer leaving this PR as is. Then I could come back to this, possibly after further discussions with project partners.

sbernard31:
This is now integrated in oscore branch (all commits squashed in 4f7e0c71942183174e87f2de1c9c56a2d9029cea)

rikard-sics:
> This is now integrated in oscore branch (all commits squashed in [4f7e0c7](https://github.com/eclipse/leshan/commit/4f7e0c71942183174e87f2de1c9c56a2d9029cea))

Thanks!

PR #761, opened 2019-10-24
add some executor housekeeping at server side

sbernard31:


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/request/RequestSender.java
```diff
@@ -390,4 +401,14 @@ public void onCancel() {
...
         }
     }
+
+    @Override
+    public void destroy() {
+        executor.shutdownNow();
+        try {
+            executor.awaitTermination(5, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            LOG.warn("Destroying InMemoryRegistrationStore was interrupted.", e);
```
msangoi: Bad copy/paste: `InMemoryRegistrationStore`

sbernard31: (fixed)

PR #766, opened 2019-10-31
Implements an identity and matching of the identity for OSCORE

rikard-sics:
An identity string based on OSCORE parameters is now set for connected clients that use OSCORE. The identity string contains the OSCORE Sender and Recipient ID the client and server are using. These parameters can be retrieved from the source EndpointContext.

The identity for OSCORE is also set as a SecurityInfo based on the OSCORE context configured on the server to use with a specific client.

The identity of a connected client and the configured SecurityInfo are then matched in the SecurityChecker.

### DIFF
--- a/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
+++ b/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
@@ -38,12 +38,6 @@ public class EndpointContextUtil {
 
     public static Identity extractIdentity(EndpointContext context) {
         InetSocketAddress peerAddress = context.getPeerAddress();
-        // Build identity for OSCORE if it is used
-        if (context.get(OSCoreEndpointContextInfo.OSCORE_SENDER_ID) != null) {
-            String oscoreIdentity = "sid=" + context.get(OSCoreEndpointContextInfo.OSCORE_SENDER_ID) + ",rid="
-                    + context.get(OSCoreEndpointContextInfo.OSCORE_RECIPIENT_ID);
-            return Identity.oscoreOnly(peerAddress, oscoreIdentity.toLowerCase());
-        }
         Principal senderIdentity = context.getPeerIdentity();
         if (senderIdentity != null) {
             if (senderIdentity instanceof PreSharedKeyIdentity) {
@@ -57,6 +51,13 @@ public class EndpointContextUtil {
                 return Identity.x509(peerAddress, x509CommonName);
             }
             throw new IllegalStateException("Unable to extract sender identity : unexpected type of Principal");
+        } else {
+            // Build identity for OSCORE if it is used
+            if (context.get(OSCoreEndpointContextInfo.OSCORE_SENDER_ID) != null) {
+                String oscoreIdentity = "sid=" + context.get(OSCoreEndpointContextInfo.OSCORE_SENDER_ID) + ",rid="
+                        + context.get(OSCoreEndpointContextInfo.OSCORE_RECIPIENT_ID);
+                return Identity.oscoreOnly(peerAddress, oscoreIdentity.toLowerCase());
+            }
         }
         return Identity.unsecure(peerAddress);
     }
--- a/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
+++ b/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
@@ -14,6 +14,7 @@
  *     Sierra Wireless - initial API and implementation
  *     Achim Kraus (Bosch Software Innovations GmbH) - add support for californium
  *                                                     endpoint context
+ *     Rikard Höglund (RISE SICS) - Additions to support OSCORE
  *******************************************************************************/
 package org.eclipse.leshan.core.californium;
 
@@ -30,12 +31,19 @@ import org.eclipse.californium.elements.EndpointContext;
 import org.eclipse.californium.elements.auth.PreSharedKeyIdentity;
 import org.eclipse.californium.elements.auth.RawPublicKeyIdentity;
 import org.eclipse.californium.elements.auth.X509CertPath;
+import org.eclipse.californium.oscore.OSCoreEndpointContextInfo;
 import org.eclipse.leshan.core.request.Identity;
 
 public class EndpointContextUtil {
 
     public static Identity extractIdentity(EndpointContext context) {
         InetSocketAddress peerAddress = context.getPeerAddress();
+        // Build identity for OSCORE if it is used
+        if (context.get(OSCoreEndpointContextInfo.OSCORE_SENDER_ID) != null) {
+            String oscoreIdentity = "sid=" + context.get(OSCoreEndpointContextInfo.OSCORE_SENDER_ID) + ",rid="
+                    + context.get(OSCoreEndpointContextInfo.OSCORE_RECIPIENT_ID);
+            return Identity.oscoreOnly(peerAddress, oscoreIdentity.toLowerCase());
+        }
         Principal senderIdentity = context.getPeerIdentity();
         if (senderIdentity != null) {
             if (senderIdentity instanceof PreSharedKeyIdentity) {
@@ -54,7 +62,8 @@ public class EndpointContextUtil {
     }
 
     /**
-     * Create californium endpoint context from leshan identity.
+     * Create californium endpoint context from leshan identity. OSCORE does not use a Principal but automatically sets
+     * properties in the endpoint context at message transmission/reception.
      * 
      * @param identity leshan identity received on last registration.
      * @return californium endpoint context for leshan identity
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityChecker.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityChecker.java
@@ -88,16 +88,14 @@ public class SecurityChecker {
 
                 return checkX509Identity(endpoint, clientIdentity, securityInfo);
 
-            } else if (clientIdentity.isOSCORE()) {
-
-                return checkOscoreIdentity(endpoint, clientIdentity, securityInfo);
-
             } else {
                 LOG.debug("Unable to authenticate client '{}': unknown authentication mode", endpoint);
                 return false;
             }
         } else {
-            if (securityInfo != null) {
+            if (clientIdentity.isOSCORE()) {
+                return checkOscoreIdentity(endpoint, clientIdentity, securityInfo);
+            } else if (securityInfo != null) {
                 LOG.debug("Client '{}' must connect using DTLS", endpoint);
                 return false;
             }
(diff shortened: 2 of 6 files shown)


rikard-sics:
Currently the identity for OSCORE is a string. It could be a more complex data structure containing the Recipient and Sender IDs as separate variables.

Right now the identity for OSCORE is generated in SecurityInfo. Another way would be to generate the identity for OSCORE in SecurityDeserializer but since I provide the OSCORE context to SecurityInfo either way, I thought I would generate it in SecurityInfo.

Another coming step is making the OSCORE context in SecurityInfo serializeable (at least regarding some parts of it).

leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityChecker.java
```diff
@@ -88,12 +88,16 @@ public boolean checkSecurityInfo(String endpoint, Identity clientIdentity, Secur
 
                 return checkX509Identity(endpoint, clientIdentity, securityInfo);
 
+            } else if (clientIdentity.isOSCORE()) {
+
+                return checkOscoreIdentity(endpoint, clientIdentity, securityInfo);
+
```
sbernard31: so this check must be in else block (clientIdentity.isSecure() == false)

leshan-server-demo/src/main/resources/webapp/partials/client-list.html
```diff
@@ -21,7 +21,7 @@ <h3><small><p class="text-right">Connected clients: {{clients.length}}</p></smal
 				<td>{{client.registrationDate | date:'medium'}}</td>
 				<td>{{client.lastUpdate | date:'medium'}}</td>
 				<td><i class="glyphicon glyphicon-info-sign" tooltip-html-unsafe="{{clientTooltip(client)}}"></i></td>
-				<td><span ng-style="{'visibility': client.secure ? 'visible': 'hidden'}" class="glyphicon glyphicon-lock" tooltip-html-unsafe="Communication over DTLS"></span></td>
+				<td><span ng-style="{'visibility': client.secure ? 'visible': 'hidden'}" class="glyphicon glyphicon-lock" tooltip-html-unsafe="Communication over DTLS/OSCORE"></span></td>
```
sbernard31: I would keep secure = DTLS for now.
If we want to improve this we should replace the boolean by something we could show the 4 possible states (coap, coaps, coap+oscore, coaps+oscore).

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
```diff
@@ -30,12 +31,19 @@
...
 public class EndpointContextUtil {
 
     public static Identity extractIdentity(EndpointContext context) {
         InetSocketAddress peerAddress = context.getPeerAddress();
+        // Build identity for OSCORE if it is used
+        if (context.get(OSCoreEndpointContextInfo.OSCORE_SENDER_ID) != null) {
+            String oscoreIdentity = "sid=" + context.get(OSCoreEndpointContextInfo.OSCORE_SENDER_ID) + ",rid="
+                    + context.get(OSCoreEndpointContextInfo.OSCORE_RECIPIENT_ID);
+            return Identity.oscoreOnly(peerAddress, oscoreIdentity.toLowerCase());
+        }
```
sbernard31: This should be done in an `else` block of the code above. (senderIdentity == null)

leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityInfo.java
```diff
@@ -53,6 +54,7 @@
 
     // OSCORE (FIXME: Save content properly information here. Must be serializable.)
     boolean useOSCore;
```
sbernard31: this attributes seems to be used anymore ?

rikard-sics:
I have now updated the code according to your comments. Done in commit 5c166fb3ae747d7587206330083ae084554e5956

Let me get back regarding your general comments.

rikard-sics:
> For now I don't get why we give the whole context in the UI, if finally we use only RID an SID ? thinking

All parameters given in the UI for the context are used in SecurityDeserializer and SecurityInfo to actually generate the OSCORE context and add it to the context database.

This is also why I give the whole context to SecurityInfo. My understanding is that for things to be retained on the server when it is closed it has to be in the SecurityInfo and serialized. That's why I have the TODO "// OSCORE (FIXME: Save content properly information here. Must be serializable.)" 

> As OSCORE is still not clear to me, I still do the parallel with PSK to help me to understand.
> 
> The psk identity is a kind of "login" which is unique and could identify the device. In OSCORE the equivalent is "RID+SID" ?

When a client connects to a server using OSCORE it includes its own Sender ID in the message (OSCORE CoAP option). The server then retrieves the appropriate context to use for decrypting the message based on the Sender ID. (Looking for a context with Recipient ID matching the Sender ID of the client.) An ID Context can also be used to further distinguish between contexts. So the pair Sender ID and ID Context should allow the server to uniquely identify the context to use for this message. (The context associated to the client.) Note that the Sender ID of the client is the Recipient ID of the server and vice-versa.

So strictly speaking I would prefer to use Sender ID+Recipient ID+ID Context as identity. That should uniquely identify a client (we can enforce restrictions on adding several clients with same Sender ID and ID Context.) However currently the ID Context is not part of the OSCORE object as defined in LWM2M (although that may change). So for now I stuck with the pair Sender ID+Recipient ID as identity.

> For PSK, the shared secret is store In SecurityStore(securityInfo) which is used by LwM2mPskStore which is the interface used by californium to retrieve PSK key.
> How does/will it work with OSCORE ? (I'm a bit lost)

I was thinking that rather than using the current design for storing OSCORE contexts we could have a similar solution as for DTLS with a LwM2MOScoreContextStore where OSCORE contexts can be retrived from. This could in turn use the HashMapCtxDB internally.

sbernard31:
I integrate this is `oscore` branche (commit 59acc4e6da48223db8156b4454cb24637f3e2ed0)

Thx for your explanation, it's clearer now.

>  SecurityInfo to actually generate the OSCORE context and add it to the context database.

Currently the fact that creating a SecurityInfo add it do the contextDB is an issue.
```java
public static SecurityInfo newOSCoreInfo(String endpoint, OSCoreCtx oscoreCtx) {
        Validate.notNull(oscoreCtx);

        // Add the OSCORE Context to the context database
        HashMapCtxDB db = OscoreHandler.getContextDB();
        db.addContext(oscoreCtx);
}
```
(let me explain this more in details below)

> I would prefer to use Sender ID+Recipient ID+ID Context as identity.

I see the contextID as a kind of session identifier. Am I wrong ?
If I'm not,  I'm not sure this make so much sense to use it as identifier at application layer. I mean at this level we don't care we change session as we have warranty that we talk to the right person.
(I mean contextID make sense at coap level but not sure at LWM2M one)

>  the current design for storing OSCORE contexts we could have a similar solution as for DTLS 

LwM2MOScoreContextStore which will take in parameter the `securityStore` like `LwM2mPskStore` ? and so no more add in HashMapCtxDB at securityStore creation ?

I ask myself if there is not several kind of data in OSCORE contexts.
 1. some kind of long lifetime data, which should not be modified often and which could be persisted and shared easily.
 2. some kind of ephemeral data, which can be lost without so much trouble and which should be store in memory.

I asked this because you must keep in mind that SecurityStore could have persistent implementation (using redis, or relation db) and so `LwM2MOScoreContextStore` should works with this kind of store.

rikard-sics:
> I integrate this is `oscore` branche (commit [59acc4e](https://github.com/eclipse/leshan/commit/59acc4e6da48223db8156b4454cb24637f3e2ed0))

Thanks!

> Thx for your explanation, it's clearer now.

No problem.

> I see the contextID as a kind of session identifier. Am I wrong ?
> If I'm not, I'm not sure this make so much sense to use it as identifier at application layer. I mean at this level we don't care we change session as we have warranty that we talk to the right person.
> (I mean contextID make sense at coap level but not sure at LWM2M one)

Basically the ID Context can be used to disambiguate between several contexts in the case when a node communicating using OSCORE has 2 contexts with the same Recipient ID. So if I get a message from Sender ID 0x20 and I have two contexts with Recipient ID 0x20 I can use the ID Context (also included in the message) to understand to which context it refers to. The ID Context also has some other uses such as in the context re-derivation procedure described in Appendix B.2.

So to identify a context using the triple Sender ID, Recipient ID and ID Context makes sense to me. (Of course the ID Context can be unset/null). But there is flexibility and we can for instance make a restriction on the server to not allow duplicate Recipient IDs at all, then having to use the ID Context to disambiguate will not be a problem. But then perhaps we lose flexibility in more advanced scenarios where clients are configured without a full picture of when IDs will overlap with other existing clients. Or scenarios involving bootstrapping.

See also:
https://tools.ietf.org/html/rfc8613#section-5.1
https://tools.ietf.org/html/rfc8613#appendix-B.2

> LwM2MOScoreContextStore which will take in parameter the `securityStore` like `LwM2mPskStore` ? and so no more add in HashMapCtxDB at securityStore creation ?

As things are currently I believe it would still need to be added to the map of contexts because the OSCORE code uses the map internally in Californium when sending and receiving. 

> 
> I ask myself if there is not several kind of data in OSCORE contexts.
> 
>     1. some kind of long lifetime data, which should not be modified often and which could be persisted and shared easily.
> 
>     2. some kind of ephemeral data, which can be lost without so much trouble and which should be store in memory.
> 
> 
> I asked this because you must keep in mind that SecurityStore could have persistent implementation (using redis, or relation db) and so `LwM2MOScoreContextStore` should works with this kind of store.

Yes you are very right that there is different types of information stored in a context. All the information read in the UI of the server as it is now is what I would consider "static" information that changes very rarely. That is what I would wish to store long-term so the server has it even after a reboot. (The whole OSCORE context object(s) that are in the map should not be stored, only some parts.)

Then for the ephemeral information like sequence numbers we can solve this by using the context re-derivation procedure detailed in Appendix B.2. to re-establish a connection if the server or client loses power/reboots/crashes etc. That will allow re-establishing a secure channel based on the same static information and still allow for re-starting sequence numbers and other things at 0 without issues like re-using the same key and nonce. (Using a kind of handshake that changes the ID Context in several steps.)

sbernard31:
> That will allow re-establishing a secure channel based on the same static information and still allow for re-starting sequence numbers and other things at 0 without issues like re-using the same key and nonce. (Using a kind of handshake that changes the ID Context in several steps.)

This reinforces my idea that Context ID is a CoAP layer concept and should maybe not be used as identifier at LWM2M level. (But clearly I maybe missed something)

> the ID Context can be used to disambiguate between several contexts in the case when a node communicating using OSCORE has 2 contexts with the same Recipient ID. 

I get it but I think same SID RID means same peers and different CID means a kind of different session between us.

>  But there are options and we can for instance make a restriction on the server to not allow duplicate Recipient IDs at all,

It sounds to me a good first step for a minimal viable version of this feature.
(But if you think this is a better idea to handle Context ID as part of peer identifier in the first version, do not hesitate to insist)

(I close this issue as I already integrate the PR in master but we can continue the discussion if needed :-) )

PR #769, opened 2019-11-07
Remove TODO about usage of "bad request" on update and delete registration if device is unauthorized.

sbernard31:
See https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/181

The specification says : 
> The LwM2M specification defines the use of the endpoint client name in
the Bootstrap-Request and in the Register messages. Since the endpoint
client name is not authenticated at the application layer the LwM2M
Server MUST compare the received endpoint client name identifier with
the identifier used at the DTLS handshake. This comparison may either be
an equality match or may involve a dedicated lookup table to ensure that
LwM2M Clients cannot intentionally or due to misconfiguration
impersonate other LwM2M Clients. The LwM2M Server MUST respond with a
“4.00 Bad Request” to the LwM2M Client if these fields do not match.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationHandler.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/RegistrationHandler.java
@@ -97,8 +97,6 @@ public class RegistrationHandler {
         }
 
         if (authorizer.isAuthorized(updateRequest, registration, sender) == null) {
-            // TODO replace by Forbidden if https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/181 is
-            // closed.
             return new SendableResponse<>(UpdateResponse.badRequest("forbidden"));
         }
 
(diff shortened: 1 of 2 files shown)


sbernard31:
About **Registration Update** and **De-Registration**, the specification and the OpenMobileAlliance/OMA_LwM2M_for_Developers#181 seem to confirm the usage of BAD_REQUEST.

For **Registration**, this is not so simple as the specification also defines : 
> **4.00 Bad Request**
>The mandatory parameter is not specified or unknown parameter is specified
>Unknown Endpoint Client Name
>Endpoint Client Name does not match with CN field of X.509 Certificates
>
> **4.03 Forbidden**
>The Endpoint Client Name registration in the LwM2MServer is not allowed.

So in our case the `Authorizer` covers both cases (Forbidden & Bad request)
To respect strictly the specification, we should rewrite the code to separate both cases.
But from a security perspective, this is generally not a good idea to give information like this.
So the best solution seems to always return forbidden as :
 - this is the generic "unauthorized" code
 - this will not change the current Leshan behavior.

PR #771, opened 2019-11-08
Server SHOULD not initiate DTLS connection for device using Queue mode

sbernard31:
We consider that initiates a connection (server act as DTLS client to initiate a handshake) does not make sense for QueueMode as if we lost connection device is probably absent.


leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java
```diff
@@ -85,6 +89,13 @@ public static EndpointContext extractContext(Identity identity) {
                 peerIdentity = new X500Principal("CN=" + identity.getX509CommonName());
             }
         }
+
+        if (peerIdentity != null) {
+            if (preventConnectionInitiation) {
```
msangoi: single line "ifs"

PR #779, opened 2019-11-29
Fix race condition issue about PresenceService and simplify the code.

sbernard31:


leshan-server-core/src/main/java/org/eclipse/leshan/server/queue/PresenceServiceImpl.java
```diff
@@ -63,36 +64,47 @@ public void removeListener(PresenceListener listener) {
...
+            boolean stateChanged;
+            final AtomicReference<ScheduledFuture<?>> timerFuture = new AtomicReference<>();
+            // set this device as awake
+            AtomicReference<ScheduledFuture<?>> previous = clientPrecenses.put(reg.getEndpoint(), timerFuture);
             if (previous != null) {
-                // We already have a status for this reg.
-                status = previous;
+                stateChanged = false;
+                // cancel previous timer
+                previous.get().cancel(false);
```
msangoi: potential NPE if the previous future is not set yet.

sbernard31:
@msangoi, Could you have a look again ? 

(especially the NPE/race condition issue)

PR #810, opened 2020-02-17
Update README.md

jaimejim:
Without the -U flag (that forces a check for missing releases and updated snapshots) there is an installation error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ git clone https://github.com/eclipse/leshan.git
 Compile it, by running in leshan root folder :
 
 ```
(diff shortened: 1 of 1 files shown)


sbernard31:
@jaimejim  Thx for the PR.

But generally `mvn clean install` is enough. Personnaly, I never needed to use -U to compile leshan. 
Could you explain a bit more which kind of issue you faced and under what circumstances ?

sbernard31:
Without more explanation, I prefer to no integrate this, but I'm open to discussion. :)

PR #812, opened 2020-02-18
Add UnconnectedPeerException and consider HanshakeTimeout as "regular" timeout

sbernard31:
About considering "consider HanshakeTimeout as regular timeout" (see https://github.com/eclipse/californium/issues/1223)


sbernard31:
@msangoi, I merged it in master but you can still have a look and giving feedback if you want :)

PR #834, opened 2020-04-21
#833 : Support that Executable resource model should not have a type.

### LINKED ISSUE #833: DDFFileParser does not handle resource without a data type ("none")
The specifications of both LWM2M 1.0 and 1.1 have normative listings of resources' data type. There's a "none" entry in that listing, which states

>No specific data type affected to that resource: it exclusively concerns Executable Resource.

When reading resource model definitions using DDFFileParser, and the resource is specified with `<Type></Type>`, that resource model still gets `Type.STRING` as type.

See https://github.com/eclipse/leshan/blob/dea23774923a2abfb09968128802ce3fcde0ab4b/leshan-core/src/main/java/org/eclipse/leshan/core/model/DDFFileParser.java#L134 for the implicit default case, and https://github.com/eclipse/leshan/blob/dea23774923a2abfb09968128802ce3fcde0ab4b/leshan-core/src/main/java/org/eclipse/leshan/core/model/DDFFileParser.java#L158-L180 for the switch case which we "fallthrough" with an empty type.

I would expect either `null` or an explicit sentinel value such as `Type.NONE`, where the latter is preferred both since it avoids having to null check, and because it's name matches the specification(s).

sbernard31:
This aims to fix #833 for the 1.0.x branch.

- I added more logs to warn about invalid model. (This should help people to fix their model)
- Do not use the `Type.String` by default for executable resource. (I keep the default value for none executable, maybe not a so good idea :thinking:)
- Raise a `CodecException` if we try to encode a "without type" (probably executable) resource.

The new code reveals an issue in one of our included model : 
```
Model for Resource Reset Min and Max Measured Values(5605) in /models/3303.xml is invalid : an executable resource MUST NOT have a type(STRING)
```

So I will probably update models regarding new version available at [OMA LWM2M Registry](https://github.com/OpenMobileAlliance/lwm2m-registry).


sbernard31:
I updated all models and added the missing one in all demos.
I removed `Communications Characteristics` (17) and `Non-Access Stratum (NAS) Configuration`(18) as they are not yet published (see https://github.com/OpenMobileAlliance/lwm2m-registry/issues/443)

sbernard31:
I add a fix on 10260 object model (see https://github.com/OpenMobileAlliance/lwm2m-registry/issues/446)

Issue #832, opened 2020-04-17
Could not acquire a lock from redis error - on creating observe request.


madhushreegc:
Hi ,

Leshan version M10.

I am triggering more than 20 observe request towards the same device from leshan server . Some time I see below error.  If I change the lock key used now for add, addObservation method in RedisStoreRegistraion like LOCK:EP:<deviceName>:lwm2mpath (example : LOCK:EP:urn:imei:abcd:3:0:3) . Will I face any issues any where on any operations? 

04:36:39,985 ERROR [StripedExchangeJob] Exception in striped thread: Could not acquire a lock from redis
java.lang.IllegalStateException: Could not acquire a lock from redis
        at org.eclipse.leshan.server.cluster.RedisLock.acquire(RedisLock.java:52) ~[leshan-server-cluster-1.0.0-M10.jar:?]
        at org.eclipse.leshan.server.cluster.RedisRegistrationStore.add(RedisRegistrationStore.java:551) ~[leshan-server-cluster-1.0.0-M10.jar:?]
        at org.eclipse.leshan.server.cluster.RedisRegistrationStore.putIfAbsent(RedisRegistrationStore.java:533) ~[leshan-server-cluster-1.0.0-M10.jar:?]
        at org.eclipse.californium.core.network.BaseMatcher.registerObserve(BaseMatcher.java:187) ~[californium-core-2.0.0-M12.jar:?]
        at org.eclipse.californium.core.network.UdpMatcher.sendRequest(UdpMatcher.java:115) ~[californium-core-2.0.0-M12.jar:?]
        at org.eclipse.californium.core.network.CoapEndpoint$OutboxImpl.sendRequest(CoapEndpoint.java:713) ~[californium-core-2.0.0-M12.jar:?]
        at org.eclipse.californium.core.network.stack.BaseCoapStack$StackBottomAdapter.sendRequest(BaseCoapStack.java:187) ~[californium-core-2.0.0-M12.jar:?]
        at org.eclipse.californium.core.network.stack.ReliabilityLayer.sendRequest(ReliabilityLayer.java:107) ~[californium-core-2.0.0-M12.jar:?]
        at org.eclipse.californium.core.network.stack.BlockwiseLayer.sendRequest(BlockwiseLayer.java:296) ~[californium-core-2.0.0-M12.jar:?]
[...]

PR #837, opened 2020-04-22
Fix race condition in RedisLock and add log to detect unintended expiration

sbernard31:
This fix was triggered by #832.

### DIFF
--- a/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisLock.java
+++ b/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisLock.java
@@ -18,7 +18,12 @@ package org.eclipse.leshan.server.redis;
 import java.util.Arrays;
 import java.util.Random;
 
+import org.eclipse.leshan.core.util.Hex;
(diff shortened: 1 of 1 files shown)


leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisLock.java
```diff
@@ -63,10 +69,36 @@
...
+                // the key must not be deleted.
+                LOG.warn(
+                        "Nothing to release for key {}, meaning the key probably expired because of acquiring the lock for too long (more than {}ms)",
+                        new String(lockKey), LOCK_EXP);
+                j.unwatch();
+            }
+
             if (Arrays.equals(j.get(lockKey), lockValue)) {
-                j.del(lockKey);
+                Long deleted = j.del(lockKey);
```
msangoi, line 96: Do we still need this block? 🤔

sbernard31: Uups, I will fix this.

Issue #852, opened 2020-06-12
Unable to "talk" to long sleeping devices behind NAT (patch)


sdrsdr:
For a NB-IoT devices we're facing NAT inflicted communications problems:
As lwm2m protocol us  UDP based and our devices usually sit in mobile operators private networks we can't always assume working communication path from server to device. To be able to talk to our device we always observer our device on registration and have command queues for every device holding commands to that device to be executed when a periodic "refresh" of the observation is triggered by the device and the NAT channel is granted (usually for just 30 seconds). When the NAT session expires the next UDP packet from the device is with different port and probably different host. With default Network config this breaks the registration matching but luckily this can be overcome with some `coapConfig.setString(NetworkConfig.Keys.RESPONSE_MATCHING,"PRINCIPAL");` and timeout tweaking. Yet  in the current code the registration identity is immutable and never changes  but this does not reflect the reality in our scenario and when the NAT session is recreated you should "talk" to the identity of the coap responce and not to the original registration identity. I've managed to device a minimal change to the current code base that allow proper communication in our scenario. The commit that does the trick is here:

https://github.com/sdrsdr/leshan/commit/6d569616e52c9d151eaddef4199fbf8922fa7d23

If there is a interest in this patch I can do a proper PR :)

sbernard31:
LWM2M behind a NAT is a bit tricky.

The way we make it works : 
1. Using LWM2M Queue Mode.
2. When LWM2M client wakes up 
2.1.At DTLS level : do a new handshake or abbreviated one (as IP/port changed so DTLS connection can not be retrieved)  or use [connection id](https://tools.ietf.org/html/rfc6347) (to identify connection by Id instead of IP/port).
2.2  At LWM2M level, client should start communication by doing an update (this will update the ip address of the registration). This seems to be mandatory. (see https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/391)
3. When LWM2M server receives an update, we send pending request sequentially.

So If you want to strictly respect the standard, I advice you to try to do something like this.
Could it be possible for you or do you see some blockers here ?

sdrsdr:
We don't use DTLS. NB-IoT have ridiculous  restrictions on number of packets send/received per hour too, so to keep user experience at somewhat satisfactory level we must make sure any packet we send is data packet and not "maintenance" one. On the other hand we use specific hardware with vendor provided LwM2M stack  that is far from feature complete ..

sdrsdr:
[...]

PR #854, opened 2020-06-15
add  Registration Update on Notification mode (out of specification)

sbernard31:
It was triggered by #852.


sdrsdr:
I can confirm this works great in our scenario!

sbernard31:
Thx @sdrsdr. This is now integrated in master.

PR #856, opened 2020-06-19
Add support of additional attributes for Bootstrap

sbernard31:
We already have a way to access to  additional attributes for REGISTER request.
(Additional attributes are attributes which are not defined in the specification. see https://github.com/eclipse/leshan/pull/66)

This PR aims to do the same thing but for BOOTSTRAP request.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager2.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager2.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Sierra Wireless and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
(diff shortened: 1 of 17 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapSessionManager2.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.eclipse.leshan.core.request.BootstrapRequest;
+import org.eclipse.leshan.core.request.Identity;
+
+/**
+ * A new version of {@link BootstrapSessionManager}.
+ * 
+ * @since 1.1
+ */
+public interface BootstrapSessionManager2 extends BootstrapSessionManager {
```
msangoi, line 26: Same question for this PR. Can't we update the existing interface and mark the old method as deprecated?

sbernard31: Unfortunately, [default method was added since java8](https://www.oracle.com/java/technologies/javase/8-whats-new.html). And for now we support java7. (that was mainly an android constraint) 

Maybe for the 2.0 we should consider to move to java8.

PR #858, opened 2020-06-26
Add Bootstrap Discover Request support.

sbernard31:
This PR add support for Bootstrap discover request.

This is fully implemented at client side.  
At server side this allow to customize `BootstrapHandler` behavior to send `BootstrapDiscoverRequest` if needed but the API should maybe be improved. I don't know really how for now ...  
This is how this looks like : 
```java
LeshanBootstrapServerBuilder builder = new LeshanBootstrapServerBuilder();
builder.setBootstrapHandlerFactory(new BootstrapHandlerFactory() {
    @Override
    public BootstrapHandler create(BootstrapConfigStore store, LwM2mBootstrapRequestSender sender,
            BootstrapSessionManager sessionManager) {
        return new DefaultBootstrapHandler(store, sender, sessionManager) {
            @Override
            protected void startBootstrap(final BootstrapSession session, final BootstrapConfig cfg) {
                send(session, request, new SafeResponseCallback<BootstrapDiscoverResponse>(session) {
                    @Override
                    public void safeOnResponse(BootstrapDiscoverResponse response) {
                        // Do what you want depending on response value.

                       // To continue bootstrap session
                       // (if you don't use anonymous class this is probably better to call super.startBootstrap())
                        delete(session, cfg, new ArrayList<>(cfg.toDelete));
                    }
                }, new SafeErrorCallback(session) {
                    @Override
                    public void safeOnError(Exception e) {
                        // do what you want on failure, here we just stop the session.
                        stopSession(session, BootstrapFailureCause.INTERNAL_SERVER_ERROR);
                    }
                });
            }
        };
    }
});
```

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/request/DownlinkRequestVisitor2.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/request/DownlinkRequestVisitor2.java
@@ -1,0 +1,26 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Sierra Wireless and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v20.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *     Sierra Wireless - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.leshan.core.request;
+
+/**
+ * A new version of {@link DownlinkRequestVisitor} which supports {@link BootstrapDiscoverRequest}
+ * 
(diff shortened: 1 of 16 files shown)


leshan-core/src/main/java/org/eclipse/leshan/core/request/DownlinkRequestVisitor2.java
```diff
@@ -0,0 +1,26 @@
...
+ *     Sierra Wireless - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.leshan.core.request;
+
+/**
+ * A new version of {@link DownlinkRequestVisitor} which supports {@link BootstrapDiscoverRequest}
+ * 
+ * @since 1.1
+ */
+public interface DownlinkRequestVisitor2 extends DownlinkRequestVisitor {
```
msangoi, line 23: Another solution could be to update the existing interface and provide a default implementation so that we don't need a new one. wdyt?
Same for LwM2mObjectEnabler.

sbernard31: (same reason : https://github.com/eclipse/leshan/pull/856#discussion_r451645352)

Issue #859, opened 2020-07-08
Roadmap to EST support


dachaac:
EST support is partly supported already in LWM2M specification 1.0.2 but is further improved in LWM2M specification 1.1.(1).

I have started writing wiki page with a bit more details in here:
https://github.com/dachaac/leshan/wiki/Improving-support-for-LWM2M-with-Certificates

Will try to make as small changes as possible as PRs to make integration work easier and also to get earlier feedback from other developers.

Will try to tag all related PRs with this ticket.

I believe the goal is achieved when the EST functionality is matching the latest released LWM2M specification (1.1.1 at the moment).

sbernard31:
Very interesting wiki page.
@msangoi you could be interested by this as you worked on [EST](https://github.com/eclipse/leshan/tree/est) by the past.

> Will try to make as small changes as possible as PRs to make integration work easier and also to get earlier feedback from other developers.
> Will try to tag all related PRs with this ticket.

:+1:

dachaac:
@sbernard31 I am currently modifying the web user interface for bootstrap server. 

When I click on "Add new client bootstrap configuration" it displays a configuration.

Let's assume we have a device with device certificate which is connecting (before auto registration support is ready) I was thinking I would do following:
- Fill in "client endpoint" field
- Select "LWM2M Bootstrap Server" tab
- Observation: Don't really understand why there is LWM2M Server URL in here. Can this be removed?
- For certificate auth -- I would select Security Mode: X.509 Certificate
- Now we get fields: Client Certificate and Client Private key
  - don't really understand why we have private key in here -- should really be removed.
- Now I can just copy client certificate here -- but the whole benefit with certificates is that you only need to care with CA certificates in configurations -- so should this be Client Certificate or Client Root CA Certificate? Or even empty and configure Client Root CA thru trust store?
- Observation: We have Server Certificate here as read only field -- can we remove this as one cannot really use the HEX value for anything in here?
- Then to configure LWM2M Server details for client I would go for "LWM2M Server" tab.
- I would fill in LWM2M Server URL setting
- I would select Security Mode as "X.509 Certificate with EST"
- Then in Server Certificate field I would paste the Server Certificate (or with upcoming Certificate Usage field having other value than default 3 the CA certificate)

Now related question:
[...]

PR #860, opened 2020-07-08
SecurityMode: Add missing "Certificate mode with EST" (4)

dachaac:
This adds missing EST security mode enum value.

Now the list should be complete vs. specification.

This change relates to: https://github.com/eclipse/leshan/issues/859

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/SecurityMode.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/SecurityMode.java
@@ -19,7 +19,7 @@ package org.eclipse.leshan.core;
  * The different DTLS security modes
  */
 public enum SecurityMode {
-    PSK(0), RPK(1), X509(2), NO_SEC(3);
+    PSK(0), RPK(1), X509(2), NO_SEC(3), EST(4);
 
     public final int code;
 


sbernard31:
> I don't want to cause you extra work on re-creating 2.0.x branch so that you might receive some contributions for that -- I don't personally see that model viable contributing method. I feel better to have volatile working branch for a feature and then just rebasing periodically with master to receive new changes from there.

Just to let you know, I will try to release the 1.1 soon and I will create the official new 2.0.x/master branch.
Do you want I create you a dedicate branch for EST ? name : `est_v11` ? (feel free to propose a better name ?)

dachaac:
I have my working branch in my fork:
https://github.com/dachaac/leshan/tree/est-support

I like to keep work-in-progress stuff in my fork and only integrate tidied up stuff. So no need to create a specific branch at this time. Let's see how it goes when we progress further.

This particular small change can go also with current version of master as it should not have any effect for anything else. But if you are about to make release feel free to wait with this.

sbernard31:
(now master is the 2.0.0 version which aims LWM2M 1.1, you should not be annoyed by revapi anymore, tell me if this is not case)

PR #869, opened 2020-07-31
LeshanServerBuilder: Fix handling of X.509 servers with serverAuth

dachaac:
When proper X.509 TLS Server Certificate is present it has extended key usage
specified as server auth.

Californium's DtlsConnectorConfig.setDefaultHandshakeMode() is a bit picky so
that it doesn't allow method to be called when in server only mode.

Adjust integration test certificates to have proper extended key usages
specified.


sbernard31:
I get the point about updating Certificate by adding basic constraint and key usage.
I also get the point about `HANDSHAKE_MODE_NONE`.

But didn't get why using `clientOnly` or `ServerOnly` in IntegrationTests ?

dachaac:
> I get the point about updating Certificate by adding basic constraint and key usage.
> I also get the point about `HANDSHAKE_MODE_NONE`.
> 
> But didn't get why using `clientOnly` or `ServerOnly` in IntegrationTests ?

ClientOnly and ServerOnly affects how key usage is verified. If they do not match the used certificate it torpedoes itself with exception which is rather hard to debug as that happens behind the scenes during integration testing.

Eg. if you have server certificate with serverAuth and you do not set it serverOnly it causes exception because there is no clientAuth in the certificate. Exception is triggered from Californium/DtlsConnectorConfig.java:2887.

sbernard31:
> ClientOnly and ServerOnly affects how key usage is verified. If they do not match the used certificate it torpedoes itself with exception which is rather hard to debug as that happens behind the scenes during integration testing.

Currently supporting both DTLS role (client&server) for both (device&server) is the only way we found to support failover on U mode. (https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/410)

So except there is a good reason, I would prefer to keep it activated by default meaning that I should probably have a deeper look at this issue at californium side.

(I will not be able to do that soon as I'm out of office the next 2 weeks)

dachaac:
> > ClientOnly and ServerOnly affects how key usage is verified. If they do not match the used certificate it torpedoes itself with exception which is rather hard to debug as that happens behind the scenes during integration testing.
> 
> Currently supporting both DTLS role (client&server) for both (device&server) is the only way we found to support failover on U mode. ([OpenMobileAlliance/OMA_LwM2M_for_Developers#410](https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/410))

If you want to do that with X.509 certificates the extended key usages need to match. Otherwise it fails on verifying connectivity. Skipping verification rules is a bit worrying in my opinion as they have been thought with different attack cases and skipping that could easily cause new problems.

Eg. if client is deployed with X.509 TLS Client Certificate and there is no DNS/IP Address in Subject / Subject Alternative Name it cannot be verified by LWM2M server when it is connecting to client. If it has additional storage to cache certificate details and IP/port pair then perhaps but I am not sure if this is a good idea.

In theory one could detect the mode from certificate details. If there is only clientAuth then it would set clientOnly, if there is only serverAuth then serverOnly, if both (or missing) then don't set any of those.

If you really want to do that kind of scenario then you could deploy X.509 Client Certificate as dual role (both clientAuth and serverAuth). In Subject DN/in SAN you would have the the Endpoint Client Name and then in other SAN entries you would have DNS/IP Address for the device.

SAN support is not implemented at this time in Leshan -- I have something for that in my local code for the future PRs.

Practical solution if you don't cache the DTLS state would be that client notices the disconnection and then initiates connection back when the DTLS ping/request fails/timeouts.

Caching the DTLS state information might even be better if you do "continuous delivery" with the server side software. In CD case you may want to separate DTLS connectivity from other functionality and only update DTLS connectivity side very rarely.

If you have battery powered device it is very important that you don't do the DTLS handshaking all the time. With battery powered device it might send data every hour or once a day or so. In this case of scenario the DTLS state cache might be the best solution. (in this scenario you most likely don't have thou static IP nor DNS address so if it is behind NAT that forgets mappings then there will be excessive DTLS handshakes -- IPv6 should fix that issue thou)

> So except there is a good reason, I would prefer to keep it activated by default meaning that I should probably have a deeper look at this issue at californium side.
> 
> (I will not be able to do that soon as I'm out of office the next 2 weeks)

Well the changes in the PR for common code allows one to do that -- so that it is not limiting. Code in master however doesn't allow normal X.509 TLS Client Certificate deployment to work so fix is needed to common code. Testing normal deployment model in my opinion is important (changes for certificates in this PR is matching the LWM2M 1.1 specification as best as I know). One can always add additional testing certificates and integration tests for dual role cases.

Actually DTLS connections should fail in integration tests as certificates do not have localhost/127.0.0.1 in subject/SANs for the server. But let's look at that later.

Whether there should be this server/client/dual role auto detection or additional changes to californium then I would say that is a different issue.

dachaac:
Note: Added matching changes to RedisSecureIntegrationTestHelper.java as future changes started failing and this is the right place to change that.

dachaac:
Speaking on future changes. I have cleaned up the code quite a bit and now latest certificate usage improvements are in here:
https://github.com/dachaac/leshan/commits/improve-certificate-support

I am wondering how should these be split for review and if there are some changes you would like.

boaks:
I'm not sure, if this discussion reflects just a misunderstanding, a gap in the functions, or real issues.

Let me start:

> Californium's DtlsConnectorConfig.setDefaultHandshakeMode() is a bit picky so
that it doesn't allow method to be called when in server only mode.

Historically (see [The LWM2M could make DTLS server role optional for LWM2M Clients](https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/2#issuecomment-71997830) ) Californium implemented the DTLS stack to enable peers to operate in both roles.
Practically, I got much more trouble with outgoing DTLS handshakes from servers, then outgoing handshakes had helped. For that reason I introduced the "server only", if I remember well, "client only" was already implemented. For the DTLS role exchange we then introduced the "handshake modes" to be able to control the start of  handshakes dynamically, and by that way, control the peer's role. But sure, if "server only" is used, the "handshake mode" doesn't make sense, because the server doesn't start handshakes at all. Or do I miss something?

> When proper X.509 TLS Server Certificate is present it has extended key usage
specified as server auth. 

According [RFC5280](https://tools.ietf.org/html/rfc5280#section-4.2.1.3) that stuff seems to be optional for me. So, it is checked for the role, if available, but it doesn't require it.
Californium applies that checks of certificates at two time point: configuration and usage/verification. For the configuration it's checked according the "client only", "server only", or "both roles" setup (assuming that it is optional). The same happens the at the verification of the other peers certificate in [AdvancedCertificateVerifier](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/x509/AdvancedCertificateVerifier.java#L46) (again, I assume it's optional). So, if you prefer to add a configuration to make these extension required, you may either open an issue (or PR) in Californium to do so, or implement your `AdvancedCertificateVerifier`according your requirements.

dachaac:
> I'm not sure, if this discussion reflects just a misunderstanding, a gap in the functions, or real issues.
> 
> Let me start:
> 
> > Californium's DtlsConnectorConfig.setDefaultHandshakeMode() is a bit picky so
> > that it doesn't allow method to be called when in server only mode.
> 
> Historically (see [The LWM2M could make DTLS server role optional for LWM2M Clients](https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/2#issuecomment-71997830) ) Californium implemented the DTLS stack to enable peers to operate in both roles.
> Practically, I got much more trouble with outgoing DTLS handshakes from servers, then outgoing handshakes had helped. For that reason I introduced the "server only", if I remember well, "client only" was already implemented. For the DTLS role exchange we then introduced the "handshake modes" to be able to control the start of handshakes dynamically, and by that way, control the peer's role. But sure, if "server only" is used, the "handshake mode" doesn't make sense, because the server doesn't start handshakes at all. Or do I miss something?

Well it could allow one to call the method with indication not to do a thing ;). But I don't really have preference in here. I see this as implementation detail -- I was partly surprised to see the quick kill in the method.

> > When proper X.509 TLS Server Certificate is present it has extended key usage
> > specified as server auth.
> 
> According [RFC5280](https://tools.ietf.org/html/rfc5280#section-4.2.1.3) that stuff seems to be optional for me. So, it is checked for the role, if available, but it doesn't require it.

Technically you can do all kinds of certificates -- however when you are about to operate PKI that should be properly managed then you are looking at the certificate design and make sure that they are according to requirements and best practices.

There is _RFC 7925 - Transport Layer Security (TLS) / Datagram Transport Layer Security (DTLS) Profiles for the Internet of Things_ which gives guidelines what one should do and it is referred also in LWM2M v1.1.1 transport specification as "The X.509 certificates SHOULD be conformant with the profiling given in [RFC7925]".

In [4.4.4. Certificate Content](https://tools.ietf.org/html/rfc7925#section-4.4.4) it states that both Key Usage and Extended Key Usage fields should be present.

Then on other relevant guidance side we have CABforum - they do have baseline requirements for CAs:
https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.7.0-1.pdf (latest version)

In chapter 7.1.2.3. Subscriber Certificate:
> f. extKeyUsage (required)
> Either the value id-kp-serverAuth [RFC5280] or id-kp-clientAuth [RFC5280] or both values MUST
> be present. id-kp-emailProtection [RFC5280] MAY be present. Other values SHOULD NOT be present.

Meaning that extendedKeyUsage should be present.

Then also in LWM2M v1.1.1 transport specification:
> 5.2.7. LwM2M and TLS/DTLS Roles
> In LwM2M version 1.1 the LwM2M Client is always the TLS/DTLS client. The LwM2M Bootstrap-Server and the LwM2M
Server are always acting as TLS/DTLS servers.

Which indicates what roles should be in use. And this can be used as indication what kind of certificates should be used. eg mapping to extendedKeyUsage for clients and for servers.

> Californium applies that checks of certificates at two time point: configuration and usage/verification. For the configuration it's checked according the "client only", "server only", or "both roles" setup (assuming that it is optional). The same happens the at the verification of the other peers certificate in [AdvancedCertificateVerifier](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/x509/AdvancedCertificateVerifier.java#L46) (again, I assume it's optional). So, if you prefer to add a configuration to make these extension required, you may either open an issue (or PR) in Californium to do so, or implement your `AdvancedCertificateVerifier`according your requirements.

I believe from Californium implementation point is is OK to assume that they are optional. In theory if you have client making connection it should just provide certificate out without thinking too much. It should only be part of the verification phase. In client it is only interested verifying that peer (the server) is having good certificate. And in Californium case I believe it should only follow implementation RFC. Eg. extendedKeyUsage could be preset and if so then verify that it has proper value depending on client/server role.

But when we are looking what kind of certificate should be made it is all about CA's on how they operate and on certificate profiles that are in use. In different protocols or in different deployment environments there can be differences.

boaks:
> In 4.4.4. Certificate Content it states that both Key Usage and Extended Key Usage fields should be present.

[RFC7924, 4.4.4](https://tools.ietf.org/html/rfc7925#section-4.4.4)

   +----------------------+--------------------------------------------+
   | Extension: Key Usage | The KeyUsage field MAY have the following  |
   |                      |   values in the context of this profile:   |
   |                      |     digitalSignature or keyAgreement,      |
   |                      |  keyCertSign for verifying signatures on   |
   |                      |          public key certificates.          |
   |                      |                                            |
   | Extension: Extended  |  The ExtKeyUsageSyntax field MAY have the  |
   |      Key Usage       |    following values in context of this     |
   |                      |    profile: id-kp-serverAuth for server    |
   |                      |    authentication, id-kp-clientAuth for    |
   |                      |  client authentication, id-kp-codeSigning  |
   |                      |   for code signing (for software update    |
   |                      |   mechanism), and id-kp-OCSPSigning for    |
   |                      |         future OCSP usage in TLS.          |
   +----------------------+--------------------------------------------+

I read `MAY`, not `should`, nor `MUST`.

Do I understand you right, that the implementation is OK, but you want other certificates being generated?
Aren't these demo certificates? It's up to you to use yours. So, I don't get your point.

> Well it could allow one to call the method with indication not to do a thing ;). But I don't really have preference in here. I see this as implementation detail -- I was partly surprised to see the quick kill in the method.

It would be easier to understand what you want, if you provide at least the names and not "the method". It would even be better, if you provide also the "setup" you want to apply. May be [Slightly relax rules for server only and handshake modes](https://github.com/eclipse/californium/commit/c653365ef5125ff7018c4b7000b84586307ebc8c) does it already for you.

dachaac:
> > In 4.4.4. Certificate Content it states that both Key Usage and Extended Key Usage fields should be present.
> 
> [RFC7925, 4.4.4](https://tools.ietf.org/html/rfc7925#section-4.4.4)
> 
> +----------------------+--------------------------------------------+
> | Extension: Key Usage | The KeyUsage field MAY have the following |
> | | values in the context of this profile: |
> | | digitalSignature or keyAgreement, |
> | | keyCertSign for verifying signatures on |
> | | public key certificates. |
> | | |
> | Extension: Extended | The ExtKeyUsageSyntax field MAY have the |
> | Key Usage | following values in context of this |
> | | profile: id-kp-serverAuth for server |
> | | authentication, id-kp-clientAuth for |
> | | client authentication, id-kp-codeSigning |
> | | for code signing (for software update |
> | | mechanism), and id-kp-OCSPSigning for |
> | | future OCSP usage in TLS. |
> +----------------------+--------------------------------------------+
> 
> I read `MAY`, not `should`, nor `MUST`.

You need to read a bit before the table too.
> All certificate elements listed in Table 1 MUST be implemented by clients and servers claiming support for certificate-based authentication.

And title for that table:
> Table 1 provides a summary of the elements in a certificate for use with this profile.

How I read this is that those elements in the table should be present AND then those text in boxes describe requirements for the actual values for the elements.

And in those particular cases valid values for the element are what is in MAY -- giving a hint what values to use in what use cases.

> Do I understand you right, that the implementation is OK, but you want other certificates being generated?
> Aren't these demo certificates? It's up to you to use yours. So, I don't get your point.

Well the Leshan didn't accept certificates that would be practically deployed which was a real problem here. It has been a problem with many projects that they have test certificates which are not just generated properly -> so why not test with properly generated certificates? I have no problem of also testing with more relaxed certificates but if you don't test how PKI experts design you certificate policy to use to generate your certificates you can have surprising problems in deployment scenarios.

CABforum in my view is the PKI expert forum with real Certificate Authorities and browser authors. Also in their baseline requirements we can see that in example extendedKeyUsage field must be present. Those baseline requirements are how CA's should operate.

Here are also more changes for certificate support for Leshan so that we can support real deployment scenarios for in example for EST cases as they have one or more intermediate CA certificates:
https://github.com/dachaac/leshan/commits/improve-certificate-support

Here you can see some example scripts to generate two example root CA's with intermediate CAs:
https://github.com/dachaac/device-certificate-scripts/tree/master

There is also leshan branch that create java key & trust stores. (a bit out of date that branch but for most part should give you idea)

That being said I believe in order to do testing for different kind of functionality there needs to be even more certificates and CAs. But those come when testing is added for those cases (and more features in Leshan is present)

In order to use those clientAuth only or serverAuth only certificates we need to configure serverOnly or clientOnly accordingly -- otherwise sanity tests are not passed and it throws an exception. So tests changed that too. This is also a bit of problem so it might be a good idea to just leave verification problem to other peer as there might be different kind of certificate policy deployments.

> > Well it could allow one to call the method with indication not to do a thing ;). But I don't really have preference in here. I see this as implementation detail -- I was partly surprised to see the quick kill in the method.
> 
> It would be easier to understand what you want, if you provide at least the names and not "the method". It would even be better, if you provide also the "setup" you want to apply. Maybe [Slightly relax rules for server only and handshake modes](https://github.com/eclipse/californium/commit/c653365ef5125ff7018c4b7000b84586307ebc8c) does it already for you.

With quick look in that commit I would say when you release californium with that changed code (and that is then taken in use in leshan) then that protective part is not required in leshan in order to use the certificates. Improving the certificate testing to check out problems like this has value in my opinion -- it also gives a hint to leshan users that this is how you can use the library.

boaks:
> Well the Leshan didn't accept certificates that would be practically deployed which was a real problem here.

Just to make sure, that it is not Californium: 
leshan denies such certificates, not the Californium's `DtlsConnectorConfig.Builder`, or?

About the MUST and MAY:
I read, the implementation MUST support the extensions, but MAY use it in the own certificates.
At least for Californium, that was my intention when implementing the update last December.

See also [[Uta] RFC7925, 4.4.4, interpretation, MUST-MAY](https://mailarchive.ietf.org/arch/msg/uta/jgmCuiRw7cx6CE-sdEvflegeZ5k/)

boaks:
I used the certificates of this PR without the java changes and so I see the trouble:
If Californium is configured for "dual-role", a provided certificate may be used by the stack for both roles, client and server. Therefore both is checked. If [RFC5280, 4.2.1.12.  Extended Key Usage](https://tools.ietf.org/html/rfc5280#section-4.2.1.12) is used, that usually decides, if the certificate is used for a client or server. If only one is present, californium requires to be configured for "single role", either by serverOnly or clientOnly. I'm not sure, if this is really a good idea, but to overcome that, someone may specify both extended key usages, server and client for the same certificate, if that is intended to be used in both roles. I tried it, with that the tests succeeded also with extended key usage. 

If the Californium is configured as "single-role", then the extended key usage works as usual.
But the handshake mode makes only sense for the client-role, for the server-role californium tolerates only the NONE (with 2.4). 

I hope, that makes it possible to use x509 certificates also with the extensions.

dachaac:
I believe you found the spot with https://github.com/eclipse/californium/commit/c22a5d6485c2458af209229c1604a8ba70ce8d84.

I do like the additional detail for exception.

So now the question I believe is -- should we wait for Californium 2.4 release? Is there any estimate how far away that is?

And then are we fine with the testing changes in here or is there some other changes to be done in Leshan side?

About the Californium implementation -- I believe you are right for actual code implementation in relation to Californium for MUSTs and MAYs. On this part I have no problem with current implementation in Californium. However the practicalities on how the certificates are actually done is completely different thing from implementation RFC. Changes in here adds testing for how one could expect certificates to be deployed in real life scenarios.

Here is example template from AWS Private CA how it creates client certificate:
https://docs.aws.amazon.com/acm-pca/latest/userguide/UsingTemplates.html#EndEntityClientAuthCertificate-V1

And for server certificate:
https://docs.aws.amazon.com/acm-pca/latest/userguide/UsingTemplates.html#EndEntityServerAuthCertificate-V1

And if you just want end entity certificate (which provides both):
https://docs.aws.amazon.com/acm-pca/latest/userguide/UsingTemplates.html#EndEntityCertificate-V1

boaks:
:-).

I can start the release process and in about 2 weeks the 2.4 will be released.

sbernard31:
If I well understand, this PR/discussion reveals an issue about DTLS role exchange and certificate which will be fixed in californium 2.4. (maybe should be backported in 2.3.1 this way we could integrate it in leshan 1.1.1)

About the PR itself, I see several modification.

1. fix the `serverOnly` + `handshake mode` issue.
2. adding basic constraints + key usage to all the certificates used by tests
3. add a new way to configure Leshan in DTLS Server Role only.

About 1. I understand this fix is needed for serverOnly use case.
About 2. what is the real benefits of changing all certificates in tests ? (having more "real world" certificates ?)
About 3. I'm not sure to get the benefits of this comparing in using `LeshanServerBuilder.setDtlsConfig`

dachaac:
> If I well understand, this PR/discussion reveals an issue about DTLS role exchange and certificate which will be fixed in californium 2.4. (maybe should be backported in 2.3.1 this way we could integrate it in leshan 1.1.1)
> 
> About the PR itself, I see several modification.
> 
> 1. fix the `serverOnly` + `handshake mode` issue.
> 2. adding basic constraints + key usage to all the certificates used by tests
> 3. add a new way to configure Leshan in DTLS Server Role only.
> 
> About 1. I understand this fix is needed for serverOnly use case.
> About 2. what is the real benefits of changing all certificates in tests ? (having more "real world" certificates ?)

Real benefit is that when you test with real world certificates you make sure that it works in real deployments and you can actually spot possible bug when software gets updated -- this is what I found here when I had properly made server and client certificates Leshan just didn't work properly. Thus -> lets make integration tests to have testing for this so that problem doesn't come back silently.

What one could add is auto detection how to configure dtls config based on given x.509 certificate. But I suppose that should be client/server project change only and not something done automatically by core. If you want that -- I can try to cook something for that?

> About 3. I'm not sure to get the benefits of this comparing in using `LeshanServerBuilder.setDtlsConfig`

I am not sure what you mean by about 3). When californium 2.4 is released then LeshanServerBuilder.java change is not needed. Before Californium 2.4 release it californium causes exception -- and the if there was to protect against exception.

sbernard31:
My current understanding is still that dual role is the only way to support failover when you are not using Queue-Mode. Even caching connection is not enough as you can imagine that you have limited-lifetime DTLS connection.
We begin to use it with PSK but we never use it with X509, so there is probably issue I don't see now.

As I consider dual role as really important to make LWM2M works, it's hard to me to consider the single role as the real use case.

Anyway, I will let you drive this and so I will merge this PR. We will see later if we need to re-introduce dual role in integration tests.

I will also back port the https://github.com/eclipse/leshan/pull/869/files#diff-ab2b2333ae602c4ca955af3e2805c488R532 fix in **1.x** branch

sbernard31:
integrated in master (commit 74af7a9ab91fef901fec42c04a057b309896b1cc) and in 1.x (commit 85bdaf96f13b282bdae385cf425e6bbc8fa37ffa)

sbernard31:
cf-2.4.0 is now integrated in master (commit 69e432c16ccca74f7a031cb2afa2998f8de05d5f )

Issue #729, opened 2019-08-14
Concerns about getSupportedObject impact on performance


sbernard31:
With new `VersionedModelProvider` from recently integrated #657 PR, the `getSupportedObject` will be called each time a Registration is extracted from Store.

This implies some questions about the about performance impact.
Here some result obtained with [jmh](https://openjdk.java.net/projects/code-tools/jmh/).

```
Benchmark (short is better)
(for shot registration ~4 objects)                                        Mode  Cnt      Score      Error  Units
get registration from store                                               avgt    3     75,454 ±    6,694  us/op
get registration from store + getSupportedObject                          avgt    3     90,996 ±  269,585  us/op
get registration from store + getSupportedObject with best alternative    avgt    3     75,726 ±   15,547  us/op

(for shot registration ~1000 objects)
get registration from store                                               avgt    3    747,277 ±  180,946  us/op
get registration from store + getSupportedObject                          avgt    3   1113,980 ±  836,456  us/op
get registration from store + getSupportedObject with best alternative    avgt    3    888,841 ±  669,591  us/op

JSON deserialization (total)                                              avgt    3    547,493 ±  245,424  us/op
JSON deserialization (string => minimal-json object)                      avgt    3    428,356 ±  110,013  us/op
JSON deserialization (minimal-json object => registration)                avgt    3    212,030 ±   63,589  us/op

getSupportedObject current implementation                                 avgt    3    250,588 ±   36,980  us/op
getSupportedObject using Scanner                                          avgt    3  16903,316 ± 1026,846  us/op
getSupportedObject using LWM2MPath (String.split)                         avgt    3    162,832 ±   18,660  us/op
getSupportedObject using StringTokenizer                                  avgt    3    110,568 ±    6,635  us/op
getSupportedObject using iterate over string                              avgt    3    105,756 ±   12,045  us/op
```

sbernard31:
Since 7ba642aab3ad40badcea86511dbbc31e0c364a74 : Remove lazy loading for getSupportedObjects in Registration.

This is no more relevant.

PR #879, opened 2020-08-27
Add new Json decoder based on Jackson

sbernard31:
Passing to jackson was driven by the need to **support Unsigned Long (64bits)** which is available through BigInteger support.

But there is other motivations : 
 - We also face some issue with minimal-json without any move from the maintainers => https://github.com/ralfstx/minimal-json/issues/88
- Jackson will be used for cbor maybe this could make the code more consistent. (and limit number of dependency)
- Jackson is considered as a realy fast json parser. 
  I need to do some tests to know if there are some benefits on this point (this could helps for https://github.com/eclipse/leshan/issues/729)


sbernard31:
I did some benchmarks about "Jackson" vs "Minimal" for LwM2mJson Encoder/Decoder.

```
> java -jar target/benchmark.jar -bm avgt -tu us -wi 5 -i 10 -f 1 -t1 Lwm2mJsonBench                                                                                                  

# JMH 1.17.4 (released 1318 days ago, please consider updating!)
# VM version: JDK 11.0.8, VM 11.0.8+10-post-Debian-1deb10u1
# VM invoker: /usr/lib/jvm/java-11-openjdk-amd64/bin/java
# VM options: <none>
# Warmup: 5 iterations, 1 s each
# Measurement: 10 iterations, 1 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op

# Run complete. Total time: 00:03:06

Benchmark                          Mode  Cnt    Score   Error  Units
Lwm2mJsonBench.jacksonDecode       avgt   10    5,311 ± 0,055  us/op
Lwm2mJsonBench.minimalDecode       avgt   10    7,240 ± 0,173  us/op

Lwm2mJsonBench.jacksonDecodeBig    avgt   10  300,307 ± 3,237  us/op
Lwm2mJsonBench.minimalDecodeBig    avgt   10  408,399 ± 4,041  us/op

Lwm2mJsonBench.jacksonDecodeSmall  avgt   10    0,998 ± 0,036  us/op
Lwm2mJsonBench.minimalDecodeSmall  avgt   10    0,988 ± 0,007  us/op

Lwm2mJsonBench.jacksonEncode       avgt   10    5,317 ± 0,046  us/op
Lwm2mJsonBench.minimalEncode       avgt   10    5,095 ± 0,131  us/op

Lwm2mJsonBench.jacksonEncodeBig    avgt   10  287,959 ± 1,452  us/op
Lwm2mJsonBench.minimalEncodeBig    avgt   10  311,546 ± 1,672  us/op

Lwm2mJsonBench.jacksonEncodeSmall  avgt   10    0,761 ± 0,026  us/op
Lwm2mJsonBench.minimalEncodeSmall  avgt   10    0,680 ± 0,005  us/op
```

In this use case Jackson is either faster or as fast as minimal-json.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
```diff
@@ -195,7 +195,7 @@ public void json_encode_device_object_instance() {
         b.append("{\"n\":\"9\",\"v\":100},");
         b.append("{\"n\":\"10\",\"v\":15},");
         b.append("{\"n\":\"11/0\",\"v\":0},");
-        b.append("{\"n\":\"13\",\"v\":1.367491215E9},");
+        b.append("{\"n\":\"13\",\"v\":1367491215},");
```
dachaac, line 198: Surprised to see this huge change in value? Are you sure that parsing here is working properly?

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
```diff
@@ -293,7 +293,7 @@ public void json_encode_timestamped_Object() throws CodecException {
         b.append("{\"n\":\"0/1\",\"v\":22.9,\"t\":210},");
         b.append("{\"n\":\"0/0\",\"sv\":\"a string\",\"t\":220},");
         b.append("{\"n\":\"0/1\",\"v\":22.4,\"t\":220},");
-        b.append("{\"n\":\"1/1\",\"v\":23,\"t\":220},");
+        b.append("{\"n\":\"1/1\",\"v\":23.0,\"t\":220},");
```
dachaac, line 296: This seems to be how jackson serializes double. Valid json and one could argue double is better to have decimal point.

I don't think is worrying just an effect of different library.

pom.xml
```diff
@@ -690,6 +690,17 @@ Contributors:
                 <artifactId>minimal-json</artifactId>
                 <version>0.9.5</version>
             </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
+                <version>2.9.9</version>
```
dachaac, line 696: Why not use latest stable release? (2.11.2)

pom.xml
```diff
@@ -690,6 +690,17 @@ Contributors:
...
+            </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.dataformat</groupId>
+                <artifactId>jackson-dataformat-cbor</artifactId>
+                <version>2.9.9</version>
```
dachaac, line 701: Why not upgrade to latest stable release? (2.11.2)

leshan-core/src/main/java/org/eclipse/leshan/core/util/json/JacksonJsonSerDes.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+
+/**
+ * An abstract class to easily create serializer/deserializer class based on "minimal-json".
```
dachaac, line 28: This comment is probably old? Or was this class supposed to use "minimal-json"?

dachaac:
If minmal-json is abandoned project -- why keep it in source tree?

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
sbernard31: I don't get it this is just 2 same way to write the "number".

1367491215 == 1.367491215E9

pom.xml
sbernard31: The reason was that I didn't change the version in this PR because, I would not impact cbor encoder/decoder in this PR.
But I'm not against changing this.

sbernard31:
> If minmal-json is abandoned project -- why keep it in source tree?

The idea is to remove dependency in the end, but I prefer to keep both encoder for now. (There is other source code which should be migrate, especially in demo we should not keep 3 dependency to Json libraries)

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
dachaac: Right. Sorry for confusion here.

pom.xml
dachaac: Ok. There are probably also other dependencies that should be upgraded that could be done in one go then.

msangoi (approved):

leshan-core/src/main/java/org/eclipse/leshan/core/json/jackson/JsonArrayEntrySerDes.java
```diff
@@ -0,0 +1,117 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Sierra Wireless and others.
```
dachaac, line 2: Should this be a bit more recent? (same for others)

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/text/LwM2mNodeTextDecoder.java
```diff
@@ -114,4 +119,4 @@ private static Object parseTextValue(String value, Type type, LwM2mPath path) th
             throw new CodecException("Could not handle %s value with TEXT encoder for resource %s", type, path);
         }
     }
-}
+}
```
dachaac, line 122: end-of-line lost?

leshan-core/src/main/java/org/eclipse/leshan/core/util/json/JacksonJsonSerDes.java
```diff
@@ -0,0 +1,69 @@
...
+    public abstract T deserialize(JsonNode o) throws JsonException;
+
+    public List<T> deserialize(Iterator<JsonNode> nodes) throws JsonException {
+        ArrayList<T> res = new ArrayList<>();
+        while (nodes.hasNext()) {
+            res.add(deserialize(nodes.next()));
+        }
+        return res;
+    }
+}
```
dachaac, line 69: no end-of-line?

dachaac:
Please check my minor comments but feel free to merge if you see it fit.

leshan-core/src/main/java/org/eclipse/leshan/core/json/jackson/JsonArrayEntrySerDes.java
sbernard31: I will fix it (bad copy/paste)

leshan-core/src/main/java/org/eclipse/leshan/core/util/json/JacksonJsonSerDes.java
sbernard31: Our current formatter config do not add it  (or removed it)  automatically. 

I looked at eclipse build-in formatter named : Eclipse, Eclipse 2.1 and Java Conventions and none of them add end-of-line automatically. So I guess it is ok for java file to not have end-of-line at the end. 

I didn't really no the pro/con argue about end of line ?

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/text/LwM2mNodeTextDecoder.java
sbernard31: Not really intended so I will re-add it.

pom.xml
sbernard31: This is updated in master now (commit 9ae537a4df120be160faaf0f66f858c5fe6e6e75)

PR #881, opened 2020-08-27
- added Resource Instance RW operations for Json Content Format

zakharenkodmytro:
Signed-off-by: Dmytro Zakharenko <dmytro.zakharenko@one.verizon.com>


sbernard31:
Just in case you're not aware about that : 
Instead of closing the PR and create a new one,  you can just push force on the same branch to edit it :)

zakharenkodmytro:
> Just in case you're not aware about that :
> Instead of closing the PR and create a new one, you can just push force on the same branch to edit it :)

I did not know it. Thanks, I'll try

PR #883, opened 2020-08-31
Refactor BootstrapConfigStore to use a list of "request"

sbernard31:
This aims to implements : #437

### DIFF
--- a/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/LeshanBootstrapServerDemo.java
+++ b/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/LeshanBootstrapServerDemo.java
@@ -46,6 +46,7 @@ import org.eclipse.leshan.core.model.ObjectLoader;
 import org.eclipse.leshan.core.model.ObjectModel;
 import org.eclipse.leshan.core.model.StaticModel;
 import org.eclipse.leshan.core.util.SecurityUtil;
+import org.eclipse.leshan.server.bootstrap.BootstrapConfigurationStoreAdapter;
 import org.eclipse.leshan.server.bootstrap.demo.servlet.BootstrapServlet;
 import org.eclipse.leshan.server.bootstrap.demo.servlet.ServerServlet;
 import org.eclipse.leshan.server.californium.LeshanServerBuilder;
@@ -330,7 +331,7 @@ public class LeshanBootstrapServerDemo {
         // Prepare and start bootstrap server
         LeshanBootstrapServerBuilder builder = new LeshanBootstrapServerBuilder();
         JSONFileBootstrapStore bsStore = new JSONFileBootstrapStore(configFilename);
-        builder.setConfigStore(bsStore);
+        builder.setConfigStore(new BootstrapConfigurationStoreAdapter(bsStore));
         builder.setSecurityStore(new BootstrapConfigSecurityStore(bsStore));
         builder.setModel(new StaticModel(models));
 
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/request/BootstrapDeleteRequest.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/request/BootstrapDeleteRequest.java
@@ -22,7 +22,8 @@ import org.eclipse.leshan.core.response.BootstrapDeleteResponse;
 /**
  * A LWM2M request for deleting object instances during the bootstrap phase.
  */
-public class BootstrapDeleteRequest extends AbstractDownlinkRequest<BootstrapDeleteResponse> {
+public class BootstrapDeleteRequest extends AbstractDownlinkRequest<BootstrapDeleteResponse>
+        implements BootstrapDownlinkRequest<BootstrapDeleteResponse> {
 
     /**
      * Creates a request for deleting all Instances of all Objects in the LwM2M Client (except LwM2M Bootstrap-Server
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/request/BootstrapDiscoverRequest.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/request/BootstrapDiscoverRequest.java
@@ -25,7 +25,8 @@ import org.eclipse.leshan.core.response.BootstrapDiscoverResponse;
  * 
  * @since 1.1
  */
-public class BootstrapDiscoverRequest extends AbstractDownlinkRequest<BootstrapDiscoverResponse> {
+public class BootstrapDiscoverRequest extends AbstractDownlinkRequest<BootstrapDiscoverResponse>
+        implements BootstrapDownlinkRequest<BootstrapDiscoverResponse> {
 
     /**
      * Creates a request for discovering all objects and instances supported by the client.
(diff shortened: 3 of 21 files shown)


sbernard31:
First thought about this : 

Currently,  to define what should be done during a bootstrap session, we need to create a [`BootstrapConfig`](https://github.com/eclipse/leshan/blob/leshan-1.1.0/leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/BootstrapConfig.java).

This is a bean in which you are able to define this actions in this order : 
- delete anything,
- write security objects,
- write server objects,
- write ACLs.

This has several limitation like :
- you can not choose the order of the requests,
- you can not write something else than security/server/acl,
- if models change you need to upgrade/change the `BootstrapConfig` bean and probably serialization. 

So replacing `BootstrapConfig` bean by a list of BootstrapDownlinkRequest sounds to resolve all those limitations, but this brings new ones : 
1. This makes bootstrap config validation very difficult. (it sounds difficult to have an equivalent for [ConfigurationChecker](https://github.com/eclipse/leshan/blob/leshan-1.1.0/leshan-server-core/src/main/java/org/eclipse/leshan/server/bootstrap/ConfigurationChecker.java))
2. Creating `LwM2mNode` content for `WriteRequest` is more painful, complicated and error prone than creating `BootstrapConfig`

So even if this is not so much satisfying to me, maybe we should keep both API... 
For issue 2. I also try a kind of experimental [API](https://github.com/eclipse/leshan/blob/bootstrap_refactoring/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mNodeBuilder.java) to make more simple LwM2node creation. 
At usage it looks like this : https://github.com/eclipse/leshan/pull/883/commits/7db537392c9a3381b0cc69f3172f4062788686ee#diff-cdf3b6c29cc5769b5054f2d8a660d7afR465

Working on this makes me think about other issues relative to Bootstrap server : 
1. Object Model Versioning : unlike to registration, devices does not begin bootstrap session by saying which version it supports. This could be a problem to found object models to use. (Actually I'm not sure if this is a real problem, if object models are backward compatible maybe have the last version of the object is enough)
=> If a solution is needed, we could imagine to start session with a Bootstrap discover request to know which version to use.
2. If bootstrap server want to apply a config dynamically  regarding of device states using Discover or Read request this is currently not possible. 
=> Maybe a solution could be that `BootstrapConfigurationStore` can return read/discover requests OR delete/write requests. If read/discover is returned, once we get the result we call again `BootstrapConfigurationStore` with read/discover responses in parameter. If delete/write is returned, once we apply all requests we send a bootstrap finished.

sbernard31:
@dachaac do you have any opinions on this ?

dachaac:
@sbernard31 kinda sounds into right direction and there might be a needs in certain installations to do custom object writing during bootstrapping -- at least it gives flexibility. What one need to make sure that all operations has been done successfully before bootstrap finish is issued.

About the implementation you have here -- didn't have time to look at it yet -- I will rebase my work on the master now that unsigned integer is there and push it for you for a preview on how I tried to go in between the bootstrapping process and connecting to lwm2m server. With that done we can see if there is something that should be taken into account and I can hopefully give some more though on this issue then.

dachaac:
@sbernard31 rebased my https://github.com/dachaac/leshan/commits/est-support branch.

It still needs tidying up and has some traces of ugliness while hacking in ;) and is missing CoAP-EST server stuff as I am wondering currently how to split it and where and what kind of interfaces would be good -- but I would say that is not related to this issue as such.

sbernard31:
> About the implementation you have here -- didn't have time to look at it yet

No problem I was mainly interested about your opinion on the idea itself and the new issues it could bring. (The code is currently no so clean it's more a kind of draft)

sbernard31:
I cleaned my code and limit the scope of this PR by letting the `bsserver-demo` untouched.

The new `BootstrapConfigurationStore`  replace the old `BootstrapConfigStore`(deprecated now).
About backward compatibility, `BootstrapConfigurationStoreAdapter` can be used.

About API issues raised in **https://github.com/eclipse/leshan/pull/883#issuecomment-687043565**
If we still want to keep the old `BootstrapConfig` bean for convenience, there is a `BootstrapUtil` utility to facilitate conversion to requests.
Another solution could be to explore a new API  like : this experimental [`LwM2mNodeBuilder`](https://github.com/eclipse/leshan/blob/node_builder/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mNodeBuilder.java) ...

About `bsserver-demo`, I didn't touch it for now, meaning it is using lot of deprecated stuff...

sbernard31:
I think this is in reviewable state now.

dachaac:
Lots of changes were just change from on way to another "in cosmetic category". Actual change seems to be the FIFO that gets generated from BootstrapConfig which is probably OK but then again you had BootstrapConfiguration which was planned to replace it?

I am not sure if BootstrapConfiguration is actually serializable in case it needs to be stored in database for persistence?

Or should this just be runtime thing and it gets generated from other config/code?

sbernard31:
> I am not sure if BootstrapConfiguration is actually serializable in case it needs to be stored in database for persistence?

This is not serializable as java.io.Serializable meaning, but this is just a list of bean so we should be able to serialize this list of requests.

> or should this just be runtime thing and it gets generated from other config/code?

Both are possible.
E.g :  In our case in production we don't persist BootstrapConfig or BootstrapConfiguration we create it from data extracted from different places

> Lots of changes were just change from on way to another "in cosmetic category". Actual change seems to be the FIFO that gets generated from BootstrapConfig which is probably OK but then again you had BootstrapConfiguration which was planned to replace it?

My first idea was to replace BootstrapConfiguration by BootstrapConfig to bring more flexibility.
But this brings new issues (see https://github.com/eclipse/leshan/pull/883#issuecomment-687043565)
So for now I didn't get rid off old `BootstrapConfig` (but if we can I would prefer) or maybe we need both, one for convenience on main bootstrap objects (security, server) and the other for all other objects.

dachaac:
Ok. I don't see problems with the changes in here.

Just a comment on deprecated stuff -- Both LWM2M Server Demo and LWM2M Bootstrap Server Demo has a Web UI. They are built with different frameworks. 

LWM2M Server is utilizing AngularJS which is going for EOL:
https://blog.angular.io/stable-angularjs-and-long-term-support-7e077635ee9c

LWM2M Bootstrap server then again is utilizing Riot.

On my other changes I just tried to adapt what ever was used but was a bit puzzled why the difference.

Do you have some plans on this sector?

sbernard31:
> Both LWM2M Server Demo and LWM2M Bootstrap Server Demo has a Web UI. They are built with different frameworks.

As demos are not the priority,  we don't invest as many time we would like.
There is a lot of technical debt and the more we add features to it the more it is hard to migrate.
I will create a dedicated issue about current identified problem with our demos.

>  why the difference.

This is just historical reasons + lack of time.
`Angular` was chosen at the beginning, I don't know why (I was not in the project at this time) but the team was really disappointing by it. So when we create the bootstrap server demo we try something else : `Riot`.
My personal feeling is `Riot` is not so bad, but maybe not enough popular and so could be an obstacle for contribution.

sbernard31:
see https://github.com/eclipse/leshan/issues/907

msangoi (approved):
LGTM

PR #891, opened 2020-09-14
Fix issue avoiding to choose mix ciphersuite psk/rpk/x509 with client

sbernard31:
When you build a LeshanClient you can provide a DTLSConfig. (see [LeshanClientBuilder](https://github.com/eclipse/leshan/blob/leshan-1.1.0/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java#L151))
This config will be not used directly but will serve as base to create endpoint in [CaliforniumEndpointsManager](https://github.com/eclipse/leshan/blob/leshan-1.1.0/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java#L86).

A use case could be to use a custom dtls config when you create your leshan client to limit cipher suites you want to use.
You will choose cipher compatible for psk, x509 and rpk, because a bootstrap server could set any kind 

If you are using this ciphersuite config directly Californium/Scandium will raise an exception because config is not consistent (e.g. you have "rpk" cipher either only psk is configured)

To fix it you need to filter cipher which is done in this PR.

(I didn't check yet but I suspect we could face same kind of issue for other consistency issue about dtlsconfig)

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java
@@ -22,6 +22,7 @@ import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 
 import org.eclipse.californium.core.CoapServer;
 import org.eclipse.californium.core.network.CoapEndpoint;
@@ -37,6 +38,7 @@ import org.eclipse.californium.scandium.dtls.AlertMessage.AlertLevel;
 import org.eclipse.californium.scandium.dtls.CertificateMessage;
 import org.eclipse.californium.scandium.dtls.DTLSSession;
 import org.eclipse.californium.scandium.dtls.HandshakeException;
+import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
 import org.eclipse.californium.scandium.dtls.pskstore.StaticPskStore;
 import org.eclipse.californium.scandium.dtls.rpkstore.TrustedRpkStore;
 import org.eclipse.californium.scandium.dtls.x509.CertificateVerifier;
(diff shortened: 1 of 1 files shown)


sbernard31:
@boaks, do you think there is something to do at californium side to make this use case more easy ?
(I mean using dtlsconfig as a kind of super-set of configuration setting ?)

boaks:
I'm not sure, if that ends up in a dynamic "cipher suite selection" based on the returns of PskStore, CertificateVerifier etc. . I guess, that would require much more changes.
I'm not sure, if it is not easier, to destroy and create new dtls connectors, though these bootstrap credentials may not change too frequently.

sbernard31:
@boaks, We already destroy and create new dtls connectors after each bootstrap.

The issue here is that we use a common dtls config (psk/x509/prk) which is used as base to create the real specific config.
And for cipher suite, I need to filter it (see this PR code) before to create connector else builder raise exception (which makes sense)

So the need would not to have a dynamic "cipher suite selection" but rather to have a kind of "smart" static creation of dtlsconfig. Like it's done when you don't set any cipher suite. 
(or maybe this is exactly what you mean by dynamic? :sweat_smile: )

I can continue to do the smart part in Leshan but I just ask if this should not make more sense to propose something in scandium directly.

The issue is that I have no clear idea about how this could looks like.
Maybe adding a new setter which would be : setDefaultCipherSuitePool which will be used in `DtlsConnectorConfig.Builder.determineCipherSuitesFromConfig()` ?

(I hope I'm clearer)

boaks:
Ah, OK. It's more about being able to limit the cipher suites without selecting them.

sbernard31:
>  It's more about being able to limit the cipher suites without selecting them.

Yep exactly ... it sounds so simple saying like this :sweat_smile:

boaks:
`setSupportedCipherSuites` ?

Failing, if it's not supported by the JCE.

sbernard31:
Not sure to get the question ? :sweat_smile:

boaks:
Does the name `setSupportedCipherSuites` fit for your?

sbernard31:
This is the current name or ?

boaks:
Uuups ... you're right.
Ok, then "Pool".

sbernard31:
(I merge this PR in master () and 1.x ())
Code would be adapted later if/when a new Scandium API will be available.

boaks:
@sbernard31 
Please check, if [Californium - PR1416](https://github.com/eclipse/californium/pull/1416) works for you. It introduces "preselected cipher suites".

PR #900, opened 2020-10-06
Add X509CertUtil helper class for handling X.509 certificate details

dachaac:
This will later be utilized in more later on for in example subjectAltName support. Test suite is added to exercise Subject DN parsing.

It also improves SecurityTesting to use correct end point names as it will be more strict with later changes.


sbernard31:
I will look at this tomorrow

sbernard31:
I didn't find time today :sweat_smile: ... tomorrow my first task.

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
```diff
@@ -0,0 +1,310 @@
...
+    private static boolean isHex(char ch) {
+        return (ch >= '0' && ch <= '9' || ch >= 'A' && ch <= 'F' || ch >= 'a' && ch <= 'f');
+    }
+
+    /**
+     * Parses RFC 2253 name string.
+     * <p>
+     * Extracts field keys and their values in form as they are given in name string.
+     *
+     * @param name RFC 2254 name string
```
sbernard31, line 126:
```suggestion
     * @param name RFC 2253 name string
```
Typo ?

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
```diff
@@ -0,0 +1,310 @@
...
+     *
+     * @param ch Char to test
+     * @return true if HEX char, false otherwise.
+     */
+    private static boolean isHex(char ch) {
+        return (ch >= '0' && ch <= '9' || ch >= 'A' && ch <= 'F' || ch >= 'a' && ch <= 'f');
+    }
+
+    /**
+     * Parses RFC 2253 name string.
```
sbernard31, line 122: are you referring to : https://tools.ietf.org/html/rfc2253#section-3 ?

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
```diff
@@ -0,0 +1,310 @@
...
+     * @return True if match, false otherwise
+     */
+    public static boolean matchSubjectDnsName(X509Certificate certificate, String dnsName) {
+        try {
+            // First one need to check SANs if they are present
+
+            final Collection<List<?>> sans = certificate.getSubjectAlternativeNames();
+
+            if (sans != null) {
+                for (final List<?> san : sans) {
```
sbernard31, line 250: Generally we avoid final key word for variable or parameter if this is not needed.
Any reason you use it ?

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
```diff
@@ -0,0 +1,310 @@
...
+                    if (generalName == GeneralName.DNS_NAME.value) {
+                        String value = (String) san.get(1);
+                        if (dnsNameMatch(value, dnsName)) {
+                            return true;
+                        }
+                    }
+                }
+            }
+
+            // Then one can fallback to old ways at looking in Subject DN
```
sbernard31, line 261: Should we fallback :
-  only if san/dnsname does not exist 
OR 
- even if it exists and it does'nt match ?

sbernard31 (commented):
Very clean code ! :+1:

sbernard31:
1. Not directly linked to this PR but I asked myself if we should move `certificates` folder to `src/test/resources` ? :thinking: 
(If we do that, we need to do it for `credentials` folder in `leshan-integration-tests` too)

2. About commit `SecurityTest: Fix X.509 end point name comparison` (52f3a4d), I have a vague memory that having different endpoint name for tests was important but I can remember why ...
I ran the tests locally (even using -Predis) and it sounds green :heavy_check_mark:. So I suppose we can go with this, we will see later if this bring issues.

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
```diff
@@ -0,0 +1,310 @@
...
+import java.security.cert.CertificateParsingException;
+import java.security.cert.X509Certificate;
+import java.util.*;
+
+/**
+ * X.509 Certificate Utilities for accessing certificate details.
+ */
+public class X509CertUtil {
+    /**
+     * rfc3280#section-4.2.1.7 - GeneralName
```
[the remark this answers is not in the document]
dachaac: OK.

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
dachaac: Will remove. Probably habit from some other programming language.

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
dachaac: yes, in simplified value form, will add reference here too.

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
dachaac: seems so, will fix.

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
dachaac: https://tools.ietf.org/html/rfc6125 seems to address the issue how it should be interpreted.

> Does your technology need to use DNS domain names in the Common
> Name of certificates for the sake of backward compatibility?  If
> so, consider recommending support for the CN-ID identifier type as
> a fallback."

dachaac:
All comments should be now addressed.

dachaac:
> 1. Not directly linked to this PR but I asked myself if we should move `certificates` folder to `src/test/resources` ? thinking
>    (If we do that, we need to do it for `credentials` folder in `leshan-integration-tests` too)

I think those will then get into test JAR's which is probably not something that we want?

> 2. About commit `SecurityTest: Fix X.509 end point name comparison` ([52f3a4d](https://github.com/eclipse/leshan/commit/52f3a4d43be66f77cc6cf5ebc372dd1c6ba78c25)), I have a vague memory that having different endpoint name for tests was important but I can remember why ...
>    I ran the tests locally (even using -Predis) and it sounds green heavy_check_mark. So I suppose we can go with this, we will see later if this bring issues.

If you find out the reason then we need to craft fixed amount of certificates or issue them on fly with BouncyCastle or so.

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
sbernard31: Let me try to reformulate because maybe I was not clear.

Imagine you want to check "server.mydomain.com"  domain.
And you have a certificate with SAN=dns:**another**.mydomain.com and CN=server.mydomain.com
Should we consider that as a match or not ?

sbernard31:
>  think those will then get into test JAR's which is probably not something that we want?

I don't know. If the purpose of this jar is to be able to run unit-test, I suppose it need to embed resources needed for tests.
(Anyway clearly not an essential point)

leshan-core/src/main/java/org/eclipse/leshan/core/util/X509CertUtil.java
sbernard31: I just raise this question. We don't need to answer to this question quickly.
I will merge the PR in master.

dachaac: Yes, I understood your question.

SAN is list any-match-here is good.

In purest form answer would be no match for server.mydomain.com -- as if SAN is present it should be used as single truth.

However people is still trying to learn how the SAN should be configured so starting with SAN -> CN fallback might be the good start to make sure that no existing installations gets broken.

Making it more strict is probably a good idea and I believe some software are like that already. Making it more strict right away might be also a good option to force people to make things correctly.

When looking form openssl:
https://www.openssl.org/docs/man1.1.1/man3/X509_check_host.html

It says that any CN or SAN is good enough in generic text but then again has control flags that can flex or strict the operation. And even says multiple default operations that are a bit conflicting ;).

sbernard31: Maybe we should start with strict mode and if some user complain we will try to adapt it (make it configurable)?

dachaac: Sure. I will make a PR for that.

dachaac: And the PR: https://github.com/eclipse/leshan/pull/910

PR #903, opened 2020-10-07
DefaultEndpointFactory: Add new endpoint context matcher for clients

dachaac:
There is separation between LWM2M Client and Server side
EndpointContextMatchers as when client is performing initial connection it
doesn't know real peer identity due to flexibility of certificate usage
configuration.

With certificate usage the server identity may be the intermediate CA or root
CA. LWM2M client also knows where it is connecting and peer identity is
verified for DTLS secured connections during DTLS handshake phase.

When server is receiving DTLS connection it always knows the whole peer
identity as DTLS handshake is performed first before any message is being sent
to client and then EndpointContext is built with real client information.
Matcher is needed in server side as one datagram server socket received
connections from multiple clients.

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
@@ -257,7 +257,7 @@ public class LeshanClientBuilder {
             engineFactory = new DefaultRegistrationEngineFactory();
         }
         if (endpointFactory == null) {
-            endpointFactory = new DefaultEndpointFactory("LWM2M Client") {
+            endpointFactory = new DefaultEndpointFactory("LWM2M Client", true) {
                 @Override
                 protected Connector createSecuredConnector(DtlsConnectorConfig dtlsConfig) {
                     DTLSConnector dtlsConnector = new DTLSConnector(dtlsConfig);
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
@@ -596,7 +596,7 @@ public class LeshanClientDemo {
         engineFactory.setResumeOnConnect(!forceFullhandshake);
 
         // configure EndpointFactory
-        DefaultEndpointFactory endpointFactory = new DefaultEndpointFactory("LWM2M CLIENT") {
+        DefaultEndpointFactory endpointFactory = new DefaultEndpointFactory("LWM2M CLIENT", true) {
             @Override
             protected Connector createSecuredConnector(DtlsConnectorConfig dtlsConfig) {
 
--- a/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/DefaultEndpointFactory.java
+++ b/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/DefaultEndpointFactory.java
@@ -16,6 +16,7 @@
 package org.eclipse.leshan.core.californium;
 
 import java.net.InetSocketAddress;
+import java.security.Principal;
 
 import org.eclipse.californium.core.network.CoapEndpoint;
 import org.eclipse.californium.core.network.CoapEndpoint.Builder;
@@ -24,6 +25,7 @@ import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.core.observe.ObservationStore;
 import org.eclipse.californium.elements.Connector;
 import org.eclipse.californium.elements.EndpointContextMatcher;
+import org.eclipse.californium.elements.PrincipalEndpointContextMatcher;
 import org.eclipse.californium.elements.UDPConnector;
 import org.eclipse.californium.scandium.DTLSConnector;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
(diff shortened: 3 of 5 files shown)


sbernard31:
As said more in details here : https://github.com/eclipse/leshan/pull/896#issuecomment-701296503

As we now have 1 connector/endpoint by server at client side, I think that if our connector configuration is strict enough to accept only the targeted server, we don't need so many logic in `EndpointContextMatcher`.

We can consider that principal is OK  and so don't check it.

@boaks, do you think it makes sense ? :point_up:

dachaac:
With this tests pass:
```java
    /**
     * {@inheritDoc}
     * 
     * LWM2M x.509 principals has already been checked during TLS handshake. 
     */
    @Override
    protected boolean matchPrincipals(Principal requestedPrincipal, Principal availablePrincipal) {
        if (requestedPrincipal instanceof X500Principal && availablePrincipal instanceof X509CertPath) {
            // Note: DTLS layer has already verified that certificate usage matches peer and that certificates are valid
            return true;
        } else {
            // Handle non X.509 identities in common LWM2M endpoint context matcher
            return super.matchPrincipals(requestedPrincipal, availablePrincipal);
        }
    }
```

dachaac:
@sbernard31 updated the PR with simplified matcher.

sbernard31:
My idea was to remove check for all principal, not x509 only.
But `org.eclipse.leshan.integration.tests.SecurityTest.dont_sent_request_if_identity_change()` failed.
My guess it should not, so I need to investigate this.

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/Lwm2mClientEndpointContextMatcher.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    @Override
+    protected boolean matchPrincipals(Principal requestedPrincipal, Principal availablePrincipal) {
+        if (requestedPrincipal instanceof X500Principal && availablePrincipal instanceof X509CertPath) {
+            // Note: DTLS layer has already verified that certificate usage matches peer and that certificates are valid
+            return true;
+        } else {
+            // Handle non X.509 identities in common LWM2M endpoint context matcher
+            return super.matchPrincipals(requestedPrincipal, availablePrincipal);
+        }
```
sbernard31, line 69:
```suggestion
  return true
```

As I think this is enough. (as explained in https://github.com/eclipse/leshan/pull/903#issuecomment-706158960)
I tested and test pass on my side. (as long as this modification is only apply at client side)

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/Lwm2mClientEndpointContextMatcher.java
```diff
@@ -0,0 +1,71 @@
...
+ * there is DTLS Handshake which has verified that certificate usage and chain is expected.
+ *
+ * When server is receiving DTLS connection it always knows the whole peer identity as DTLS handshake is performed first
+ * before any message is being sent to client and then EndpointContext is built with real client information. In server
+ * more lightweight and strict matching can be done there and there is no need to inspect the chain.
+ * 
+ * For more detailed about why this is needed, see
+ * <a href="https://github.com/eclipse/leshan/wiki/LWM2M-Observe#for-dtls">LWM2M-Observe wiki page</a>
+ */
+public class Lwm2mClientEndpointContextMatcher extends Lwm2mEndpointContextMatcher {
```
sbernard31, line 46: I don't know why but  LWM2M in class name is written like this : `LwM2m` :exploding_head: 

I can understand this way to write LWM2M does not make so much sense, so we could eventually change it if we think this is worth (but for the whole project in another PR).

So there is typo in `Lwm2mEndpointContextMatcher` and `Lwm2mEndpointContextMatcher`.

But maybe we don't even need a dedicated class a simple inline class is maybe enough ?

If we decide to keep the class : 
- I think the javadoc should updated too. (The javadoc explanation should be something like this : https://github.com/eclipse/leshan/pull/903#issuecomment-706158960
- I think that it should directly implement `PrincipalEndpointContextMatcher`
- `Lwm2mEndpointContextMatcher` should probably be renamed in `LwM2mServerEndpointContextMatcher`

sbernard31:
I'm pretty sure https://github.com/eclipse/leshan/pull/903#issuecomment-706158960 is OK.
But I would really appreciate to get @boaks  feedback on this point before to merge.

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/Lwm2mClientEndpointContextMatcher.java
dachaac: Removed the extra class and made it anonymous class.

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/Lwm2mClientEndpointContextMatcher.java
dachaac: Simplified as suggested.

sbernard31:
It sounds good, just waiting few days in case we get @boaks answer.

boaks:
The intention of this customizable matcher was, that every project is able, to decide that on it's own.
Though I'm not deep enough in LwM2M, I don't think, that I will be too much of help.

sbernard31:
This is not specially about LWM2M, this is generic question : 
In case where you want to match by principal (e.g. LWM2M)
If connector/endpoint is configured to only accept 1 foreign peer. (e.g. it trusts only 1 PSK or only 1 RPK  ...), is matching principal in endpointcontext still useful ? 

I think it does not.

boaks:
> is matching principal in endpointcontext still useful ?

> I think it does not.

Using the matcher, verifies that assumption using additional data, including crypto artifacts.
Without, it's just the assumption. So using the matcher may ensure, that you not tricked by an adversary. But, if that is not considered/assumed, you may leave it out.

sbernard31:
> Using the matcher, verifies that assumption using additional data, including crypto artifacts.

Yep we keep that (as we inherit Principal matcher). We just don't verify the principal part.
Anyway, thx @boaks for your feedback.

I think we can integrate this PR.

sbernard31:
Integrate in master (commit 091ba2038539eca18bf96b9035fd849a05b00ca9)
I changed some wording in comment about why matching principal is not needed at client side.

PR #905, opened 2020-10-09
Trust store updates

dachaac:
These commits adds support to load trust store form Java trust store as what is common in Java applications.

Also adds support for LeshanClient to have trust store defined -- actual usage comes with later changes.

For use of Java trust store argument format utilizes Californium's URI format for specifying trust store.
    
-truststore
file:///full/path/truststore.p12#hex-store-password#aliaspattern

Where aliaspattern can be empty.

First this is added to LeshanServerDemo and then same logic is applied to LeshanClientDemo and to LeshanBootstrapServerDemo.

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
@@ -16,6 +16,7 @@
 package org.eclipse.leshan.client.californium;
 
 import java.net.InetSocketAddress;
+import java.security.cert.Certificate;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ScheduledExecutorService;
@@ -61,6 +62,7 @@ public class LeshanClientBuilder {
 
     private NetworkConfig coapConfig;
     private Builder dtlsConfigBuilder;
+    private List<Certificate> trustStore;
 
     private LwM2mNodeEncoder encoder;
     private LwM2mNodeDecoder decoder;
@@ -153,6 +155,15 @@ public class LeshanClientBuilder {
      */
     public LeshanClientBuilder setDtlsConfig(DtlsConnectorConfig.Builder config) {
         this.dtlsConfigBuilder = config;
+        return this;
+    }
+
+    /**
+     * Set optional trust store for verifying X.509 server certificates.
+     * @param trustStore List of trusted CA certificates
+     */
+    public LeshanClientBuilder setTrustStore(List<Certificate> trustStore) {
+        this.trustStore = trustStore;
         return this;
     }
 
@@ -309,7 +320,7 @@ public class LeshanClientBuilder {
         }
 
         return createLeshanClient(endpoint, localAddress, objectEnablers, coapConfig, dtlsConfigBuilder,
-                endpointFactory, engineFactory, additionalAttributes, encoder, decoder, executor);
+                this.trustStore, endpointFactory, engineFactory, additionalAttributes, encoder, decoder, executor);
     }
 
     /**
@@ -326,6 +337,7 @@ public class LeshanClientBuilder {
      *        client.
      * @param coapConfig The coap config used to create {@link CoapEndpoint} and {@link CoapServer}.
      * @param dtlsConfigBuilder The dtls config used to create the {@link DTLSConnector}.
+     * @param trustStore The optional trust store for verifying X.509 server certificates.
      * @param endpointFactory The factory which will create the {@link CoapEndpoint}.
      * @param engineFactory The factory which will create the {@link RegistrationEngine}.
      * @param additionalAttributes Some extra (out-of-spec) attributes to add to the register request.
@@ -337,10 +349,11 @@ public class LeshanClientBuilder {
      */
     protected LeshanClient createLeshanClient(String endpoint, InetSocketAddress localAddress,
             List<? extends LwM2mObjectEnabler> objectEnablers, NetworkConfig coapConfig, Builder dtlsConfigBuilder,
-            EndpointFactory endpointFactory, RegistrationEngineFactory engineFactory,
+            List<Certificate> trustStore, EndpointFactory endpointFactory, RegistrationEngineFactory engineFactory,
             Map<String, String> additionalAttributes, LwM2mNodeEncoder encoder, LwM2mNodeDecoder decoder,
             ScheduledExecutorService sharedExecutor) {
-        return new LeshanClient(endpoint, localAddress, objectEnablers, coapConfig, dtlsConfigBuilder, endpointFactory,
-                engineFactory, additionalAttributes, bsAdditionalAttributes, encoder, decoder, executor);
+        return new LeshanClient(endpoint, localAddress, objectEnablers, coapConfig, dtlsConfigBuilder, trustStore,
+                endpointFactory, engineFactory, additionalAttributes, bsAdditionalAttributes, encoder, decoder,
+                executor);
     }
 }
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
@@ -26,15 +26,10 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
+import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.security.interfaces.ECPublicKey;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Scanner;
+import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -47,6 +42,7 @@ import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.elements.Connector;
+import org.eclipse.californium.elements.util.SslContextUtil;
 import org.eclipse.californium.scandium.DTLSConnector;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
 import org.eclipse.californium.scandium.dtls.ClientHandshaker;
@@ -223,6 +219,21 @@ public class LeshanClientDemo {
                 "The path to your client certificate file.\n The certificate Common Name (CN) should generaly be equal to the client endpoint name (see -n option).\nThe certificate should be in X509v3 format (DER encoding).");
         options.addOption("scert", true,
                 "The path to your server certificate file.\n The certificate should be in X509v3 format (DER encoding).");
+
+        final StringBuilder trustStoreChapter = new StringBuilder();
+        trustStoreChapter.append("\n .");
+        trustStoreChapter.append("\n URI format: file://<path-to-trust-store-file>#<hex-strore-password>#<alias-pattern>");
+        trustStoreChapter.append("\n .");
+        trustStoreChapter.append("\n Where:");
+        trustStoreChapter.append("\n - path-to-trust-store-file is path to pkcs12 trust store file");
+        trustStoreChapter.append("\n - hex-store-password is HEX formatted password for store");
+        trustStoreChapter.append("\n - alias-pattern can be used to filter trusted certificates and can also be empty to get all");
+        trustStoreChapter.append("\n .");
+        trustStoreChapter.append("\n Default: All certificates are trusted which is only OK for a demo.");
+
+        options.addOption("truststore", true,
+                "The path to a root certificate file to trust or a folder containing all the trusted certificates in X509v3 format (DER encoding) or trust store URI."
+                        + trustStoreChapter);
 
         HelpFormatter formatter = new HelpFormatter();
         formatter.setWidth(90);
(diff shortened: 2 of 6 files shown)


dachaac:
Note: it seems some rebasing mayhem lost change for actually setting trust store for client. Will update the PR.

dachaac:
Rebased to master and fixed LeshanClientDemo's passing of trust store.

dachaac:
Rebased to master.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
```diff
@@ -163,8 +161,21 @@ public static void main(String[] args) throws FileNotFoundException {
                 "The path to your server certificate file.\n"
                         + "The certificate Common Name (CN) should generally be equal to the server hostname.\n"
                         + "The certificate should be in X509v3 format (DER encoding).");
+
+        final StringBuilder trustStoreChapter = new StringBuilder();
+        trustStoreChapter.append("\n .");
+        trustStoreChapter.append("\n URI format: file://<path-to-trust-store-file>#<hex-strore-password>#<alias-pattern>");
+        trustStoreChapter.append("\n .");
+        trustStoreChapter.append("\n Where:");
+        trustStoreChapter.append("\n - path-to-trust-store-file is path to pkcs12 trust store file");
```
sbernard31, line 170: it seems that californium API support more than pkcs12.
Should we be more specific, by listing [all supported format](https://github.com/eclipse/californium/blob/2.4.1/element-connector/src/main/java/org/eclipse/californium/elements/util/SslContextUtil.java#L578) ?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -22,19 +22,16 @@
 import static org.eclipse.leshan.core.LwM2mId.*;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateEncodingException;
```
sbernard31, line 31: seems unused

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
```diff
@@ -309,7 +320,7 @@ protected Connector createSecuredConnector(DtlsConnectorConfig dtlsConfig) {
         }
 
         return createLeshanClient(endpoint, localAddress, objectEnablers, coapConfig, dtlsConfigBuilder,
-                endpointFactory, engineFactory, additionalAttributes, encoder, decoder, executor);
+                this.trustStore, endpointFactory, engineFactory, additionalAttributes, encoder, decoder, executor);
```
sbernard31, line 323: why not setting truststore in dtlsConfigBuilder ?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -456,6 +469,47 @@ public static void main(final String[] args) {
...
+                    e.printStackTrace();
+                    formatter.printHelp(USAGE, options);
+                }
+            } else {
+                // Treat argument as file or directory
+                File input = new File(cl.getOptionValue("truststore"));
+
+                // check input exists
+                if (!input.exists())
+                    throw new FileNotFoundException(input.toString());
```
sbernard31, line 494: I know this is like this in leshan-server-demo but it sounds not so consistent.

All errors seems to rather be handled by a : 
```
System.err.println("Error message" );
formatter.printHelp(USAGE, options);
return;
```

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
```diff
@@ -299,24 +310,40 @@ public static void main(String[] args) throws FileNotFoundException {
...
+                // Treat argument as Java trust store
                 try {
-                    trustStore.add(SecurityUtil.certificate.readFromFile(file.getAbsolutePath()));
+                    Certificate[] trustedCertificates = SslContextUtil.loadTrustedCertificates(trustStoreName);
+                    trustStore.addAll(Arrays.asList(trustedCertificates));
                 } catch (Exception e) {
-                    LOG.warn("Unable to load X509 files {}:{} ", file.getAbsolutePath(), e.getMessage());
+                    System.err.println("Unable to load trust store : " + e.getMessage());
+                    e.printStackTrace();
+                    formatter.printHelp(USAGE, options);
```
sbernard31, line 324: missing `return;` ?

sbernard31:
What is your opinion about the deprecated way to set x509 in `leshan-server-demo` ? 
```
                                =======================[ X509 deprecated way]===========================
                                | By default Leshan demo uses an embedded self-signed certificate and  |
                                | trusts any client certificates allowing to use RPK or X509           |
                                | at client side.                                                      |
                                | If you want to use your own server keys, certificates and truststore,|
                                | you can provide a keystore using :                                   |
                                |         -ks, -ksp, [-kst], [-ksa], -ksap should be used together     |
 -ks,--keystore <arg>           Set the key store file.
                                If set, X.509 mode is enabled, otherwise built-in RPK credentials are used.
 -ksp,--storepass <arg>         Set the key store password.
 -kst,--storetype <arg>         Set the key store type.
                                Default: pkcs12.
 -ksa,--alias <arg>             Set the key store alias to use for server credentials.
                                Default: leshan.
                                All other alias referencing a certificate will be trusted.
 -ksap,--keypass <arg>          Set the key store alias password to use.
```
Should we remove it ?

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
dachaac: Never heard anyone using BKS. So perhaps not worth to mention.

JKS is Java KeyStore and is deprecated in favour of PKCS12 key store in Java tooling. But it is still known by people so we could add that to the list.

Perhaps:
```
path-to-trust-store-file is path to PKCS#12 (.p12) or Java trust store (.jks) file
```
Problem with PEM is that it is just encoding and is often not named as .pem like in Ubuntu CA certificates bundle (PEM) is named:
```
/etc/ssl/certs/ca-certificates.crt
```

Therefore in the config file example I used type to specify what it is. Often there is either ca path setting or ca file setting.

dachaac:
> What is your opinion about the deprecated way to set x509 in `leshan-server-demo` ?
> 
> ```
>                                 =======================[ X509 deprecated way]===========================
>                                 | By default Leshan demo uses an embedded self-signed certificate and  |
>                                 | trusts any client certificates allowing to use RPK or X509           |
>                                 | at client side.                                                      |
>                                 | If you want to use your own server keys, certificates and truststore,|
>                                 | you can provide a keystore using :                                   |
>                                 |         -ks, -ksp, [-kst], [-ksa], -ksap should be used together     |
>                                 | To get helps about files format and how to generate it, see :        |
>                                 | See https://github.com/eclipse/leshan/wiki/Credential-files-format   |
>                                 ------------------------------------------------------------------------
>  -ks,--keystore <arg>           Set the key store file.
>                                 If set, X.509 mode is enabled, otherwise built-in RPK credentials are used.
>  -ksp,--storepass <arg>         Set the key store password.
>  -kst,--storetype <arg>         Set the key store type.
>                                 Default: pkcs12.
>  -ksa,--alias <arg>             Set the key store alias to use for server credentials.
>                                 Default: leshan.
>                                 All other alias referencing a certificate will be trusted.
>  -ksap,--keypass <arg>          Set the key store alias password to use.
> ```
> 
> Should we remove it ? (in another PR)

Java KeyStore (or rather .p12) is still a good container to store private keys and associated certificates. This is how many Java libraries and software seem to prefer to use them.

I believe I have somewhere a changed that used Californium's credential loader to load private key and certificates from store.

Perhaps it was this method:
https://github.com/eclipse/californium/blob/2.4.1/element-connector/src/main/java/org/eclipse/californium/elements/util/SslContextUtil.java#L253

One could even take extra mile in here and support pkcs11 security tokens in order to protect private keys from exposure. But I would leave that for future. There seems to be some kind of boom of "inexpensive" security tokens.

We have following trust stores and key stores:

- LWM2M Bootstrap server
  - LWM2M TLS Server Certificate
  - Trust store to validate LWM2M Clients (Mfgs CA's + (EST) Service CA)
  - CoAP EST proxy -> EST server TLS Client Certificate -- this needs to be in KeyStore form. (needed to re-construct in-memory keystore for this)
  - EST Server Trust Store (to verify EST server TLS server certificate validity)

- LWM2M Server
  - Trust store to validate LWM2M Clients ((EST) Service CA -- no need to have Mfgs CA's here)
  - LWM2M TLS Server Certificate

- LWM2M Client
  - Initial Trust Store for connecting to LWM2M Bootstrap Server (or to LWM2M Server if no bootstrap)
  - Initial TLS Client Certificate (signed by Mfg CA) -- used for Initial Bootstrap Server connection -- or fallback if others doesn't work
  - Server Public Key (X.509 certificate) and certificate usage to validate LWM2M Server
  - In EST mode Service Specific TLS Client Certificate (received from CoAP EST proxy) -- use for connecting to LWM2M Server, LWM2M Bootstrap server (when server fails) or when going to LWM2M Bootstrap server to CoAP-EST server to renew certificate

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
dachaac: so it seems -> removing.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
dachaac: Will add -- it was kinda following the logic in CA file/path handling just give warnings and continue.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
dachaac: In my opinion quite common and good approach is to give clear error message if something is wrong in command line arguments and then exit with error (System.exit(<non-zero>).

What I tried to follow here was what was code in there.

I can change it how you want it to be. I would like to keep them in sync in all three demos.

Perhaps something like:
```java
                System.err.println("Failed to load trust store: file or directory does not exist: " + input.toString());
                // or LOG.error() ?
                //?? formatter.printHelp(USAGE, options);
                return;
```

Shall I change also:
```java
                System.err.println("Failed to load trust store: " + e.getMessage());
                // or LOG.error() ?
                //?? formatter.printHelp(USAGE, options);
                return;
```
(and drop stack printing)

LOG facility has throwable argument support too. So it might provide common formatting for exceptions if this is something you wish to have.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
dachaac: Of about PEM certificate bundles -- there is some helpers in certificate chain support code (SecurityUtil.certificateChain) that could be adapted to easier file path variant to load PEM (or DER) files so we could change to that kind of functionality instead of using SslContextUtil.loadTrustedCertificates() which is a bit clumsy.

What one would change for this use case is also support loading of other kind of certificates than EC.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
dachaac: We kind are -- but is is just a bit delayed.

We do configure newBuilder.setCertificateVerifier() latter which needs to have customized trust store.

And in setCertificateVerifier there is enforcement that config.trustStore must be null.

Check certificate usage code changes on how trust store gets customized.

dachaac:
> As we introduce new formats (jks, bks, p12, pem), should we add more documentation in https://github.com/eclipse/leshan/wiki/Credential-files-format ?

Yes we should updated that wiki page.

For leshan I had to do swiss knife style certificate setup scripts -- so it would be very nice to simplify it a bit in the end ;)
https://github.com/dachaac/device-certificate-scripts/blob/leshan/leshan-setup-certs.sh

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
sbernard31: > I can change it how you want it to be. I would like to keep them in sync in all three demos.

I agree.

> I can change it how you want it to be.

Thx :pray:

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
sbernard31: Ok let's start with just PKCS#12 in documentation.

sbernard31:
About X509 deprecated way params, my question was mainly : "should we remove it now ?"
I'm not sure at all, but currently my feeling is multiplying the way to provide same data is maybe not a good idea, in particular in demos.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClientBuilder.java
sbernard31: I missed that once truststore is set in dtlsConfig we can not set it to null anymore, so it's imposible to "clean" the dtlsConfig afterwards.

I create an issue at californium to discuss about this : https://github.com/eclipse/californium/issues/1427

I suppose at short term we can go for your solution.

dachaac:
> About X509 deprecated way params, my question was mainly : "should we remove it now ?"
> I'm not sure at all, but currently my feeling is multiplying the way to provide same data is maybe not a good idea, in particular in demos.

Key store is different beast from trust store so I would not do that in this PR.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
dachaac: added

dachaac:
Rebased and updated the PR.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
dachaac: Please check if the current messages are OK.

sbernard31:
Merged thx @dachaac  :pray: !

PR #908, opened 2020-10-12
Update BindingMode to be LWM2M 1.1 compliant and support Q parameter. 

sbernard31:
There is several changes in LWM2M 1.1 about binding mode : 
- new mode was added (T,N)
- QueueMode is not part of binding mode anymore but has it own parameter.

This PR aims to support new format, trying to keep compatible with LWM2M 1.0.

[Here](https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/490#issuecomment-706995813) some discussion about LWM2M changes about Binding Mode.

(I will maybe add  test about that to this PR)

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/request/RegisterRequest.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/request/RegisterRequest.java
@@ -17,10 +17,12 @@ package org.eclipse.leshan.core.request;
 
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.Map;
 
 import org.eclipse.leshan.core.Link;
+import org.eclipse.leshan.core.LwM2m.Version;
 import org.eclipse.leshan.core.request.exception.InvalidRequestException;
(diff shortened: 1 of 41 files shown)


leshan-core/src/main/java/org/eclipse/leshan/core/request/RegisterRequest.java
```diff
@@ -34,6 +35,7 @@
     private final Long lifetime;
     private final String lwVersion;
```
sbernard31, line 36: I should maybe change the type of this attribute to `Version` (same in `Registration` object) :thinking:

sbernard31:
I read more about binding mode. I think this PR is not OK I need to change it.

sbernard31:
This time I think it's OK. 
More changes about binding in LWM2M 1.1 than what I thought :sweat_smile:.

See : 
- https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/505
- https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/490

sbernard31:
Need more work about bootstrap part ... :confounded:

PR #928, opened 2020-11-12
Add `shutdownTrigger` for registration engine

moznion:
Hello,

If the procedure inside of `ClientInitiatedBootstrapTask`, `UpdateRegistrationTask`, and/or `RegistrationTask` raises an unexpected exception, that exception caught by a `catch` clause and it logs that, but that task never works after caught and unfortunately the process still alive.
For example, if `registerWithRetry()` that is in `RegistrationTask#run()` raises a `RuntimeException`, that task never scheduled anymore.
This behavior makes a little bit hard to implement a robust LwM2M client. I think it would be better to exit when it falls an unexpected situation, as like as "let it crash".
So this patch introduces the `shutdownTrigger` mechanism to shut down the client when an unexpected exception occurred. `shutdownTrigger` is an implementation of `Runnable`, so the user can hand over the user-defined shutdown procedure to the registration engine.

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
@@ -58,6 +58,11 @@ import org.slf4j.LoggerFactory;
 public class LeshanClient implements LwM2mClient {
 
     private static final Logger LOG = LoggerFactory.getLogger(LeshanClient.class);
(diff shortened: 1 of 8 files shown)


moznion:
I'm sorry, I had to make an issue at first instead of pull-request. I opened issue #933, so I close this pull-request for now.

sbernard31:
No problem, I just have time to look at this when you open the PR, after I must confess that I forget you :grimacing:

Issue #933, opened 2020-11-19
Better handle unexpected error in DefaultRegistrationEngine. 


moznion:
Hello,

If the procedure inside of `ClientInitiatedBootstrapTask`, `UpdateRegistrationTask`, and/or `RegistrationTask` raises an unexpected exception, that exception caught by a `catch` clause and it logs that, but that task never works after caught and unfortunately the process still alive.

For example, if `registerWithRetry()` that is in `RegistrationTask#run()` raises a `RuntimeException`, that task never scheduled anymore (i.e. that process does nothing).

This behavior makes it a little bit hard to implement a robust LwM2M client. I think it would be better to provide a way to exit when it falls into an unexpected situation.

What do you think about this?

sbernard31:
Hi @moznion, thx for reporting this :pray: and sorry for the delay. 

I think you raise a good point. I looked at you PR #928 and you propose to add a callback.

Before to go in this way I would discuss a bit about this.

### 1. When this happens do we really want to stop the device or we want to reschedule a task ?
I suppose that if something unexpected happens this is maybe better to stop the client.

### 2. Is there something else to do than stopping the client ? in other word do we really need a callback ?
> This behavior makes it a little bit hard to implement a robust LwM2M client. I think it would be better to provide a way to exit when it falls into an unexpected situation.

Should we directly stop the client instead of let user to do it ? (I mean stop the client in leshan library not only in leshan-client-demo)

### 3. if we need a callback maybe we can reuse `LwM2mClientObserver` ?
instead of creating a new one ?

### 4. Looking at PR #928, it seems you need to execute code in `LwM2mObjectEnabler` too ?
Maybe we can rather add a new method in this interface to handle this ?

moznion:
Thank you for your response, @sbernard31!

> 1. When this happens do we really want to stop the device or we want to reschedule a task ?

Yes, I agree with you. If something unexpected occurred, I suppose it would be better to stop the client than prolonging the lives of the process/task.

> 2. Is there something else to do than stopping the client ? in other word do we really need a callback ?

This is a good point. I focused on keeping the backward compatibility for the existing client implementation, so I intended to leave users to control the behavior to handle something in an unexpected situation, as a callback.
Anyway, I think "leshan-clinet" should stop itself if it falls into an unhealthy status, originally.

> 3. if we need a callback maybe we can reuse `LwM2mClientObserver` ?

[...]

PR #941, opened 2020-11-30
Add callback mechanism into LwM2mClientObserver for when the unexpected error occurred

moznion:
Ref #933 

## Done

- Add `onUnexpectedErrorOccurred()` method to `LwM2mClientObserver` interface
- Call `LwM2mClientObserver#onUnexpectedErrorOccurred()` on `DefaultRegistrationEngine` when it raises `RuntimeException` at task loop.
- Support this error handling system on the client demo project

### DIFF
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/observer/LwM2mClientObserver.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/observer/LwM2mClientObserver.java
@@ -69,4 +69,8 @@ public interface LwM2mClientObserver {
             String errorMessage, Exception cause);
 
     void onDeregistrationTimeout(ServerIdentity server, DeregisterRequest request);
+
+    // ============== Unexpected Error Handling =================
+
+    void onUnexpectedErrorOccurred(Throwable unexpectedError);
 }
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/observer/LwM2mClientObserver.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/observer/LwM2mClientObserver.java
@@ -72,5 +72,5 @@ public interface LwM2mClientObserver {
 
     // ============== Unexpected Error Handling =================
 
-    void onUnexpectedErrorOccurred(Throwable unexpectedError);
+    void onUnexpectedError(Throwable unexpectedError);
 }
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
@@ -895,7 +895,7 @@ public class LeshanClientDemo {
         }
 
         @Override
-        public void onUnexpectedErrorOccurred(Throwable unexpectedError) {
+        public void onUnexpectedError(Throwable unexpectedError) {
             LOG.error("unexpected error occurred", unexpectedError);
             client.destroy(true);
             myDevice.destroy();
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
@@ -624,11 +624,9 @@ public class LeshanClientDemo {
                 initializer.setInstancesForObject(SERVER, new Server(123, lifetime));
             }
         }
-        final MyDevice myDevice = new MyDevice();
-        final RandomTemperatureSensor randomTemperatureSensor = new RandomTemperatureSensor();
-        initializer.setInstancesForObject(DEVICE, myDevice);
+        initializer.setInstancesForObject(DEVICE, new MyDevice());
         initializer.setInstancesForObject(LOCATION, locationInstance);
-        initializer.setInstancesForObject(OBJECT_ID_TEMPERATURE_SENSOR, randomTemperatureSensor);
+        initializer.setInstancesForObject(OBJECT_ID_TEMPERATURE_SENSOR, new RandomTemperatureSensor());
         List<LwM2mObjectEnabler> enablers = initializer.createAll();
 
         // Create CoAP Config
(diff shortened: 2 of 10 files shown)


leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -745,6 +748,9 @@ public void handshakeFailed(Handshaker handshaker, Throwable error) {
         builder.setBootstrapAdditionalAttributes(bsAdditionalAttributes);
         final LeshanClient client = builder.build();
 
+        client.addObserver(
+            new ClientShutdownOnUnexpectedErrorObserver(client, myDevice, randomTemperatureSensor));
```
sbernard31, line 752: This should be done on `LeshanClient` not `LeshanClientDemo`  or I missed something ?
(https://github.com/eclipse/leshan/issues/933#issuecomment-730848056)

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -872,4 +878,28 @@ public void run() {
...
+            this.myDevice = myDevice;
+            this.randomTemperatureSensor = randomTemperatureSensor;
+        }
+
+        @Override
+        public void onUnexpectedErrorOccurred(Throwable unexpectedError) {
+            LOG.error("unexpected error occurred", unexpectedError);
+            client.destroy(true);
+            myDevice.destroy();
+            randomTemperatureSensor.destroy();
```
sbernard31, line 902: I think we need to do the same `Lwm2mInstanceEnabler` should also implement (startable, stopable, destroyable)
And Object enabler should do pretty much the same as "LwM2mObjectTree" by interating over instance. (see https://github.com/eclipse/leshan/issues/933#issuecomment-732775930)

leshan-client-core/src/main/java/org/eclipse/leshan/client/observer/LwM2mClientObserver.java
```diff
@@ -69,4 +69,8 @@ void onDeregistrationFailure(ServerIdentity server, DeregisterRequest request, R
             String errorMessage, Exception cause);
 
     void onDeregistrationTimeout(ServerIdentity server, DeregisterRequest request);
+
+    // ============== Unexpected Error Handling =================
+
+    void onUnexpectedErrorOccurred(Throwable unexpectedError);
```
sbernard31, line 75: maybe `onUnexpectedError` is enough ?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
moznion: Yes, that's a point. I think it would be better to implement this shutdown observer mechanism on `LeshanClient`, but it breaks the backward compatibility of the client, so I delegated this behavior to the client of the client.

What do you think about this?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
moznion: Sounds good. I'll do this 👍

leshan-client-core/src/main/java/org/eclipse/leshan/client/observer/LwM2mClientObserver.java
moznion: Thank you for the advice, I've not had confidence in this honestly :)

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
sbernard31: For `master`, we don't care about backward compatibility question and should just focus on what could be the better behavior. (of course this doesn't mean we can break everything without any reason :sweat_smile: )

For `1.x`, we should try to not break the API and behavior but here we could almost consider this as a bug ... so changing this behavior doesn't disturb me. 

WDYT ?

moznion: Yeah, I'm on the same page.
I also think it would be better to implement this on the LeshanClient.
I was scared to break the backward compatibility way too much, but it seems no problem. I'll do this.

One more thing. Would it useful to support a method to prevent shutting down even if an unexpected error has occurred on the LeshanClient? For example `LeshanClient#doNotDestroyOnUnexpectedError()`.

sbernard31: I would not add it for now. Let's see if someone need it later.

sbernard31: I still don't see what could be the use case. As task in `DefaultRegistrationEngine` are stopped. :thinking:

moznion: yeah, my suggestion was for just in case. It looks there is no reason to support that. Thank you.

moznion: fixed at 7e6d029

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
moznion: fixed at 3f73020

leshan-client-core/src/main/java/org/eclipse/leshan/client/observer/LwM2mClientObserver.java
moznion: fixed at 16c6cb7

sbernard31:
I integrated this in master (commits d0f27053efbc123eb393e12171d194de3a0e8529, a1bacb5bd15630e6099a7475839e7256ae62bca0, 1da2600459bdddded26a762396d5a106aaece21d, 7a013f04b40f60de00866eecd64ce5f89532fcf2)

I rewrite history to get more consistent commits.
I fixed a javadoc formatting issue in `LwM2mInstanceEnabler`.
I change a little bit the last commit by using anonymous class (see 7a013f04b40f60de00866eecd64ce5f89532fcf2), in this commit I also change `LeshanClient.this.destroy(true);` to ` `LeshanClient.this.destroy(false);` (because I was afraid in some condition that we could face a kind of infinite loop)

Thx @moznion for this new contribution :pray: !

moznion:
Many thanks for your support, @sbernard31 !

PR #940, opened 2020-11-26
Add support of  `Startable`, `Stoppable` and `Destroyable` to LwM2mObjectEnabler.

moznion:
Rel: #933 

## Done

- Moves `Destroyable`, `Startable` and `Stoppable` class to `leshan-core` from `leshan-server-core` to use these classes with `LwM2mObjectEnabler`
- Implements `Destroyable`, `Startable` and `Stoppable` on `LwM2mObjectTree`
  - These implemented methods call the associated objectEnablers' method inside
- Calls `LwM2mObjectTree#{stop(), start(), destroy()}` in `LeshanClient#{stop(), start(), destroy()}`

PR #944, opened 2020-12-08
Backport: error handling for unexpected error in DefaultRegistrationEngine

moznion:
This pull-request backports the changes of #940 and #941 to `1.x` version.
See also: #933.

### DIFF
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
@@ -63,6 +63,7 @@ import org.eclipse.leshan.client.californium.LeshanClient;
 import org.eclipse.leshan.client.californium.LeshanClientBuilder;
 import org.eclipse.leshan.client.engine.DefaultRegistrationEngineFactory;
 import org.eclipse.leshan.client.object.Server;
+import org.eclipse.leshan.client.observer.LwM2mClientObserverAdapter;
 import org.eclipse.leshan.client.resource.LwM2mObjectEnabler;
 import org.eclipse.leshan.client.resource.ObjectsInitializer;
 import org.eclipse.leshan.client.resource.listener.ObjectsListenerAdapter;
@@ -691,6 +692,7 @@ public class LeshanClientDemo {
                 LOG.info("Object {} enabled.", object.getId());
             }
         });
+        client.addObserver(new ShutdownOnUnexpectedErrorObserver(client));
 
         // Display client public key to easily add it in demo servers.
         if (clientPublicKey != null) {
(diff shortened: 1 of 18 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/Destroyable.java
```diff
@@ -15,6 +15,10 @@
  *******************************************************************************/
 package org.eclipse.leshan.server;
 
+/**
+ * @deprecated please consider to use {@link org.eclipse.leshan.core.Destroyable} instead.
+ */
+@Deprecated
```
sbernard31, line 21: Everywhere we use the old  `Destroyable` interface we should now be able to use the new one.
(I created a new PR with this modification : df5725cc3410e2b771989b98ac6e11a995f0dc53)

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -807,4 +809,18 @@ public void run() {
             }
         }
     }
+
+    private static class ShutdownOnUnexpectedErrorObserver extends LwM2mClientObserverAdapter {
```
sbernard31, line 813: I hope I'm not too boring with this but are you sure we should keep the backward compatibility behavior ?
As I said in https://github.com/eclipse/leshan/pull/941#discussion_r534009976, we could maybe consider the current behavior as a bug rather as a behavior which should not be changed ?

WDYT ?

Of course if you insist we can only add this behavior to `LeshanClientDemo` but it's hard to me to find why keeping a kind of process zombie alive should be a good idea.(https://github.com/eclipse/leshan/issues/933#issue-746373116)

sbernard31 (commented):
This is good.
I just detect some issue about backward behavior compatibility about Destroyable, Startable, Stoppable interface and so create a dedicated PR #945. This way you can look at the modif before I integrate it.

I also squashed some of your commits to make it more consistent. 

The remaining question should we shutdown on `LeshanClient` or `LeshanClientDemo` ?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
moznion: Yeah, this has been a point for consideration.
The previous pull-request (#941) was for the new major version 2.0, on the other hand, this pull-request aims to update the current stable version. So I'm a little bit afraid of breaking the backward compatibility...
But, as you mentioned, it is difficult to find out the reason to encourage keeping the living-dead client process.
This is just my opinion, it would be better to set the hook in `LeshanClinet`; but anyway I'd like to follow and respect the decision of this project. Let's do the same thing with #941?

sbernard31: Ok so I cherrypick the commit from master instead of the one in this PR.

sbernard31:
I close this one which is contained in #945.

Issue #957, opened 2021-01-05
add Read-Composite request support


sbernard31:
I'm currently thinking about Read-Composite API. 

See :
 -  [core§6.3.8-Read-Composite-Operation](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#6-3-8-0-638-Read-Composite-Operation)
 - [transport§6.4.4-Device-Management-and-Service-Enablement-Interface](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Transport-V1_1_1-20190617-A.html#6-4-4-0-644-Device-Management-and-Service-Enablement-Interface)

I guess the request should looks like this : 
```java
new ReadCompositeRequest(requestContentFormat,responseContentFormat, String path ...);
// or maybe but currently LwM2mPath is pretty much not exposed API
new ReadCompositeRequest(requestContentFormat,responseContentFormat, LwM2mPath path ...);
```

For the response, this is less easy, I currently have no idea how it should looks like, especially the payload (content) part.  
```java
public class ReadCompositeResponse {

    ResponseCode getCode();
    ???? getContent();
}
```

I suppose we could imagine : 

- `Collection<LwM2mNode> getContent()` pretty straightforward but not sure this is so easy to use API :/
- `Map<String,LwM2mNode>` where we match a response for each request path but this could be an issue [for `"/"` use case](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#Table-745-10-SenML-JSON-payload-in-the-request-to-all-objects) where there is not 1 LwM2mNode but several. 
- `NavigableMap<LwM2mPath,Lwm2mNode>` which should allow a better way to navigate in received data but API will not be so straightforward too
- or maybe we need a new dedicated structure based on NaviableMap which could allow to extract data for a given path ?

sbernard31:
(@yemkay, @madhushreegc you may be interested by this)

sbernard31:
I'm currently exploring the `Map<String,LwM2mNode>` way because it seems to be the less impacting way. 
(for "/" case I will maybe need a `LwM2mRootNode`)

madhushreegc:
[...]

PR #958, opened 2021-01-08
Add Read-Composite operation

sbernard31:
This aims to implement #957.

~(This is work in progress, not in reviewable state or testable state)~

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
@@ -42,6 +42,8 @@ import org.eclipse.leshan.client.observer.LwM2mClientObserverAdapter;
 import org.eclipse.leshan.client.observer.LwM2mClientObserverDispatcher;
 import org.eclipse.leshan.client.resource.LwM2mObjectEnabler;
 import org.eclipse.leshan.client.resource.LwM2mObjectTree;
+import org.eclipse.leshan.client.resource.LwM2mRootEnabler;
+import org.eclipse.leshan.client.resource.RootEnabler;
 import org.eclipse.leshan.client.resource.listener.ObjectListener;
 import org.eclipse.leshan.client.resource.listener.ObjectsListenerAdapter;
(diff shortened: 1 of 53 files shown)


sbernard31:
I finished to write a first "working" version. This was mainly to detect all the impact on current code base.
This should be enough to play with it but this is clearly not in a clean state.

Now I need to polish this (implement some missing details, clean the code, add more validation, add javadocs, some renames ...)
ReadComposite using "/" is not implemented now and will be probably not in this PR.

The API looks like this : 
```java
// read device model number
ReadCompositeResponse response = server.send(registration,
        new ReadCompositeRequest(ContentFormat.CBOR, ContentFormat.CBOR, "/3/0/0", "/1"));

if (response.isSuccess()) {
    List<LwM2mNode> content = response.getContent();
    content.get(0); // response for "/3/0/0"
    content.get(1); // response for "/1"
    ...
}
```

sbernard31:
@madhushreegc @eliasweingaertner I think this is in in a reviewable and testable state.
(If you want to look at the code I advice to check commit by commit this will be more understandable)

My next step for another PR will be to look at the WriteComposite to check if the ReadComposite design matches correctly with it.

sbernard31:
Unfortunately, I need to update the code by mainly switching all the `List<LwM2mPath>` + `List<LwM2mNode>` by `Map<LwM2mPath,  LwM2mNode>`.

This was mainly because to decode nodes from WriteComposite. In this case the list of `List<LwM2mPath>` is unknown and so the result must return Node+Path, I'm not sure I'm so clear ... (see first commit of  #959 for more details)

This impacts the `ReadCompositeResponse` API and so now it looks like : 
```java
// read device model number
ReadCompositeResponse response = server.send(registration,
        new ReadCompositeRequest(ContentFormat.CBOR, ContentFormat.CBOR, "/3/0/0", "/1"));

if (response.isSuccess()) {
    LwM2mResourceInstance resource = (LwM2mResourceInstance) response.getContent("/3/0/0");
    LwM2mObject object = (LwM2mObject) response.getContent("/1");
    ...
}
```

sbernard31:

PR #959, opened 2021-01-20
Add Write Composite Operation

sbernard31:
This aims to add Write Composite support.

Atomicity at client side is not supported.

API at server side looks like : 
```java
Map<String, LwM2mNode> nodes = new HashMap<>();
nodes.put("/3/0/14", LwM2mSingleResource.newStringResource(14, "+02"));
nodes.put("/1/0/2", LwM2mSingleResource.newIntegerResource(2, 100));

WriteCompositeResponse response = server.send(helper.getCurrentRegistration(),
        new WriteCompositeRequest(ContentFormat.SENML_COBOR, nodes));
```

This is not so good because resource Id is duplicate but for now I didn't find a better idea.

(This is a work in progress)

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/RootResource.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/RootResource.java
@@ -19,6 +19,7 @@ package org.eclipse.leshan.client.californium;
 import static org.eclipse.leshan.core.californium.ResponseCodeUtil.toCoapResponseCode;
 
 import java.util.List;
+import java.util.Map;
 
 import org.eclipse.californium.core.CoapResource;
 import org.eclipse.californium.core.CoapServer;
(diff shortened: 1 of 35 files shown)


sbernard31:
This is probably not perfect but I think it's now in reviewable/testable state.

The `WriteCompositeRequest` API at server side is still not so good. Ideas are welcomed.

At client side : 
- There isn't real transaction support for atomicity. Maybe we need a kind of rollback/commit support ?
- We probably also need a new kind of synchronization, locking/synchronized at Object level could be not enough.

sbernard31:
I changed the `WriteCompositeRequest` API to avoid this double Id issue looks like : 
```java
Map<String, Object> nodes = new HashMap<>();
nodes.put("/3/0/14", "+02");
nodes.put("/1/0/2", 100);

WriteCompositeResponse response = server.send(helper.getCurrentRegistration(),
        new WriteCompositeRequest(ContentFormat.SENML_COBOR, nodes));
```

sbernard31:

PR #971, opened 2021-02-10
Add Send Operation

sbernard31:
This aims to add support of **Send Operation** to Leshan.

See LWM2M specification for more details : 
- [core§6.4.6-Send-Operation](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#6-4-6-0-646-Send-Operation)
 - [transport§6.4.5-Infomation-Reporting-Interface](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Transport-V1_1_1-20190617-A.html#6-4-5-0-645-Information-Reporting-Interface)
 
 The API look like this :

**Client Side:**
 ```java
// send data
SendResponse r = client.sendData (server,
                                   ContentFormat.SENML_JSON, 
                                   Arrays.asList("/3/0/1", "/3/0/2"),
                                   timeoutInMs);
 ```
 
 **Server Side:**
 ```java
// listen for new data
server.getSendService().addListener(new SendListener() {
   @Override
   public void dataReceived(Registration registration, Map<LwM2mPath, LwM2mNode> data, SendRequest request) {
       System.out.println(data);
   }
});
 ```
 
 List of concerns or potential improvements : 
 - at client side, add more low level function allowing to send `SendRequest` directly.
 - at server side, current Map<LwM2mPath, LwM2mNode> data is not so good. LwM2mPath is largely used internally but generally not so exposed. (we have same problem with WriteComposite)
 - at server side, if an error occurred when handling `SendRequest`, this is not reported to application layer, it should maybe be reported in `SendListener`.
 - at server side, in addition of listener maybe we should add a kind of DataCollector which could be able to report the data handling state. This could be useful to let device know when data was not collected/handled correctly.

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java
@@ -48,6 +48,7 @@ import org.eclipse.leshan.client.resource.listener.ObjectListener;
 import org.eclipse.leshan.client.resource.listener.ObjectsListenerAdapter;
 import org.eclipse.leshan.client.servers.ServerIdentity;
 import org.eclipse.leshan.core.californium.EndpointFactory;
+import org.eclipse.leshan.core.model.LwM2mModel;
 import org.eclipse.leshan.core.node.codec.LwM2mNodeDecoder;
 import org.eclipse.leshan.core.node.codec.LwM2mNodeEncoder;
 import org.eclipse.leshan.core.util.Validate;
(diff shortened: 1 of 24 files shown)


sbernard31:
I changed the Send API at server side like this : 
```java
// listen for new data
server.getSendService().addListener(new SendListener() {
  @Override
  public void dataReceived(Registration registration, Map<String, LwM2mNode> data, SendRequest request) {
      System.out.println(data);
  }
});
```
(data attribute changed from `Map<LwM2mPath, LwM2mNode>` to  `Map<String, LwM2mNode> data`)

sbernard31:

zakharenkodmytro:
Hi @sbernard31 ,
I have a question for you, our team need to build coap server , can we theoretically reuse Leshan server for it , so our dmserver will serve both Lwm2m and Coap clients?

sbernard31:
Hi I can not see how this is related to this PR ? 
Please create a new issue for this kind of question.
(after you created the issue I will delete your comment above :point_up:  and this one)

zakharenkodmytro:
Hi 
Sorry for posting it here
https://github.com/eclipse/leshan/issues/990 I created issue for this
Please delete above comment

PR #983, opened 2021-03-15
Certificate usage config

dachaac:
Adds support to configure certificate usage from server side during bootstrapping.

Also adds support to leshan-client-demo for using custom certificate usage for initial connection.

### DIFF
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
@@ -62,6 +62,7 @@ import org.eclipse.leshan.client.object.Server;
 import org.eclipse.leshan.client.resource.LwM2mObjectEnabler;
 import org.eclipse.leshan.client.resource.ObjectsInitializer;
 import org.eclipse.leshan.client.resource.listener.ObjectsListenerAdapter;
+import org.eclipse.leshan.core.CertificateUsage;
 import org.eclipse.leshan.core.LwM2m;
 import org.eclipse.leshan.core.californium.DefaultEndpointFactory;
 import org.eclipse.leshan.core.model.LwM2mModel;
@@ -225,6 +226,9 @@ public class LeshanClientDemo {
         options.addOption("truststore", true,
                 "The path to a root certificate file to trust or a folder containing all the trusted certificates in X509v3 format (DER encoding) or trust store URI."
                         + trustStoreChapter);
+
+        options.addOption("cu", "certificate-usage", true,
+                "Certificate Usage (as integer) for security object.\n Default: domain issued certificate (3).");
 
         HelpFormatter formatter = new HelpFormatter();
         formatter.setWidth(90);
@@ -504,6 +508,11 @@ public class LeshanClientDemo {
             }
         }
 
+        CertificateUsage certificateUsage = CertificateUsage.DOMAIN_ISSUER_CERTIFICATE;
+        if (cl.hasOption("cu")) {
+            certificateUsage = CertificateUsage.fromCode(Integer.parseInt(cl.getOptionValue("cu")));
+        }
+
         // get local address
         String localAddress = null;
         int localPort = 0;
@@ -552,8 +561,8 @@ public class LeshanClientDemo {
             createAndStartClient(endpoint, localAddress, localPort, cl.hasOption("b"), additionalAttributes,
                     bsAdditionalAttributes, lifetime, communicationPeriod, serverURI, pskIdentity, pskKey,
                     clientPrivateKey, clientPublicKey, serverPublicKey, clientCertificate, serverCertificate,
-                    trustStore, latitude, longitude, scaleFactor, cl.hasOption("ocf"), cl.hasOption("oc"),
-                    cl.hasOption("r"), cl.hasOption("f"), modelsFolderPath, ciphers);
+                    trustStore, certificateUsage, latitude, longitude, scaleFactor, cl.hasOption("ocf"),
+                    cl.hasOption("oc"), cl.hasOption("r"), cl.hasOption("f"), modelsFolderPath, ciphers);
         } catch (Exception e) {
             System.err.println("Unable to create and start client ...");
             e.printStackTrace();
@@ -566,9 +575,9 @@ public class LeshanClientDemo {
             Integer communicationPeriod, String serverURI, byte[] pskIdentity, byte[] pskKey,
             PrivateKey clientPrivateKey, PublicKey clientPublicKey, PublicKey serverPublicKey,
             X509Certificate clientCertificate, X509Certificate serverCertificate, List<Certificate> trustStore,
-            Float latitude, Float longitude, float scaleFactor, boolean supportOldFormat,
-            boolean supportDeprecatedCiphers, boolean reconnectOnUpdate, boolean forceFullhandshake,
-            String modelsFolderPath, List<CipherSuite> ciphers) throws Exception {
+            CertificateUsage certificateUsage, Float latitude, Float longitude, float scaleFactor,
+            boolean supportOldFormat, boolean supportDeprecatedCiphers, boolean reconnectOnUpdate,
+            boolean forceFullhandshake, String modelsFolderPath, List<CipherSuite> ciphers) throws Exception {
 
         locationInstance = new MyLocation(latitude, longitude, scaleFactor);
 
@@ -592,7 +601,7 @@ public class LeshanClientDemo {
                 initializer.setClassForObject(SERVER, Server.class);
             } else if (clientCertificate != null) {
                 initializer.setInstancesForObject(SECURITY, x509Bootstrap(serverURI, clientCertificate.getEncoded(),
-                        clientPrivateKey.getEncoded(), serverCertificate.getEncoded()));
+                        clientPrivateKey.getEncoded(), serverCertificate.getEncoded(), certificateUsage.code));
                 initializer.setClassForObject(SERVER, Server.class);
             } else {
                 initializer.setInstancesForObject(SECURITY, noSecBootstap(serverURI));
@@ -608,7 +617,7 @@ public class LeshanClientDemo {
                 initializer.setInstancesForObject(SERVER, new Server(123, lifetime));
             } else if (clientCertificate != null) {
                 initializer.setInstancesForObject(SECURITY, x509(serverURI, 123, clientCertificate.getEncoded(),
-                        clientPrivateKey.getEncoded(), serverCertificate.getEncoded()));
+                        clientPrivateKey.getEncoded(), serverCertificate.getEncoded(), certificateUsage.code));
                 initializer.setInstancesForObject(SERVER, new Server(123, lifetime));
             } else {
                 initializer.setInstancesForObject(SECURITY, noSec(serverURI, 123));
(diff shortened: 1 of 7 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityInfo.java
```diff
@@ -50,14 +51,18 @@
     // X.509
     private final boolean useX509Cert;
 
+    // certificate usage
+    private final CertificateUsage certificateUsage;
+
     private SecurityInfo(String endpoint, String identity, byte[] preSharedKey, PublicKey rawPublicKey,
-            boolean useX509Cert) {
+            boolean useX509Cert, CertificateUsage certificateUsage) {
```
sbernard31, line 58: I don't get what we need `Certificate Usage` in `SecurityInfo`.

SecurityInfo is information about to how device should connect to the server. (This is different than bootstrap configuration which define, the value to write on the device at bootstrap time)

E.g. client "C1" should connect using RPK with the  public key P1.

Here we just need to know it use x509 but I'm not sure server need to know which `CertificateUsage` is used.

dachaac: I am observing the code from my full set of changes so it should capture all usage so far.

In client side we have `org.eclipse.leshan.client.servers.ServerInfo#certificateUsage`. Which should already be used in client side for doing client side work.

Only call for using `org.eclipse.leshan.server.security.SecurityInfo#getCertificateUsage()` seems to be in `org.eclipse.leshan.server.redis.serialization.SecurityInfoSerDes#serialize()`.

So I have probably only captured that usage from Redis part -- but is the Redis code actually something that should be changed instead to something else? One option would use same JSON serialization that is used other servers. Probably the redis backend doesn't actually work in here as there is only one getter.

I am actually having a slight problem with Redis to understand the purpose of Redis in here -- it is often used for fast cache -- eg. simple key value pair thing. But now it seems that it also does configuration storing and not as secondary cache but as primarily storage?

Because of above I was thinking to use Redis as shared storage between bootstrap and server so that also bootsrapping configuration would be stored there (eg. use as device inventory). Eg. I was looking at option to handle certificate based device auto registration in my EST support demo code.

sbernard31: I don't get your point at all :confused: . 
I read my comment above again and see there is a typo and so maybe you misunderstood my question, please read it again.

My current feeling is that we don't need to add `Certificate Usage` to  `SecurityInfo`, we don't need it to identify a Client at server side. 

> I am actually having a slight problem with Redis to understand the purpose of Redis in here -- it is often used for fast cache -- eg. simple key value pair thing. But now it seems that it also does configuration storing and not as secondary cache but as primarily storage?

We use it as a store not a cache. (and this is an optional module which allow us to test persistent store use case)

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -608,7 +617,7 @@ public static void createAndStartClient(String endpoint, String localAddress, in
                 initializer.setInstancesForObject(SERVER, new Server(123, lifetime));
             } else if (clientCertificate != null) {
                 initializer.setInstancesForObject(SECURITY, x509(serverURI, 123, clientCertificate.getEncoded(),
-                        clientPrivateKey.getEncoded(), serverCertificate.getEncoded()));
+                        clientPrivateKey.getEncoded(), serverCertificate.getEncoded(), certificateUsage.code));
```
sbernard31, line 620: Should we add this for Bootstrap case too ? 
Some line above : 
```java
 else if (clientCertificate != null) {
                initializer.setInstancesForObject(SECURITY, x509Bootstrap(serverURI, clientCertificate.getEncoded(),
                        clientPrivateKey.getEncoded(), serverCertificate.getEncoded()));
                initializer.setClassForObject(SERVER, Server.class);
            } 
```

dachaac: Sure -- was wondering why it wasn't there--- actually it was ;) but not in this code set...

dachaac:
I was doing some "real" setup test so that there were own servers for bootstrap and lwm2m server.

Noticed that with more practical setup I also needed this commit:
https://github.com/dachaac/leshan/commit/ecfb813ec2e44f23c02eed5c88d2fa5b8d8dd467

Shall I include it here?

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
dachaac: Updated the PR with it.

sbernard31:
I don't know if this is my setup or if I missed something but in the bootstrap UI I was expected the `certificate usage field` only for x509 ?

sbernard31:
About the client help it seems there is an issue, it looks like this : 
```
                                 ================================[X509]===================
                                 ===============
                                 | By default Leshan demo use non secure connection.
                                 |
                                 | To use X509, -ccert -cprik -scert options should be
                                 used together.   |
                                 | To get helps about files format and how to generate it,
                                 see :        |
                                 | See
                                 https://github.com/eclipse/leshan/wiki/Credential-files-f
                                 ormat   |
                                 ---------------------------------------------------------
                                 ---------------
 -ccert <arg>                    The path to your client certificate file.
                                 The certificate Common Name (CN) should generaly be equal
                                 to the client endpoint name (see -n option).
                                 The certificate should be in X509v3 format (DER
                                 encoding).

```
instead of : 
```
                ================================[X509]==================================
                | By default Leshan demo use non secure connection.                    |
                | To use X509, -ccert -cprik -scert options should be used together.   |
 -ccert <arg>   The path to your client certificate file.
                The certificate Common Name (CN) should generaly be equal to the client
                endpoint name (see -n option).
```
I think this is a regression since commit (9bc7a3b0eaba8e8b37e59c5f2b694fdeea9c94e0) because currently there is some issue with help prompt when argument name is too long. We should use a shorter name waiting we use another CLI framework (#569) 

I will fix the `truststore` issue and let you fix the `certificate-usage` one in this PR.

sbernard31:
> Noticed that with more practical setup I also needed this commit:
> dachaac@0421e86
> 
> Shall I include it here?

Let discuss about this in another PR because I'm not sure we need a special case for X509.

sbernard31:
> I will fix the truststore issue in master and let you fix the certificate-usage one in this PR.

If you're OK with that I will change it from `--truststore` to `-ts`.

sbernard31:
(a small details, generally I try to keep commit title/summary to 72 characters max, if you need to add more information just add a blank line then a more complete description without size limitation)

sbernard31:
> If you're OK with that I will change it from --truststore to -ts.

I put this change on master (8ce5498c6f542d76dd4b4a18bea4e412186f24f8)

sbernard31:
I will start to work on the little changes I reported above.
If you already started or you want to do it on your own. Stop me :wink:  !

sbernard31:
I started to adapt your code at #991

sbernard31:
I'm back on Monday.

dachaac:
Sorry it took a bit longer than anticipated to get back on this as I needed to debug interoperability issues between commercial client and my Leshan EST support code. I was finally able to connect commercial LWM2M client utilizing EST mode to Leshan LWM2M bootstrap+server running on my Leshan cloud test instance. (with some fixes also required for commercial client to make that more standards compliant -- will report that to them). I can finally relax a bit in less stressful state.

It seems that we fixed the SenML-CBOR opaque problem at the same time :roll_eyes: --- well at least I know how to debug that level too.

> I started to adapt your code at #991

If you want to integrate the code feel free to do so -- but I would like to understand that we do not both work on same things. And if you want to integrate more work in my branches please do tell and I can try to support you as much as possible.

I can fix that webui issue you noticed -- unless you planned to also work on that too.

As I was finally able to do interop between device side and Leshan server side work that I planned on doing is getting into completion. I can try to get Leshan parts integrated but especially the EST commits needs some kind of advice on how to go forward to get into integrated state -- but lets discuss that more on EST issue a bit later.

> > I will fix the truststore issue in master and let you fix the certificate-usage one in this PR.
> 
> If you're OK with that I will change it from `--truststore` to `-ts`.

Personally I prefer more descriptive command line arguments -- but feel free to adjust as you please.

sbernard31:
> Personally I prefer more descriptive command line arguments -- but feel free to adjust as you please.

I find a way to keep the more descriptive way (c478670d1e946fc64ec18)

> I can fix that webui issue you noticed -- unless you planned to also work on that too.

I would like to release a 2.0.0-M3 soon with this PR. So I'm working on it. 

The last part I consider to integrate in 2.0.0-M3 is https://github.com/dachaac/leshan/commit/0421e86
I guess that we can just adapt the code to check only the IP address in all case  (not exclusively for x509 and not exclusively bootstrap server) ?
But I feel that in term of abstraction comparing identity sounds a good idea... maybe the issue is that we don't create the right identity for x509 [here](https://github.com/eclipse/leshan/blob/leshan-2.0.0-M2/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java#L187) Using CN from Certificate in `Server Public Key` (0/?/4) resource does not make sense anymore with Certificate Usage. But I'm not sure we can find the right CN(maybe not the appropriate term)  at this time ? Maybe we should use the domain name or IP address in `LWM2M Server URI` (0/?/0) to create the identity ? 

I'm not sure what is the best approach. @dachaac any thoughts or feedback about this is welcomed.

sbernard31:
About  dachaac@0421e86, 
I confirm this affect not only bootstrap server, I will create a dedicated issue to track it.

sbernard31:
The related issue https://github.com/eclipse/leshan/issues/992

sbernard31:
@dachaac  about  dachaac/leshan@0421e86 let discuss at #992

sbernard31:
@dachaac, I wanted to test a **leshan-client-demo** which connects to a **leshan-bsserver-demo** with custom certificate usage. (e.g. CA constraints)

To do that I need that my **leshan-bsserver-demo** use a certificate chain but there is no CLI option to do that for now.
So I guess we need a way to set this. see #993  to discuss about this :slightly_smiling_face: 

(I tested the *leshan-server-demo** with the deprecated keystore option)

dachaac:
> @dachaac, I wanted to test a **leshan-client-demo** which connects to a **leshan-bsserver-demo** with custom certificate usage. (e.g. CA constraints)
> 
> To do that I need that my **leshan-bsserver-demo** use a certificate chain but there is no CLI option to do that for now.
> So I guess we need a way to set this. see #993 to discuss about this 
> 
> (I tested the _leshan-server-demo_* with the deprecated keystore option)

If you want to play around with things:
https://github.com/dachaac/leshan/wiki/Testing-EST-support

Those certificate scripts setup all kinds of setups and certificates in different formats to make your life easier for hacking.

and if you want to try out auto-registration there is some hack with Jedis in here:
https://github.com/dachaac/leshan/tree/est-support-with-persistence

sbernard31:
@dachaac do you think you will be able to have a look at #995 and #991 quickly ? 
If you can not, I think I will integrate it in master and release the 2.0.0-M3 today and you could review it later and we adapt the code after the release. 
If you can, I'm waiting for your review :-)

sbernard31:
Maybe it's not not so smart to hurry, as I'm unavailable from this evening too next Tuesday, so I will wait until this day to integrate and release.

sbernard31:
@dachaac as #991 is now integrated in master we can close this one, correct ?

sbernard31:
@dachaac ping ?

sbernard31:
I close this PR as most should be implemented in #991.
Do not hesitate to re-open if needed.

Issue #936, opened 2020-11-23
PKIX: CA Constraint after TrustAnchor


sbernard31:
(To better understand this point, you need to first read #935)

We are talking about Certificate Usage 0 "CA constraint" defined in  [RFC6698](https://tools.ietf.org/html/rfc6698#section-2.1.1) (used since LWM2M 1.1)

>   0 -- Certificate usage 0 is used to specify a CA certificate, or
      the public key of such a certificate, that MUST be found in any of
      the PKIX certification paths for the end entity certificate given
      by the server in TLS.  This certificate usage is sometimes
      referred to as "CA constraint" because it limits which CA can be
      used to issue certificates for a given service on a host.  The
      presented certificate MUST pass PKIX certification path
      validation, and a CA certificate that matches the TLSA record MUST
      be included as part of a valid certification path.  Because this
      certificate usage allows both trust anchors and CA certificates,
      the certificate might or might not have the basicConstraints
      extension present.

Generally in this use case the chains looks like this : 
```
Server -> CA (n) --> CA (n-1) --> ... --> CA 1  --> Root CA
```
(My understanding : this is possible to eventually have a chain without CA.)

A typical use case : 
```
Server -> CA 2 --> CA 1 --> Root CA
           ^        ^
           |        |
 CA Constraint     TrustAnchor
```
As  "CA constraint" can also be consider as a TrustAnchor and following #935 we will probably stop the PKIX validation at CA2. (From Server to CA2).

**The question is : What happens if CA Constraint is after another TrustAnchor .** 
E.g : 
```
Server -> CA 2 --> CA 1 --> Root CA
           ^        ^
           |        |
   TrustAnchor    CA Constraint
```

Following same logic we will  stop PKIX Validation to CA2 and we should also check that CA constraint (CA1 in this case) is in the certificate path. In this case CA1 is in the path but not really in the PKIX validation path which sounds not so good ?

sbernard31:
Note that we the current implementation, It seems to me that  that "CA constraint" works pretty much the same as "Trust Anchor Assertion".

This comes from [this code](https://github.com/eclipse/leshan/pull/923/commits/6e96cee3e69ae7c6922946b56c88e1c2a868ff30#diff-22330e28a7d3c02e6f39f6db974b1345eb580153ac0943100226d9dd8164b049R156) where we choose to add the  "CA constraint" as a "trusted certificate"(TrustAnchor)

I suppose this is because of this RFC sentence (which is not so clear to me): 
[...]

PR #991, opened 2021-03-25
Add Certificate usage support to leshan demos

sbernard31:
Modified version of #983

### DIFF
--- a/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java
+++ b/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java
@@ -32,6 +32,7 @@ import org.eclipse.californium.elements.Connector;
 import org.eclipse.californium.elements.auth.RawPublicKeyIdentity;
 import org.eclipse.californium.elements.util.CertPathUtil;
 import org.eclipse.californium.scandium.DTLSConnector;
+import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig.Builder;
 import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
 import org.eclipse.californium.scandium.dtls.pskstore.AdvancedSinglePskStore;
@@ -101,7 +102,8 @@ public class CaliforniumEndpointsManager implements EndpointsManager {
         // Create new endpoint
         Identity serverIdentity;
         if (serverInfo.isSecure()) {
-            Builder newBuilder = new Builder(dtlsConfigbuilder.getIncompleteConfig());
+            DtlsConnectorConfig incompleteConfig = dtlsConfigbuilder.getIncompleteConfig();
+            Builder newBuilder = new Builder(incompleteConfig);
 
             // Support PSK
             if (serverInfo.secureMode == SecurityMode.PSK) {
@@ -191,6 +193,13 @@ public class CaliforniumEndpointsManager implements EndpointsManager {
             } else {
                 throw new RuntimeException("Unable to create connector : unsupported security mode");
             }
+
+            // For bootstrap no need to have DTLS role exchange
+            // and so we can set DTLS Connection as client only by default.
+            if (serverInfo.bootstrap && incompleteConfig.isClientOnly() == null) {
+                newBuilder.setClientOnly();
+            }
+
             currentEndpoint = endpointFactory.createSecuredEndpoint(newBuilder.build(), coapConfig, null);
         } else {
             currentEndpoint = endpointFactory.createUnsecuredEndpoint(localAddress, coapConfig, null);
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Security.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Security.java
@@ -110,6 +110,15 @@ public class Security extends BaseInstanceEnabler {
     }
 
     /**
+     * Returns a new security instance (X509) for a bootstrap server.
+     */
+    public static Security x509Bootstrap(String serverUri, byte[] clientCertificate, byte[] clientPrivateKey,
+            byte[] serverPublicKey, ULong certificateUsage) {
+        return new Security(serverUri, true, SecurityMode.X509.code, clientCertificate.clone(), serverPublicKey.clone(),
+                clientPrivateKey.clone(), null, certificateUsage);
+    }
+
+    /**
      * Returns a new security instance (NoSec) for a device management server.
      */
     public static Security noSec(String serverUri, int shortServerId) {
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
@@ -31,6 +31,7 @@ import java.security.cert.X509Certificate;
 import java.security.interfaces.ECPublicKey;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
@@ -68,6 +69,7 @@ import org.eclipse.leshan.client.object.Server;
 import org.eclipse.leshan.client.resource.LwM2mObjectEnabler;
 import org.eclipse.leshan.client.resource.ObjectsInitializer;
 import org.eclipse.leshan.client.resource.listener.ObjectsListenerAdapter;
+import org.eclipse.leshan.core.CertificateUsage;
 import org.eclipse.leshan.core.LwM2m;
 import org.eclipse.leshan.core.californium.DefaultEndpointFactory;
 import org.eclipse.leshan.core.model.LwM2mModel;
@@ -215,7 +217,7 @@ public class LeshanClientDemo {
         options.addOption("ccert", true,
                 "The path to your client certificate file.\n The certificate Common Name (CN) should generaly be equal to the client endpoint name (see -n option).\nThe certificate should be in X509v3 format (DER encoding).");
         options.addOption("scert", true,
-                "The path to your server certificate file.\n The certificate should be in X509v3 format (DER encoding).");
+                "The path to your server certificate file (see -certificate-usage option).\n The certificate should be in X509v3 format (DER encoding).");
 
         final StringBuilder trustStoreChapter = new StringBuilder();
         trustStoreChapter.append("\n .");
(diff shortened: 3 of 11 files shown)


leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Security.java
```diff
@@ -109,6 +109,15 @@ public static Security x509Bootstrap(String serverUri, byte[] clientCertificate,
                 clientPrivateKey.clone(), null, CertificateUsage.DOMAIN_ISSUER_CERTIFICATE.code);
     }
 
+    /**
+     * Returns a new security instance (X509) for a bootstrap server.
+     */
+    public static Security x509Bootstrap(String serverUri, byte[] clientCertificate, byte[] clientPrivateKey,
+            byte[] serverPublicKey, ULong certificateUsage) {
+        return new Security(serverUri, true, SecurityMode.X509.code, clientCertificate.clone(), serverPublicKey.clone(),
+                clientPrivateKey.clone(), null, certificateUsage);
```
sbernard31, line 118: I changed the 0 to null since (fdc4a9834fbc8691c5c55ab65473f421078b663a)

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -228,12 +230,22 @@ public static void main(final String[] args) {
...
 
+        final StringBuilder certUsageChapter = new StringBuilder();
+        certUsageChapter.append("\n - 0 : CA constraint");
+        certUsageChapter.append("\n - 1 : service certificate constraint");
+        certUsageChapter.append("\n - 2 : trust anchor assertion");
+        certUsageChapter.append("\n - 3 : domain issued certificate (Default value)");
+        certUsageChapter.append("\n (Usage are described at https://tools.ietf.org/html/rfc6698#section-2.1.1)");
+
+        options.addOption("cu", "certificate-usage", true,
+                "Certificate Usage (as integer) defining how to use server certificate." + certUsageChapter);
```
sbernard31, line 247: I added more description here.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -303,6 +315,15 @@ public static void main(final String[] args) {
             }
         }
 
+        // Abort if cu is used without complete X509 config
+        if (cl.hasOption("cu")) {
+            if (!x509config) {
+                System.err.println("cu should be used with ccert and scert for X509 config.");
+                formatter.printHelp(USAGE, options);
+                return;
+            }
```
sbernard31, line 324: And some checks about `-cu` usage option.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -513,6 +534,19 @@ public static void main(final String[] args) {
...
 
+        CertificateUsage certificateUsage = CertificateUsage.DOMAIN_ISSUER_CERTIFICATE;
+        if (cl.hasOption("cu")) {
+            certificateUsage = CertificateUsage.fromCode(Integer.parseInt(cl.getOptionValue("cu")));
+            if (trustStore.isEmpty()) {
+                if (certificateUsage == CertificateUsage.SERVICE_CERTIFICATE_CONSTRAINT) {
+                    System.err.println(
+                            "You need to set a truststore when you are using \"service_certificate_constraint\" usage");
+                    formatter.printHelp(USAGE, options);
+                    return;
```
sbernard31, line 545: If I'm not wrong only SERVICE_CERTIFICATE_CONSTRAINT need a truststore as other mode can use the server certificate as truststore. 

(For CA constraint I think it should no be used as truststore too : see https://github.com/eclipse/leshan/issues/936)

leshan-client-core/src/main/java/org/eclipse/leshan/client/object/Security.java
dachaac: OK

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
dachaac: OK

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
dachaac: OK

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
```diff
@@ -513,6 +534,19 @@ public static void main(final String[] args) {
             }
         }
 
+        CertificateUsage certificateUsage = CertificateUsage.DOMAIN_ISSUER_CERTIFICATE;
+        if (cl.hasOption("cu")) {
+            certificateUsage = CertificateUsage.fromCode(Integer.parseInt(cl.getOptionValue("cu")));
+            if (trustStore.isEmpty()) {
+                if (certificateUsage == CertificateUsage.SERVICE_CERTIFICATE_CONSTRAINT) {
+                    System.err.println(
+                            "You need to set a truststore when you are using \"service_certificate_constraint\" usage");
```
dachaac, line 543: In above help you use `service certificate constraint` and in here `service_certificate_constraint` -- I would recommend using what is in help.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
dachaac: Need to have a bit of fresher mind in order to think this.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
sbernard31: I will change this.

sbernard31: Done.

sbernard31:
I added change about the UI.

I think the missing part for this feature are : 
- #992 + some unit test related to it. 
- #993.
- maybe adding units test about bootstrap server setting certificate usage (but it needs #992)

@dachaac I'm waiting for your feedback about all of those before to go further. :)

sbernard31:
I added certchain support. (#993)

I also limited the certificate usage but only for the config of bootstrap server. (more explanation in commit 2de240b )

sbernard31:
(rebased on master)
(first 2 commits are out of topic)

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java
```diff
@@ -191,6 +193,13 @@ public synchronized ServerIdentity createEndpoint(ServerInfo serverInfo) {
             } else {
                 throw new RuntimeException("Unable to create connector : unsupported security mode");
             }
+
+            // For bootstrap no need to have DTLS role exchange
+            // and so we can set DTLS Connection as client only by default.
+            if (serverInfo.bootstrap && incompleteConfig.isClientOnly() == null) {
+                newBuilder.setClientOnly();
+            }
```
dachaac, line 201: I ended up having auto detection from certificate what to configure:
https://github.com/dachaac/leshan/commit/fea11c87b0629a9e6ecf9a167b6d33c04bd28b2a

and the matching one for server:
https://github.com/dachaac/leshan/commit/613b6fbd7edd1fee8a4517d5f5e2ea0bde32f08b

leshan-integration-tests/credentials/generate_credentials.sh
```diff
@@ -104,7 +104,7 @@ keytool -genkeypair -alias server -keyalg EC -dname 'CN=localhost' \
         -storetype $DEFAULT_STORE_TYPE \
         -ext BasicConstraints=ca:false \
         -ext KeyUsage:critical=digitalSignature,keyAgreement \
-        -ext ExtendedkeyUsage=serverAuth \
+        -ext ExtendedkeyUsage=serverAuth,clientAuth \
```
dachaac, line 107: Just a note -- making multi purpose certificates can hide problems from real deployments where one should not have multi purpose certificates. -- in server it is easier but especially in client you do not have necessary information to verify that device is acting as a server. That is the reason why I made them explicit.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java
sbernard31: This changes is about the library itself and not really related to certificate.

Having LWM2M client and server which could act as both  DTLS client and  DTLS server  is mainly needed when DTLS connection is lost (reboot/crash/connection expired) and server initiated request. (for [more details](https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/410))

In case of bootstrap session we don't need it, because session is short and always initiated by client. So I prefer to set the LWM2M client as DTLS clientOnly for bootstrap.

About "auto detection" this could be a good idea for demo, we add a warning explaining that failover will not work in case we provide a `clientAuth` or `serverAuth` only certificate.

leshan-integration-tests/credentials/generate_credentials.sh
sbernard31: As explained [here](https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/410), I can not see how we can make "none queue mode" work in reliable way without supporting both role at both side. 

So in my mind, "multi purpose" is the default behavior.

But I can imagine that I maybe missed some real life constraint about x509 world. 

So could you elaborate about : 

> "in server it is easier but especially in client you do not have necessary information to verify that device is acting as a server" ? 


About : 
> making multi purpose certificates can hide problems from real deployments where one should not have multi purpose certificates.

A solution could be to add some test with clientOnly  ?

sbernard31:
@dachaac, I will integrate this PR without commit.

> 20d6493 - To not use DTLS ClientOnly mode for integration tests.  
>
>I think this is better to have default behavior in tests and also have
generate_credentials.sh as example to generate both role certificates.

Hoping you're fine with the rest of the PR.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
sbernard31: @dachaac, the PR is merged but please do not hesitate to give me feedback about this here or at #936.

sbernard31:
@dachaac this is integrated in master now but feel free to give feedback anyway.

leshan-integration-tests/credentials/generate_credentials.sh
sbernard31: @dachaac, we can continue the discussion at #997 if you want.

PR #997, opened 2021-04-06
To not use DTLS ClientOnly mode for integration tests.

sbernard31:
I think this is better to have default behavior in tests and also have
generate_credentials.sh as example to generate both role certificates.

(This commits was initially integrated in #991 but it needs more discussion : https://github.com/eclipse/leshan/pull/991#pullrequestreview-626482522)


sbernard31:
Following @dachaac 's comment at https://github.com/eclipse/leshan/pull/991#discussion_r605817941 : 

> Just a note -- making multi purpose certificates can hide problems from real deployments where one should not have multi purpose certificates. -- in server it is easier but especially in client you do not have necessary information to verify that device is acting as a server. That is the reason why I made them explicit.

I guess It make also sense to keep the client only, I close this PR for now.
We will maybe reconsider this  when we will work again on X509.

PR #999, opened 2021-04-08
Add connection ID support to demos

sbernard31:
This is a work in progress.
Not so much tested and missing bootstrap server demo.

CID support is activated by default at server side and should be activated manually at client side (-cid 0)

### DIFF
--- a/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/LeshanBootstrapServerDemo.java
+++ b/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/LeshanBootstrapServerDemo.java
@@ -38,6 +38,7 @@ import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.core.network.config.NetworkConfig.Keys;
 import org.eclipse.californium.elements.util.SslContextUtil;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
+import org.eclipse.californium.scandium.dtls.SingleNodeConnectionIdGenerator;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.servlet.ServletHolder;
 import org.eclipse.jetty.webapp.WebAppContext;
@@ -111,7 +112,13 @@ public class LeshanBootstrapServerDemo {
         options.addOption("m", "modelsfolder", true, "A folder which contains object models in OMA DDF(.xml) format.");
         options.addOption("cfg", "configfile", true,
                 "Set the filename for the configuration.\nDefault: " + JSONFileBootstrapStore.DEFAULT_FILE + ".");
-        options.addOption("oc", "activate support of old/deprecated cipher suites." + RPKChapter);
+        options.addOption("oc", "activate support of old/deprecated cipher suites.");
+        options.addOption("cid", true, "Control usage of DTLS connection ID." //
+                + "\n - 'on' to activate Connection ID support (same as -cid 6)" //
+                + "\n - 'off' to deactivate it" //
+                + "\n - Positive value define the size in byte of CID generated."
+                + "\n - 0 value means we accept to use CID but will not generated one for foreign peer."
+                + "\n (Default: on)" + RPKChapter);
         options.addOption("pubk", true,
                 "The path to your server public key file.\n The public Key should be in SubjectPublicKeyInfo format (DER encoding).");
         options.addOption("prik", true,
@@ -233,6 +240,20 @@ public class LeshanBootstrapServerDemo {
             configFilename = JSONFileBootstrapStore.DEFAULT_FILE;
         }
 
+        // Get CID config
+        String cidOption = cl.getOptionValue("cid");
+        Integer cid = 6;
+        if (cidOption != null) {
+            if ("off".equals(cidOption)) {
+                cid = null;
+            } else if ("on".equals(cidOption)) {
+                // we keep default value
+            } else {
+                cid = Integer.parseInt(cidOption);
+                cid = cid < 0 ? null : cid;
+            }
+        }
+
         // get RPK info
         PublicKey publicKey = null;
         PrivateKey privateKey = null;
@@ -312,7 +333,7 @@ public class LeshanBootstrapServerDemo {
         try {
             createAndStartServer(webAddress, webPort, localAddress, localPort, secureLocalAddress, secureLocalPort,
                     modelsFolderPath, configFilename, cl.hasOption("oc"), publicKey, privateKey, certificate,
-                    trustStore);
+                    trustStore, cid);
         } catch (BindException e) {
             System.err.println(String
                     .format("Web port %s is already in use, you can change it using the 'webport' option.", webPort));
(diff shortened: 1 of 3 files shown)


boaks:
Let's try it out and see, if we get feedback.

sbernard31:
@boaks I'm talking about leshan demo command line option. (I'm not sure If this was clear)

> Let's try it out and see, if we get feedback.

You mean with my last proposal (with on/off) or the current implemented way (without on/off just <positive number>) ?

boaks:
> You mean with my last proposal (with on/off) 

That one.

aleparmi:
on/off/fine_tuning sounds simple and direct

sbernard31:
@aleparmi, I push the modification about on/off/fine_tuning and the missing one for **leshan-bsserver-demo**.

If you test it, let me know and if all is good I integrated in `master`.

aleparmi:
Client-Server should work:

image

Bootstrap-Client I do not have it displayed on Wireshark but it is probably normal, what do you think?

image

Server-Bootstrap was not really able to capture it, but the client registered successfully every time

boaks:
The bootstrap server may run on a different port (5685 ?). 
I'm not sure, what you use as filter for wireshark, may be you provide that filter?

sbernard31:
> Bootstrap-Client I do not have it displayed on Wireshark but it is probably normal, what do you think?

I'm on debian 10 stable and I use the old wireshark version in official debian stable repository.
This version don't support ConnectionID so I don't know the expected behavior. 
I could update to a more recent one but as debian 11 should be released soon and comes with a recent version of whireshark, I prefer to wait. :grimacing: 

> The bootstrap server may run on a different port (5685 ?).

Correct, as the sandbox host both demo lwm2m server and lwm2m bsserver. Both can not used the official port for CoAPs. 
The bootstrap server is using 5784.

aleparmi:
> The bootstrap server may run on a different port (5685 ?).
> I'm not sure, what you use as filter for wireshark, may be you provide that filter?

My Bootstrap runs on 5784 and its capture for the bootstrap-client communication is actually in the second screenshot. Server and BS run on the same machine. Somehow tshark does not like to capture in localhost on the loopback interface. I used the following commands:

1)  `sudo tshark -i ens5 -f "udp port 5684" -w LwM2M.pcap`
2)  `sudo tshark -i ens5 -f "udp port 5784" -w LwM2M.pcap`
3)  `sudo tshark -i loopback -f "udp port 5784" -w LwM2M.pcap` (no captures)

On client side the connection id info is also only present when communicating with the Server

boaks:
Do you run on Windows? As far as I remember, year ago there was some magic required to capture localhost on windows, It depends also, if you use "127.0.0.1" or the external host-address, e.g. "192.168.....".

aleparmi:
> Do you run on Windows? As far as I remember, year ago there was some magic required to capture localhost on windows, It depends also, if you use "127.0.0.1" or the external host-address, e.g. "192.168.....".

Ubuntu in my case

I have also tried with tcpdump but nothing

boaks:
Hm, are you using one or two raspberries?
If you use one, then the traffic should be visible on the loopback interface, at least in my experience.
If you use two, then the traffic is not visible on the loopback interface.

sbernard31:
I feel that CID is working, the remaining problems seems more wireshark one.

I will integrate this in `master`.

aleparmi:
I'm currently performing connection updated every 15 and 60 seconds within a 10 minutes timeframe. Using PSK, Certificate and CID for both server and client. Ciphers: TLS_PSK_WITH_AES_128_CCM_8 for PSK and TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 for X.509 Cert.

As a result I get the following packets on wireshark:

image

Am I missing something?

boaks:
If possible, please select the packages and append the capture instead of a screenshot. See [Californium - how to capture](https://github.com/eclipse/californium/wiki/Logs-and-IP-Capturing-%E2%80%90-How-To-Provide-The-Right-Information).
Of interest is the related dtls handshake and application data / connection id messages.

boaks:
Maybe as hint, if wireshark is not aware of the handshake, the connection id can not be decoded.
If the addresses are changing, I guess, wireshark will also not be able to decode it.

boaks:
And, if only wireshark has an issue, but lwm2m/coap/dtls seems to work, then it may also just be an issue of that tool.

sbernard31:
> And, if only wireshark has an issue, but lwm2m/coap/dtls seems to work, then it may also just be an issue of that tool.

ConnectionID in wireshark is pretty new.
ConnectionId RFC is still a draft.
So I will not be surprised if there is issue in wireshark.
(but for same reason I guess it could be Scandium/Leshan issue too)

> I'm currently performing connection updated every 15 and 60 seconds within a 10 minutes timeframe

What do you call "connection updated", you mean registration update by sending an `UPDATE` request ?

Probably sharing the capture is the best move but eventually if you are using leshan client,  you can also share command line you run and if you modify the code it share too (link to a repository?).

aleparmi:
I see, I started the capture after the handshake; this could be the reason. Would this explanation fit with the TLSCipherText and CID missing errors as well?

```
Datagram Transport Layer Security
    DTLSv1.2 Record Layer: Connection ID
        Special Type: Connection ID (25)
        Version: DTLS 1.2 (0xfefd)
        Epoch: 1
        Sequence Number: 6
        Connection ID: <MISSING>
        Length: 47553
            [Expert Info (Error/Protocol): TLSCiphertext length MUST NOT exceed 2^14 + 2048]
                [TLSCiphertext length MUST NOT exceed 2^14 + 2048]
                [Severity level: Error]
                [Group: Protocol]
```

boaks:
Yes, the CID is exchanged in the handshake, but the length is not self-contained.
It's therefore not easy, to decode a record without knowing the cid-length.  Though the length field is after the cid-field, it may be also wrong.

boaks:
> ConnectionID in wireshark is pretty new.
> ConnectionId RFC is still a draft.

The draft made much progress in the last days, so may be it is released in the next weeks/months
What was mainly discussed was the data for the MAC and the IANA code point for the hello extension. Californium uses currently the old definitions. With the release and the new IANA code point, I can "finalize" the work there (OK, the RRC is still left ;-) ).

> So I will not be surprised if there is issue in wireshark.

It's not really an issue, maybe more a display issue. It's not intended to decode tls_cid records without knowing the cid. you may assume, that the udp package contains exactly one dtls-record and with that assumption you may try to find a valid "dtls record length" field. But I guess, that's not implemented.

> (but for same reason I guess it could be Scandium/Leshan issue too)

I guess, not :-).

aleparmi:
> What do you call "connection updated", you mean registration update by sending an UPDATE request ?

Yes 

Ok the errors disappear if I include the handshake in the capture

```
Datagram Transport Layer Security
    DTLSv1.2 Record Layer: Connection ID
        Special Type: Connection ID (25)
        Version: DTLS 1.2 (0xfefd)
        Epoch: 1
        Sequence Number: 0
        Connection ID: 245992d79d89
        Length: 49
        Encrypted Record Content: 00010000000000000993af71d6da470a6dac07d023858c659c04bdb9e98418ab621d89ec…
```

PR #1001, opened 2021-04-14
Avoid NPE when processing response message in unit tests

sophokles73:
The CoapAsyncRequestObserver spawns a cleaning task when an outbound
message is ready to be sent. When invoking the onResponse method in a
unit test to trigger the processing of an incoming response, a NPE is
thrown if onReadyToSend has not been invoked before.

All other methods use the private cancelCleaningTask method to stop the
cleaning task which performs a non-null check before stopping the task.
The onResponse method has been changed to also use this method in order
to prevent the NPE.

### DIFF
--- a/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/CoapAsyncRequestObserver.java
+++ b/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/CoapAsyncRequestObserver.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 Sierra Wireless and others.
+ * Copyright (c) 2016, 2021 Sierra Wireless and others.
  * 
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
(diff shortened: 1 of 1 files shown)


sbernard31:
I'm glad to see you back on Leshan :slightly_smiling_face: 

I read this code again, I can not see why we are using `cancelCleaningTask();` in some case and `cleaningTask.cancel(false);` in others. 

I think your fix makes the code more consistent and even maybe could avoid a unlikely race condition.

But reading this code again I'm not sure to understand why we have sometime this pattern : 
```
cancelCleaningTask();
 if (eventRaised.compareAndSet(false, true)) 
```
and sometime this one : 
```
if (eventRaised.compareAndSet(false, true)) {
    cancelCleaningTask();
```

Maybe we should make this more consistent too ? 
I feel that the second one is better as we call `cancelCleaningTask()` only once.

sophokles73:
> Maybe we should make this more consistent too ?
I feel that the second one is better as we call cancelCleaningTask() only once.

I agree, I will amend the PR accordingly ...

sophokles73:
@sbernard31 I have pushed the changes ...

sbernard31 (approved):
:+1: 
 Thx for the PR ! :pray:

sbernard31:
Integrated in `master` (commit afc43f2e472703d48151ed835a35dcfe5d6a9289)
Backported in `1.x` (commit 1d4ebbf8b0b0d3b5e12c9a510ffc6d3f2dc77164)

PR #1007, opened 2021-05-05
Refactoring in Registration class

sbernard31:
Some refactoring needed to prepare the support of "ct=" attribute at server Side.

Notable behavior breaks : 
 - equals/hashcode
 - Version used instead of String in Registration
 - Registration is no more Serializable
 - Registration Constructor takes the builder in argument instead of a too long list of arguments. 
 
 About the last point, we have many constructors with too many constructor argument in the Leshan code base  maybe we should generalized this approach ? :thinking:


msangoi (approved):
👍

PR #1008, opened 2021-05-05
Add support of "ct=" attribute at Server side

sbernard31:
Add support to `ct=` attribute at **Server** side.
"ct" attribute allow to let know to the server content formats supported by the client. 

> If the LwM2M Client supports optional data formats, it MAY inform the LwM2M Server by including the content type in the root path link using the ct= link attribute. (note that the content type value 110 is the value assigned in the CoAP Content-Format Registry for the SenML JSON format used by LwM2M).
>
> </>;ct=110, </1/0>,</1/1>,</2/0>,</2/1>,</2/2>,</2/3>,</2/4>,</3/0>,</4/0>,</5>
 
 (source [core§6.2.1](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#6-2-1-0-621-Register-Operation))

Since LWM2M 1.1, there is no more one mandatory content format  to rule them all (TLV) and so this kind of negotiation is more needed. 
(see https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/514)

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
@@ -24,7 +24,9 @@ import java.util.Comparator;
 import java.util.Date;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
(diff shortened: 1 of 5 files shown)


msangoi (approved):
👍

PR #1009, opened 2021-05-10
Remove lazy loading for getSupportedObjects in Registration.

sbernard31:
By the past `getSupportedObjects` was added in "Lazy loading" way.

But there is now more and more data to extract from LinkObject : 
 - root path
 - supported objects
 - supported content format
 - (To be done : available instances)
 
 It sounds better parse ObjectLink and extract data only when needed, then use classic serialize/deserialize for those data. (and so for supported objects too)


msangoi (approved):
👍

PR #1010, opened 2021-05-10
Add getAvailableInstances method to Registration.

sbernard31:
Object link at registration contains supported object and available instance.
This PR add an API to easily access to available instances.


leshan-server-core/src/main/java/org/eclipse/leshan/server/registration/Registration.java
```diff
@@ -82,6 +83,9 @@
     // All supported object (object id => version)
     private final Map<Integer, String> supportedObjects;
 
+    // All available instances)
```
msangoi, line 86: You can remove the trailing ')'

msangoi (approved):

PR #1018, opened 2021-06-03
Add a way to add application Data 

sbernard31:
Some Application Data can be attached to the Registration at Authorizing time using :
```
return new Registration.Builder(registration).applicationData(myAppData).build();
````
in `Authorizer`.


msangoi (approved):
Looks good! Thanks 🙏

PR #1021, opened 2021-06-07
Fix invalid syntax for ct attribute (QUOTE must be used)

### LINKED ISSUE #1020: Client registration sends ct with invalid syntax
I tried the Leshan demo client towards a non-Leshan server and noticed that the client uses invalid syntax in the Register operation payload:

```
</>;ct=60 110 112 11542 11543;rt="oma.lwm2m",</1>;ver=1.1,</1/0>,</3>;ver=1.1,</3/0>,</6/0>,</3303>;ver=1.1,</3303/0>
```

The syntax for `ct` is given in RFC 7252 (CoAP) as follows:

```ABNF
ct-value =  cardinal
         /  DQUOTE cardinal *( 1*SP cardinal ) DQUOTE
```

Even in the absence of RFC 7252 there is an ABNF in RFC 6690 (CoRE link format) that does not allow space in unquoted values.

Leshan should quote the `ct` attribute if it contains spaces, like so:

```
</>;ct="60 110 112 11542 11543";rt="oma.lwm2m",</1>;ver=1.1,</1/0>,</3>;ver=1.1,</3/0>,</6/0>,</3303>;ver=1.1,</3303/0>
```

Leshan's own parser for this data is very forgiving (too forgiving, really) which is probably why this was not caught right away.

sbernard31:
It aims to fix #1020.

### DIFF
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/util/LinkFormatHelper.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/util/LinkFormatHelper.java
@@ -56,14 +56,20 @@ public final class LinkFormatHelper {
         attributes.put("rt", "\"oma.lwm2m\"");
         // serialize contentFormat;
         if (supportedContentFormats != null && !supportedContentFormats.isEmpty()) {
-            StringBuilder b = new StringBuilder();
-            Iterator<ContentFormat> iterator = supportedContentFormats.iterator();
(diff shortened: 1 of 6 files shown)


weinholt:
I have tested the changes and the demo client's Registration operation is no longer being rejected. Looks good.

sbernard31:
:pray:  Thx for confirming that.

PR #1023, opened 2021-06-07
Migrate to Californium 3.0.

boaks:
Use 3.0.0-M2 for tests.

Signed-off-by: Achim Kraus <achim.kraus@bosch.io>

### DIFF
--- a/leshan-integration-tests/logback-leshan-test.xml
+++ b/leshan-integration-tests/logback-leshan-test.xml
@@ -28,7 +28,7 @@ Contributors:
 		</encoder>
 	</appender>
 
-	<root level="ERROR">
+	<root level="DEBUG">
 		<appender-ref ref="STDOUT" />
 	</root>
 </configuration>
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
@@ -468,9 +468,21 @@ public class LeshanServerBuilder {
             if (incompleteConfig.getStaleConnectionThreshold() == null)
                 dtlsConfigBuilder.setStaleConnectionThreshold(coapConfig.getLong(Keys.MAX_PEER_INACTIVITY_PERIOD));
 
-            // check conflict for private key
-            if (privateKey != null) {
-
+            // check conflict in configuration
+            if (incompleteConfig.getCertificateIdentityProvider() != null) {
+                if (privateKey != null) {
+                    throw new IllegalStateException(String.format(
+                            "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for private key"));
+                }
+                if (publicKey != null) {
+                    throw new IllegalStateException(String.format(
+                            "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for public key"));
+                }
+                if (certificateChain != null) {
+                    throw new IllegalStateException(String.format(
+                            "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for certificate chain"));
+                }
+            } else if (privateKey != null) {
                 // if in raw key mode and not in X.509 set the raw keys
                 if (certificateChain == null && publicKey != null) {
 
@@ -482,22 +494,22 @@ public class LeshanServerBuilder {
                     dtlsConfigBuilder.setCertificateIdentityProvider(new SingleCertificateProvider(privateKey, certificateChain, CertificateType.X_509,
                             CertificateType.RAW_PUBLIC_KEY));
                 }
+            }
 
-                // handle trusted certificates or RPK
-                if (incompleteConfig.getAdvancedCertificateVerifier() != null) {
-                    if (trustedCertificates != null) {
-                        throw new IllegalStateException(
-                                "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder: if a AdvancedCertificateVerifier is set, trustedCertificates must not be set.");
-                    }
-                } else {
-                    StaticNewAdvancedCertificateVerifier.Builder verifierBuilder = StaticNewAdvancedCertificateVerifier.builder();
-                    // by default trust all RPK
-                    verifierBuilder.setTrustAllRPKs();
-                    if (trustedCertificates != null) {
-                        verifierBuilder.setTrustedCertificates(trustedCertificates);
-                    }
-                    dtlsConfigBuilder.setAdvancedCertificateVerifier(verifierBuilder.build());
+            // handle trusted certificates or RPK
+            if (incompleteConfig.getAdvancedCertificateVerifier() != null) {
+                if (trustedCertificates != null) {
+                    throw new IllegalStateException(
+                            "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder: if a AdvancedCertificateVerifier is set, trustedCertificates must not be set.");
                 }
+            } else if (incompleteConfig.getCertificateIdentityProvider() != null){
+                StaticNewAdvancedCertificateVerifier.Builder verifierBuilder = StaticNewAdvancedCertificateVerifier.builder();
+                // by default trust all RPK
+                verifierBuilder.setTrustAllRPKs();
+                if (trustedCertificates != null) {
+                    verifierBuilder.setTrustedCertificates(trustedCertificates);
+                }
+                dtlsConfigBuilder.setAdvancedCertificateVerifier(verifierBuilder.build());
             }
 
             // Deactivate SNI by default
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
@@ -23,7 +23,6 @@ import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
-import java.util.Arrays;
 
 import org.eclipse.californium.core.network.CoapEndpoint;
 import org.eclipse.californium.core.network.config.NetworkConfig;
@@ -34,7 +33,8 @@ import org.eclipse.californium.scandium.DTLSConnector;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig.Builder;
 import org.eclipse.californium.scandium.dtls.CertificateType;
-import org.eclipse.californium.scandium.dtls.x509.BridgeCertificateVerifier;
+import org.eclipse.californium.scandium.dtls.x509.SingleCertificateProvider;
+import org.eclipse.californium.scandium.dtls.x509.StaticNewAdvancedCertificateVerifier;
 import org.eclipse.leshan.core.LwM2m;
 import org.eclipse.leshan.core.californium.DefaultEndpointFactory;
 import org.eclipse.leshan.core.californium.EndpointFactory;
@@ -470,33 +470,17 @@ public class LeshanServerBuilder {
 
             // check conflict for private key
             if (privateKey != null) {
-                if (incompleteConfig.getPrivateKey() != null && !incompleteConfig.getPrivateKey().equals(privateKey)) {
-                    throw new IllegalStateException(String.format(
-                            "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for private key: %s != %s",
-                            privateKey, incompleteConfig.getPrivateKey()));
-                }
 
                 // if in raw key mode and not in X.509 set the raw keys
                 if (certificateChain == null && publicKey != null) {
-                    if (incompleteConfig.getPublicKey() != null && !incompleteConfig.getPublicKey().equals(publicKey)) {
-                        throw new IllegalStateException(String.format(
-                                "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for public key: %s != %s",
-                                publicKey, incompleteConfig.getPublicKey()));
-                    }
 
-                    dtlsConfigBuilder.setIdentity(privateKey, publicKey);
+                    dtlsConfigBuilder.setCertificateIdentityProvider(new SingleCertificateProvider(privateKey, publicKey));
                 }
                 // if in X.509 mode set the private key, certificate chain, public key is extracted from the certificate
                 if (certificateChain != null && certificateChain.length > 0) {
-                    if (incompleteConfig.getCertificateChain() != null
-                            && !Arrays.asList(certificateChain).equals(incompleteConfig.getCertificateChain())) {
-                        throw new IllegalStateException(String.format(
-                                "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for certificate chain: %s != %s",
-                                Arrays.toString(certificateChain), incompleteConfig.getCertificateChain()));
-                    }
 
-                    dtlsConfigBuilder.setIdentity(privateKey, certificateChain, CertificateType.X_509,
-                            CertificateType.RAW_PUBLIC_KEY);
+                    dtlsConfigBuilder.setCertificateIdentityProvider(new SingleCertificateProvider(privateKey, certificateChain, CertificateType.X_509,
+                            CertificateType.RAW_PUBLIC_KEY));
                 }
 
                 // handle trusted certificates or RPK
@@ -506,33 +490,11 @@ public class LeshanServerBuilder {
                                 "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder: if a AdvancedCertificateVerifier is set, trustedCertificates must not be set.");
                     }
                 } else {
-                    BridgeCertificateVerifier.Builder verifierBuilder = new BridgeCertificateVerifier.Builder();
-                    if (incompleteConfig.getRpkTrustStore() != null) {
-                        verifierBuilder.setTrustedRPKs(incompleteConfig.getRpkTrustStore());
-                    } else {
-                        // by default trust all RPK
-                        verifierBuilder.setTrustAllRPKs();
-                    }
-                    if (incompleteConfig.getTrustStore() != null) {
-                        if (trustedCertificates != null
-                                && !Arrays.equals(trustedCertificates, incompleteConfig.getTrustStore())) {
-                            throw new IllegalStateException(String.format(
-                                    "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for trusted Certificates (trustStore) : \n%s != \n%s",
-                                    Arrays.toString(trustedCertificates),
-                                    Arrays.toString(incompleteConfig.getTrustStore())));
-                        }
-                        verifierBuilder.setTrustedCertificates(incompleteConfig.getTrustStore());
-                    } else {
-                        if (trustedCertificates != null) {
-                            verifierBuilder.setTrustedCertificates(trustedCertificates);
-                        }
-                    }
-                    if (incompleteConfig.getCertificateVerifier() != null) {
-                        if (trustedCertificates != null) {
-                            throw new IllegalStateException(
-                                    "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder: if a CertificateVerifier is set, trustedCertificates must not be set.");
-                        }
-                        verifierBuilder.setCertificateVerifier(incompleteConfig.getCertificateVerifier());
+                    StaticNewAdvancedCertificateVerifier.Builder verifierBuilder = StaticNewAdvancedCertificateVerifier.builder();
+                    // by default trust all RPK
+                    verifierBuilder.setTrustAllRPKs();
+                    if (trustedCertificates != null) {
+                        verifierBuilder.setTrustedCertificates(trustedCertificates);
                     }
                     dtlsConfigBuilder.setAdvancedCertificateVerifier(verifierBuilder.build());
                 }
--- a/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/bootstrap/LeshanBootstrapServerBuilder.java
+++ b/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/bootstrap/LeshanBootstrapServerBuilder.java
@@ -463,10 +463,24 @@ public class LeshanBootstrapServerBuilder {
             if (incompleteConfig.getStaleConnectionThreshold() == null)
                 dtlsConfigBuilder.setStaleConnectionThreshold(coapConfig.getLong(Keys.MAX_PEER_INACTIVITY_PERIOD));
 
-            if (privateKey != null) {
-
+            // check conflict in configuration
+            if (incompleteConfig.getCertificateIdentityProvider() != null) {
+                if (privateKey != null) {
+                    throw new IllegalStateException(String.format(
+                            "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for private key"));
+                }
+                if (publicKey != null) {
+                    throw new IllegalStateException(String.format(
+                            "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for public key"));
+                }
+                if (certificateChain != null) {
+                    throw new IllegalStateException(String.format(
+                            "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for certificate chain"));
+                }
+            } else if (privateKey != null) {
                 // if in raw key mode and not in X.509 set the raw keys
                 if (certificateChain == null && publicKey != null) {
+
                     dtlsConfigBuilder.setCertificateIdentityProvider(new SingleCertificateProvider(privateKey, publicKey));
                 }
                 // if in X.509 mode set the private key, certificate chain, public key is extracted from the certificate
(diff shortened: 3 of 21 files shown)


sbernard31:
I tried to build it locally with  `mvn clean  install  javadoc:javadoc`.
And get some javadoc issue in LeshanBootstrapServerBuilder about referencing removed Californium API. 
```
An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - /home/sbernard/git/leshan/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/bootstrap/LeshanBootstrapServerBuilder.java:185: error: reference not found
[ERROR]      * and {@link Builder#setTrustCertificateTypes(CertificateType...)}
[ERROR]                   ^
[ERROR] /home/sbernard/git/leshan/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/bootstrap/LeshanBootstrapServerBuilder.java:212: error: reference not found
[ERROR]      * and {@link Builder#setCertificateVerifier(org.eclipse.californium.scandium.dtls.x509.CertificateVerifier)}
```


I fixed it in a [cf-3.0.0](https://github.com/eclipse/leshan/tree/cf-3.0.0) branch which I used to test to build it on jenkins. (https://ci.eclipse.org/leshan/job/leshan-hudson-test/68/)

Does it sound OK to you : https://github.com/eclipse/leshan/commit/cb78dd0cb5827534588b1b43212f5a63c8f2aff3 ?
(Visibly there was some formatting issue too)

sbernard31:
Tests are passing on Jenkins with JDK 1.7, does it means that for now Californium is still usable with JDK 1.7 ? I know you thought to move to Java 8 (https://github.com/eclipse/californium/issues/1159#issuecomment-726746250)?

sbernard31:
When I tried to launch **redis** tests with `mvn clean  install  -Predis`, I get some tests failure : 
```
[ERROR] Errors: 
[ERROR]   RedisRegistrationTest>RegistrationTest.register_observe_deregister_observe:214 » NullPointer
[ERROR]   RedisSecurityTest>SecurityTest.registered_device_with_rpk_to_server_with_rpk:491 » Runtime
[ERROR]   SecurityTest.registered_device_with_rpk_to_server_with_x509cert » RequestCanceled
[ERROR] org.eclipse.leshan.integration.tests.observe.RedisObserveTest.can_handle_error_on_notification
[ERROR]   Run 1: RedisObserveTest>ObserveTest.can_handle_error_on_notification:295 » NullPointer
[ERROR]   Run 2: RedisObserveTest>ObserveTest.stop:71 » NullPointer raw-data connectos's addres...
[INFO] 
[ERROR] org.eclipse.leshan.integration.tests.observe.RedisObserveTest.can_observe_instance
[ERROR]   Run 1: RedisObserveTest>ObserveTest.can_observe_instance:219 » NullPointer raw-data c...
[ERROR]   Run 2: RedisObserveTest>ObserveTest.stop:71 » NullPointer raw-data connectos's addres...
[INFO] 
[ERROR] org.eclipse.leshan.integration.tests.observe.RedisObserveTest.can_observe_object
[ERROR]   Run 1: RedisObserveTest>ObserveTest.can_observe_object:257 » NullPointer raw-data con...
[ERROR]   Run 2: RedisObserveTest>ObserveTest.stop:71 » NullPointer raw-data connectos's addres...
[INFO] 
[ERROR] org.eclipse.leshan.integration.tests.observe.RedisObserveTest.can_observe_resource
[ERROR]   Run 1: RedisObserveTest>ObserveTest.can_observe_resource:80 » NullPointer raw-data co...
[ERROR]   Run 2: RedisObserveTest>ObserveTest.stop:71 » NullPointer raw-data connectos's addres...
[INFO] 
[ERROR] org.eclipse.leshan.integration.tests.observe.RedisObserveTest.can_observe_resource_then_active_cancel
[ERROR]   Run 1: RedisObserveTest>ObserveTest.can_observe_resource_then_active_cancel:163 » NullPointer
[ERROR]   Run 2: RedisObserveTest>ObserveTest.stop:71 » NullPointer raw-data connectos's addres...
[INFO] 
[ERROR] org.eclipse.leshan.integration.tests.observe.RedisObserveTest.can_observe_resource_then_passive_cancel
[ERROR]   Run 1: RedisObserveTest>ObserveTest.can_observe_resource_then_passive_cancel:114 » NullPointer
[ERROR]   Run 2: RedisObserveTest>ObserveTest.stop:71 » NullPointer raw-data connectos's addres...
```

All tests for same reason : 
```
java.lang.NullPointerException: raw-data connectos's address must not be null!
	at org.eclipse.californium.core.network.serialization.DataParser.parseMessage(DataParser.java:58)
	at org.eclipse.leshan.server.redis.serialization.ObservationSerDes.deserialize(ObservationSerDes.java:75)
	at org.eclipse.leshan.server.redis.RedisRegistrationStore.deserializeObs(RedisRegistrationStore.java:747)
	at org.eclipse.leshan.server.redis.RedisRegistrationStore.getObservations(RedisRegistrationStore.java:567)
	at org.eclipse.leshan.server.redis.RedisRegistrationStore.addObservation(RedisRegistrationStore.java:506)
	at org.eclipse.leshan.server.californium.observation.ObservationServiceImpl.addObservation(ObservationServiceImpl.java:106)
	at org.eclipse.leshan.server.californium.request.CaliforniumLwM2mRequestSender.send(CaliforniumLwM2mRequestSender.java:111)
	at org.eclipse.leshan.server.queue.QueueModeLwM2mRequestSender.send(QueueModeLwM2mRequestSender.java:60)
	at org.eclipse.leshan.server.californium.LeshanServer.send(LeshanServer.java:558)
	at org.eclipse.leshan.server.californium.LeshanServer.send(LeshanServer.java:531)
	at org.eclipse.leshan.server.californium.LeshanServer.send(LeshanServer.java:505)
	at org.eclipse.leshan.integration.tests.observe.ObserveTest.can_observe_object(ObserveTest.java:257)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:89)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:41)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:541)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:763)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:463)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:209)
```


It seems this is because we are using `RawData` and `DataParser` in a not expected way : 

https://github.com/eclipse/leshan/blob/847fb113bbbb7f3e382f6357aa2709d68f592d01/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/ObservationSerDes.java#L68-L76

Any advice about a less unorthodox way ?

leshan-integration-tests/logback-leshan-test.xml
```diff
@@ -28,7 +28,7 @@ Contributors:
 		</encoder>
 	</appender>
 
-	<root level="ERROR">
+	<root level="DEBUG">
```
sbernard31, line 31: I guess it is not expected to be changed ?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
```diff
@@ -470,33 +470,17 @@ public LeshanServer build() {
...
                 // if in raw key mode and not in X.509 set the raw keys
                 if (certificateChain == null && publicKey != null) {
-                    if (incompleteConfig.getPublicKey() != null && !incompleteConfig.getPublicKey().equals(publicKey)) {
-                        throw new IllegalStateException(String.format(
-                                "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for public key: %s != %s",
-                                publicKey, incompleteConfig.getPublicKey()));
-                    }
 
-                    dtlsConfigBuilder.setIdentity(privateKey, publicKey);
+                    dtlsConfigBuilder.setCertificateIdentityProvider(new SingleCertificateProvider(privateKey, publicKey));
```
sbernard31, line 477: I think here we should add a check in case user set a `CertificateIdentityProvider` like we do for  `CertificateVerifier` : 

```java
if (incompleteConfig.getAdvancedCertificateVerifier() != null) {
    if (trustedCertificates != null) {
        throw new IllegalStateException(
                "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder: if a AdvancedCertificateVerifier is set, trustedCertificates must not be set.");
    }
} else {
  // create default Certificate Verifier 
}

File: leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/bootstrap/LeshanBootstrapServerBuilder.java [hunk 3]
```diff
@@ -464,33 +464,16 @@ public BootstrapHandler create(BootstrapConfigurationStore store, LwM2mBootstrap
...
-                            && !Arrays.asList(certificateChain).equals(incompleteConfig.getCertificateChain())) {
-                        throw new IllegalStateException(String.format(
-                                "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for certificate chain: %s != %s",
-                                Arrays.toString(certificateChain), incompleteConfig.getCertificateChain()));
-                    }
 
-                    dtlsConfigBuilder.setIdentity(privateKey, certificateChain, CertificateType.X_509,
-                            CertificateType.RAW_PUBLIC_KEY);
+                    dtlsConfigBuilder.setCertificateIdentityProvider(new SingleCertificateProvider(privateKey, certificateChain, CertificateType.X_509,
+                            CertificateType.RAW_PUBLIC_KEY));
```
sbernard31, line 476: Same as `LeshanServerBuilder` (just above)

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
```diff
@@ -67,13 +71,12 @@ public static JsonObject serialize(EndpointContext context) {
...
         if (!attributes.isEmpty()) {
-            JsonObject attContext = Json.object();
-            for (String key : attributes.keySet()) {
-                attContext.set(key, attributes.get(key));
-            }
-            peer.set(KEY_ATTRIBUTES, attContext);
+            DatagramWriter writer = new DatagramWriter();
+            SerializationUtil.write(writer, attributes);
+            String base64 = StringUtil.byteArrayToBase64(writer.toByteArray());
+            peer.set(KEY_ATTRIBUTES, base64);
```
sbernard31, line 79: I'm not sure but to you break backward compatibility here ?

boaks:
> Tests are passing on Jenkins with JDK 1.7, does it means that for now Californium is still usable with JDK 1.7 ?

AFAIK, there is still no java 7 breaking change. The difference is more, that I focus to test with java 11. That would cause some pain, if java 7 is used and drawbacks of that are hit.

boaks:
> Does it sound OK to you : cb78dd0 ?

Yes, great. I didn't run the javadoc.

leshan-integration-tests/logback-leshan-test.xml
boaks: Yep. I'll revert it.

sbernard31:
> Should a I squash it into the PR? Or do you prefer to keep it separate and merge it in a second step?

As you prefer, but if you modify this PR just add more commit, this will be easier to review. (then I will squash all when I merge it or you do it just before I merge it)

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
boaks: The code before comes with the assumption, that if there is already something set for the cert identity, that is checked against the leshan config. 

```java
if (incompleteConfig.getPrivateKey() != null && !incompleteConfig.getPrivateKey().equals(privateKey)) {
   throw new IllegalStateException(String.format(
                            "Configuration conflict between LeshanBuilder and DtlsConnectorConfig.Builder for private key: %s != %s",
                            privateKey, incompleteConfig.getPrivateKey()));
}
```

The new interface API doesn't support that check, for Californium itself, there is no need for such a validation.
 
So there are different ways to go:
- provide a leshan specific implementation (or additional extended interface), which does that check (means, it will not work with any `CertificateProvider` but only the extended one).
- remove the check and accept, that different keys are used as configured in leshan
- or just, that's they way I did it. use the leshans certificate and provide them in a  CertificateProvider. that disables the feature to use a own implementation.

boaks:
```
java.lang.NullPointerException: raw-data connectos's address must not be null!
	at org.eclipse.californium.core.network.serialization.DataParser.parseMessage(DataParser.java:58)
```

I'm on that. But for now I didn't see the exception.
The reason behind the change is the improved support for multicast-servers. That makes to old way, using the Connectors/CoapEndpoint's address not longer a good choice. So I decided to add the receiving local address there.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
boaks: The new endpointcontext uses now a typed map, not longer only string based values. Therefore the serialization was changed as well. Yes, that breaks the API and is not backwards compatible.

sbernard31:
> (Ah, I guess, that requires more infrastructure :.-) ).

Uups, I forgot to precise that to use -Predis you need redis install.  (sry about that)
But If you don't want to install it, you should be able to reproduce this writing some code which use `ObservationSerDes.deserialize`

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
sbernard31: I proposed a fourth way : 
- if user didn't set  `CertificateProvider` set it automatically as you do, and if user set  a `CertificateProvider`  and a private key/publikey/certificate with Leshan API raise IllegalStateException. (exactly as `CertificateVerifier` example above)

Said differently : user should use the API: Either the simple Leshan one OR the full feature Californium one but not both. 

Is there any problem I missed with this way ? :thinking:

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
sbernard31: Any change to make at least the deserialization backward compatible ?

boaks:
If possible, can you test, if using

```
public final Message parseMessage(final byte[] msg);
```

instead of 

```
public final Message parseMessage(final RawData raw)
```

solves it? 
The second setup the message as "incoming", which would not really match the case her, it's outgoing.
Therefore the local-address and source-context will not be used, and the setNanoTimestamp may require additional work ( but for now, I would test it without).

adapt 

```
        RawData rawData = RawData.outbound(req, endpointContext, null, false);
        Request request = (Request) parser.parseMessage(rawData);
        request.setDestinationContext(endpointContext);
```

into

```
        Request request = (Request) parser.parseMessage(req);
        request.setDestinationContext(endpointContext);
```

and see, if that works.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
boaks: From the view point of Californium, OK.
But for leshan, I'm not that sure. For me, the  lines

```
            // check conflict for private key
            if (privateKey != null) {
```

will also disable a "default  verifier to be set". so the docu must then contain a hint, that both are required.

So, I can adapt it, but my feeling is, it will undergo some fine tuning inside leshan when using it.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
boaks: Using typed values requires either a "scheme in code" or to add the type to the serialized data. I chose the second one, because that seems to me to be most future proven. Generally, there may be solution, to add that "scheme in code" for that case.

sbernard31:
> If possible, can you test

Tested, it seems to work. :+1:

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
sbernard31: > Generally, there may be solution for backwards compatibility, to add that "scheme in code" for that case.

I'm not sure if it represent a lot of work ?
Currently I tried to keep backward compatibility for our distributed store. (based on redis)
I hope (not sure), I will be able to do it for the 2.0.0 final release. 

If we go with this kind of "scheme in code" (just for backward compatibility), should this put in Leshan or Californium ?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
sbernard31: I guess I missed something I need to double check how californium builder works now.

sbernard31: OK I think I get it. 

The code should be adapt in a way that : 
- either we set  private key/publikey/certificate  with Leshan API OR CertificateProvider with Scandium API
- either we set truststore with Leshan API OR CertificateVerifier with Scandium API.

Second check should probably not be in the ` if (privateKey != null) {`

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
boaks: leshan, EndpointContextSerDes. The downside is not the time of coding (at least I'm on it for a first shot), the main pain will be: we can't introduce new "mandatory" fields ;-(. (I even don't know, if that's the case at all, we will see. & months of change that lot isn't easy to remember.)

boaks:
The commit #0b7ca857008e132eed074476ac5b3e1ec2a25ee6 is FMPOV a bugfix, which may be worth being ported back.

#bb465cd27887b0d1465b3b263ad0394fdc5e6b38 to be tested with care. I don't have such a setup to test that migration.

#ffc86ffe412807f59c0bf25f3d24256e9305e319 next try for better config usability.

sbernard31:
> The commit 0b7ca85 is FMPOV a bugfix, which may be worth being ported back.

You mean in leshan v1.x ? 

> ffc86ff next try for better config usability.

Look good to me.

> bb465cd to be tested with care. I don't have such a setup to test that migration.

I guess writing some unit test could help (I didn't say that to make you write it)
About this point and thinking more about it, I wonder if this is not a bit strange to have a json serializer which serialize a map of attribute as a base64 payload. :thinking:

boaks:
> You mean in leshan v1.x ?

Yes.

> Look good to me.

OK, I will adapt the bootstrap server config as well, right?

> About this point and thinking more about it, I wonder if this is not a bit strange to have a json serializer which serialize a map of attribute as a base64 payload. 

If you want a json.serialization, it's no issue to replace the binary stuff. AFAIK, 3(4) types are supported, string, number (or integer and long) and "bytes". You may use JSON types, if you think, that pays off. I just didn't want to maintain two ser/der implementations.

sbernard31:
> OK, I will adapt the bootstrap server config as well, right?

Yes please.

> If you want a json.serialization, it's no issue to replace the binary stuff. AFAIK, 3(4) types are supported, string, number (or integer and long) and "bytes". You may use JSON types, if you think, that pays off. I just didn't want to maintain two ser/der implementations.

I know that's the dilemma : 
- 1 already maintained serializer but without "true" json
- or 1 new serializer to maintain.

I really don't know for now what I should do :thinking:

boaks:
Just to "justify" the typed attributes:
Adding more and more numbers, I wanted to reduce the conversions a little.
Right now, I feel, that maybe Configuration and Attributes may have more in common 

... unfortunately time runs so fast. In the meantime dtls cid is a step further and I hope to get it released in the next time (what ever that means). Zephyr has a first experimental CID PR. And I want to add dtls cid to tinydtls (by the way, you're a committer in tinydtls? I currently try to convince obgm to add my PR 89 there. And discussion about use or not to use close notify.)
So I will continue with typing/using the new Configuration. If I can adapt the Attributes as well ... I don't know.
Maybe that a good reason, to postpone a different json-serialization.

sbernard31:
> Maybe that a good reason, to postpone a different json-serialization.

I'm not sure to get what you call a different json-serialization ?
You mean a different than before or a different than the default one provided in cf-3.0.0-M2 ?

boaks:
> I'm not sure to get what you call a different json-serialization ?

Replacing the base64 with a structured json for ser/der, as you proposed. 
The point is, if the `Attributes` are changed again, that json approch may require adaption as well.
Therefore I think, for now, it's easier to postpone such a different attributes-json ser/der implementation.

sbernard31:
> The point is, if the Attributes are changed again, that json approch may require adaption as well.
> Therefore I think, for now, it's easier to postpone such a different attributes-json ser/der implementation.

Yep but I have some concern about providing a new serialization (I talked about base64 one) and change it again soon. 
Maybe we should just keep the old one with "scheme in code" for now ?

boaks:
Ah OK, that would require a serialize as well.

boaks:
>  you need redis install

I tried to adapt the serialization similar to the previos-json. But I still can't test with redis. Do I just need to install it? Or does it require a special setup?

sbernard31:
From what I remember just to install it :thinking: 
Maybe you need to run it to  ?  (on my debian systemd exists and run automatically : redis-server.service)

sbernard31:
> (by the way, you're a committer in tinydtls? I currently try to convince obgm to add my PR 89 there. And discussion about use or not to use close notify.)

I was added as committer to the project at it creation because I created a lua binding for tinydtls. Eclipse wanted more than 1 committer on the project at its creation and insist to add me.  I also did some very small contributions but I don't really feel as a real committer.  :confused:

boaks:
> From what I remember just to install it thinking

works! And the tests are successful as well.

For the client, leshan overwrites the certificate identities, if provided in the dtls-config. So there is no check required?

Do the changed behavior according certs (at least I assume it's a change) require some more documentation somewhere in javdoc. I'm not aware of that.

boaks:
> the setNanoTimestamp may require additional work ( but for now, I would test it without).

I checked that. Californium uses the time in the exchange for receiving duplicated notifies. So, that one of the message is not important for outgoing messages. it can be simply ignored.

boaks:
FMPOV, it's done, or?

sbernard31:
> For the client, leshan overwrites the certificate identities, if provided in the dtls-config. So there is no check required?

You make me think that maybe we should do a check in LeshanClientBuilder to warn that this is manage by CaliforniumEndpointsManager from data in LWM2M Security Object but that was not the case before so not really part of this PR. 

> FMPOV, it's done, or?

Yep I will give it another look then if it's OK integrate it in `master`.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
```diff
@@ -78,10 +75,12 @@ public static JsonObject serialize(EndpointContext context) {
...
         Map<String, Object> attributes = context.entries();
         if (!attributes.isEmpty()) {
-            DatagramWriter writer = new DatagramWriter();
-            SerializationUtil.write(writer, attributes);
-            String base64 = StringUtil.byteArrayToBase64(writer.toByteArray());
-            peer.set(KEY_ATTRIBUTES, base64);
+            JsonObject attContext = Json.object();
+            for (String key : attributes.keySet()) {
+                // write all values as string
+                attContext.set(key, attributes.get(key).toString());
```
sbernard31, line 81: Looking at deserialization, I saw that byte[] are decoded as hexString.
I'm not sure but it seems here this is not the case ?

boaks: Yep. I adapted it to:

```
                Object value = attributes.get(key);
                if (value instanceof Bytes) {
                    value = ((Bytes) value).getAsString();
                }
                attContext.set(key, value.toString());
``` 

I will push after running the tests.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
```diff
@@ -78,7 +79,11 @@ public static JsonObject serialize(EndpointContext context) {
             JsonObject attContext = Json.object();
             for (String key : attributes.keySet()) {
                 // write all values as string
-                attContext.set(key, attributes.get(key).toString());
+                Object value = attributes.get(key);
+                if (value instanceof Bytes) {
```
sbernard31, line 83: I was right for bad reason because I thought that byte array was `byte[]` and not `Bytes` :sweat_smile:

boaks: :-) I found so a bug in `AddressEndpointContext.getString(String key)`. `Bytes.toString()` is for logging ...

sbernard31:
Do you think it's in mergeable state now ?

boaks:
>Do you think it's in mergeable state now ?

I think so.

boaks:
Should I squash?

sbernard31:
Yes (or I can do it). 

If you squash, please integrate my javadoc fix in the final squashed commit  ( and rebase on master).

(I'm currently out of office, I'm back on thursday, I will integrate when I'm back)

boaks:
Please, do it :-).

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
sbernard31: About  `AddressEndpointContext.getString(String key)`, 

I read those code and I saw that the behavior is different than the getNumber() or getBytes()

I mean getNumber and getBytes will failed if the value is not of the expected type and getString will never failed (as toString() is called instead of a cast in String). 

Is it expected ?

sbernard31:
Integrated in master (commit 957c2791936e3dae38b295d0cdab0bcb9f63ff73)

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java
boaks: The intention was to provide a "shortcut for migration". Anyway, it's till open, if the Attributes are changing "again" to a similar API as Configuration. I'm still working the Configuration and DTLS.

sbernard31:
By the way, thx @boaks for this contribution ! :wink:

boaks:
You're welcome. Beside of thingsboard (which sticks long to version 1.0), Leshan is the most popular upstream project and so always a good PoC for changes in Californium.
( I hope the latest/upcoming Configuration redesign doesn't kill too much ;-). )

sbernard31:
> Beside of thingsboard (which sticks long to version 1.0), Leshan is the most popular upstream project and so always a good PoC for changes in Californium.
 
It seems Thingsboard are using Leshan too 
https://github.com/thingsboard/thingsboard/blob/559aa282ca0f12f949d6d1983b6744bf77b854c4/pom.xml#L71-L72

I didn't know that.

sbernard31:
I backported 0b7ca85  in `1.x` branch (commit fa2d6a9)

PR #1031, opened 2021-06-18
Added console command to send using lwm2m 1.1

bocajim:
Added a console command that can be used to send objects, instances, or resources to a LWM2M server using the LWM2M 1.1 send functionality.

example:

send 3/0 x
send 3/0 666/0 x (uses composites to send the data)

The x is used to detect the end of the input.

It currently also sends to all servers connected, not sure if there is value to target a specific server.

### DIFF
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java
@@ -71,6 +71,7 @@ import org.eclipse.leshan.client.object.Server;
 import org.eclipse.leshan.client.resource.LwM2mObjectEnabler;
 import org.eclipse.leshan.client.resource.ObjectsInitializer;
 import org.eclipse.leshan.client.resource.listener.ObjectsListenerAdapter;
+import org.eclipse.leshan.client.servers.ServerIdentity;
 import org.eclipse.leshan.core.CertificateUsage;
 import org.eclipse.leshan.core.LwM2m;
 import org.eclipse.leshan.core.californium.DefaultEndpointFactory;
@@ -83,6 +84,7 @@ import org.eclipse.leshan.core.model.ResourceModel.Type;
 import org.eclipse.leshan.core.model.StaticModel;
 import org.eclipse.leshan.core.node.codec.DefaultLwM2mNodeDecoder;
 import org.eclipse.leshan.core.node.codec.DefaultLwM2mNodeEncoder;
+import org.eclipse.leshan.core.request.ContentFormat;se
 import org.eclipse.leshan.core.util.Hex;
 import org.eclipse.leshan.core.util.SecurityUtil;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


sbernard31:
Thx for the contribution :pray: 

> It currently also sends to all servers connected, not sure if there is value to target a specific server.

Leshan client API tries to take into account support of several servers but concretely we support only 1 server for now.  

> The x is used to detect the end of the input.

I'm not sure why this is needed for other commands we don't need a special end character, end-of-line is enough.  Any reason about it ? 

Not directly linked but if I find time or if some else is interested to work on this we should migrate to another CLI (command line interface) framework : see https://github.com/eclipse/leshan/issues/569

sbernard31:
Thx for the contribution :pray: 

> It currently also sends to all servers connected, not sure if there is value to target a specific server.

Leshan client API tries to take into account support of several servers but concretely we support only 1 server for now.  

> The x is used to detect the end of the input.

I'm not sure why this is needed for other commands we don't need a special end character, end-of-line is enough.  Any reason about it ? 

Not directly linked but if I find time or if some else is interested to work on this we should migrate to another CLI (command line interface) framework : see https://github.com/eclipse/leshan/issues/569  And so the way we handle interactive command will change (see https://github.com/eclipse/leshan/issues/569#issuecomment-864910358)

sbernard31:
> I'm not sure why this is needed for other commands we don't need a special end character, end-of-line is enough. Any reason about it ?

OK, I get it. This is because of scanner behavior which handle space or end-of-line pretty much in the same way.

Maybe, we could choose this syntax : send /3/0/1;/3/0/2 ? 
This way we test if command start by send then we get next token (here /3/0/1;/3/0/2) and we split it in list of path.

-----------------------------

Some documentation should be added here : 
https://github.com/eclipse/leshan/blob/847fb113bbbb7f3e382f6357aa2709d68f592d01/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LeshanClientDemo.java#L867-L885

----------------------------

Unexpected input for this new command should be handled properly.

sbernard31:
@bocajim 

Do you plan to improve this PR ?  (let me know if this is not the case, this way I know I should adapt it myself)
As I said this code will change a lot when we will implement  :  #569.

Do you plan to add "send support" at demo server side too ? 

Could I ask how you are using Leshan ?

sbernard31:
I rewrote all the CLI interface of **leshan-client-demo** using Jline and Picocli (#1059, #1064). So this PR is a bit obsolete and should be rewritten with new API. This should be really easier to do now. I keep this PR open waiting you or me propose a new PR about this.

sbernard31:
@bocajim, I started to work on the re-implementation of this : see #1086.

I plan to finish this next week.

sbernard31:
I think #1086 is reviewable/testable now. @bocajim  you can review or give feedback if you find time. 

I think I can close this PR. Thx again for your contribution. :pray:

sbernard31:
@bocajim, since #1110 is integrated in master, Send Operation is added to **leshan-server-demo** too.

Issue #1034, opened 2021-06-24
Implementation of Observe-Composite


Michal-Wadowski:
Hello everyone!

I want to participate in the implementation of Observe-Composite functionality.

I'm working on prototype of this functionality and I have a problem:

In COAP module there is ```org.eclipse.californium.core.server.ServerMessageDeliverer.deliverRequest()``` method. This method gets ```Resource``` according to UriPath passed in request.
Then method ```checkForObserveOption()``` addeds ```ObserveRelation``` relation to the exchange. This is necessary to correlate observe client responses with server observe request.

But for the Observe-Composite should be passed multiple paths in payload. I could set ```LwM2mPath.ROOTPATH``` as target and paths in payload like it's done in ```CoapRequestBuilder.visit(ReadCompositeRequest request)```, but in that case in method ```ServerMessageDeliverer.deliverRequest()``` I will get ```RootResource``` and ```checkForObserveOption()``` will not pass, because ```RootResource``` is not observable.

Am I doing this in wrong way, or implementation of COAP may be incomplete?

Thank you in advance for your answer.

boaks:
AFAIK, this is not implemented in Californium. It may be possible to replace the `ServerMessageDeliverer` with implementation, which can deal with that. If you're interested to implement such an alternative `ServerMessageDeliverer`, let me know, if some changes (e.g. removing the `final` in `checkForObserveOption`) are required to do so.

sbernard31:
> I want to participate in the implementation of Observe-Composite functionality.

Good to know. :+1: 

I didn't look deeply at the LWM2M `Observe-Composite` feature. By the past OMA didn't really get the spirit of CoAP Observe ([rfc7641](https://datatracker.ietf.org/doc/html/rfc7641)), I hope this new observe feature will not be a new nest of issues :sweat_smile:.

Anyway, If I understand you correctly you want to make the root resource of LWM2M client observable ? 
If I'm correct, just edit : `org.eclipse.leshan.client.californium.RootResource` constructor : 
```java
    public RootResource(RegistrationEngine registrationEngine, CaliforniumEndpointsManager endpointsManager,
            BootstrapHandler bootstrapHandler, CoapServer coapServer, LwM2mRootEnabler rootEnabler,
            LwM2mNodeEncoder encoder, LwM2mNodeDecoder decoder) {
        super("", registrationEngine, endpointsManager);
        this.bootstrapHandler = bootstrapHandler;
        setVisible(false);
[...]

PR #1060, opened 2021-07-29
Observe-composite - Generalize methods parameters

Michal-Wadowski:
Partial implementation of #1034 - Generalize methods parameters to use Observation instead of SingleObservation whenever is possible


leshan-server-core/src/main/java/org/eclipse/leshan/server/observation/ObservationListener.java
```diff
@@ -50,7 +50,7 @@
      * @param response the lwm2m response received (successful or error response)
      * 
      */
-    void onResponse(SingleObservation observation, Registration registration, ObserveResponse response);
+    void onResponse(Observation observation, Registration registration, AbstractLwM2mResponse response);
```
sbernard31, line 53: You finally go in a different way than what we talk about in https://github.com/eclipse/leshan/issues/1034#issuecomment-876411280

Is it intended ? any reason about this ?

leshan-core/src/main/java/org/eclipse/leshan/core/response/ObserveResponse.java
```diff
@@ -40,17 +40,34 @@ public ObserveResponse(ResponseCode code, LwM2mNode content, List<TimestampedLwM
...
+    }
+
+    private static LwM2mNode getFirstListItemOrContent(LwM2mNode content,
+            List<TimestampedLwM2mNode> timestampedValues) {
+        return timestampedValues != null && timestampedValues.size() > 0 ? timestampedValues.get(0).getNode() : content;
+    }
+
+    private static boolean hasSingleItemWithoutTimestamp(List<TimestampedLwM2mNode> timestampedValues) {
+        return timestampedValues != null && timestampedValues.size() == 1 && !timestampedValues.get(0).isTimestamped();
     }
```
sbernard31: Does it change the behavior or is it just because you think the code is cleaner like this ?

Michal-Wadowski: I moved the logic from ObservationServiceImpl.createObserveResponse():
```java
// create lwm2m response
if (timestampedNodes.size() == 1 && !timestampedNodes.get(0).isTimestamped()) {
    return new ObserveResponse(toLwM2mResponseCode(coapResponse.getCode()),
            timestampedNodes.get(0).getNode(), null, observation, null, coapResponse);
} else {
    return new ObserveResponse(toLwM2mResponseCode(coapResponse.getCode()), null, timestampedNodes,
            observation, null, coapResponse);
}
```
into ObserveResponse class.
I don't have strong opinion about that if the code is cleaner now. It's probably less cleaner than I intended :)

leshan-server-core/src/main/java/org/eclipse/leshan/server/observation/ObservationListener.java
Michal-Wadowski: It's rather my misunderstanding your intentions. By the comment https://github.com/eclipse/leshan/issues/1034#issuecomment-877268619 I understand that we should avoid methods multiplying and we can accept casting.

Do you think I should introduce two versions of onResponse() ?

sbernard31: > Do you think I should introduce two versions of onResponse() ?

I feel this would be a better API. WDYT ?

leshan-core/src/main/java/org/eclipse/leshan/core/response/ObserveResponse.java
sbernard31: I'm not sure too. :sweat_smile: 

Generally if benefits is not so clear I prefer to not change the code so much.
(If there is clear benefits I think its important to change the code to make it cleaner)

For this one, I let you decide :)

PR #1061, opened 2021-07-29
Observe-composite server side feature

Michal-Wadowski:


leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/RootResource.java
```diff
@@ -28,10 +29,11 @@
     public RootResource() {
         super("");
         setVisible(false);
+        setObservable(true);
```
sbernard31, line 32: I'm not sure but I guess this is not intended this the client root resource which should be observable not the server one ?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java
```diff
@@ -167,8 +167,7 @@ public LeshanServerBuilder setLocalSecureAddress(InetSocketAddress localSecureAd
 
     /**
      * <p>
-     * Set your {@link RegistrationStore} implementation which stores {@link Registration}
-     * and {@link SingleObservation}.
+     * Set your {@link RegistrationStore} implementation which stores {@link Registration} and {@link Observation}.
```
sbernard31: I'm not sure but is this change belong to #1060 ?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/observation/ObservationServiceImpl.java
```diff
@@ -313,20 +315,28 @@ private AbstractLwM2mResponse createObserveResponse(Observation observation, LwM
...
         } catch (CodecException e) {
             if (LOG.isDebugEnabled()) {
                 byte[] payload = coapResponse.getPayload() == null ? new byte[0] : coapResponse.getPayload();
-                LOG.debug(String.format("Unable to decode notification payload [%s] of observation [%s] ",
+                LOG.debug(String.format("Unable to decode notification payload [%s] of abstractObservation [%s] ",
                         Hex.encodeHexString(payload), observation), e);
             }
-            throw new InvalidResponseException(e, "Unable to decode notification payload  of observation [%s] ",
+            throw new InvalidResponseException(e, "Unable to decode notification payload  of abstractObservation [%s] ",
                     observation);
```
sbernard31, line 340: Some rest from previous naming ?

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/request/CaliforniumLwM2mRequestSender.java
```diff
@@ -107,8 +105,16 @@ public CaliforniumLwM2mRequestSender(Endpoint secureEndpoint, Endpoint nonSecure
...
         // Handle special observe case
-        if (response != null && response.getClass() == ObserveResponse.class && response.isSuccess()) {
-            observationService.addObservation(destination, ((ObserveResponse) response).getObservation());
+        if (response != null && response.isSuccess()) {
+            Observation observation = null;
+            if (response instanceof ObserveResponse) {
+                observation = ((ObserveResponse) response).getObservation();
+            } else if (response instanceof ObserveCompositeResponse) {
+                observation = ((ObserveCompositeResponse) response).getObservation();
+            }
```
sbernard31, line 114: Reading this code, I asked myself I we need an interface to avoid this double `if instanceof`. But maybe this is overkill.

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/request/LwM2mResponseBuilder.java
```diff
@@ -263,7 +249,31 @@ public void visit(ReadCompositeRequest request) {
...
     @Override
     public void visit(ObserveCompositeRequest request) {
-        throw new UnsupportedOperationException("Not implemented yet.");
+        if (coapResponse.isError()) {
+            // handle error response:
+            lwM2mresponse = new ObserveCompositeResponse(toLwM2mResponseCode(coapResponse.getCode()), null,
+                    coapResponse.getPayloadString(), coapResponse, null
+            );
+
+        } else if (isResponseCodeContent() || isResponseCodeChanged()) {
```
sbernard31, line 258: I don't think we need to support response code CHANGED :
The spec says that the right code is CONTENT. (see http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Transport-V1_1_1-20190617-A.html#Table-645-1-Operation-to-Method-and-URI-Mapping-Information-Reporting-Interface)

The reason why we still support CHANGED for "single" observe response is because of historical spec changes : 
See https://github.com/eclipse/leshan/blob/a81877c5dc635f381f4dd451edb8adcfa6fbcf88/leshan-core/src/main/java/org/eclipse/leshan/core/response/ObserveResponse.java#L46L50

(I don't know if there is other code about this, if yes it should be removed I think)

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/RootResource.java
Michal-Wadowski: That's right, my mistake :)

leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/request/CaliforniumLwM2mRequestSender.java
Michal-Wadowski: Maybe we could add interface with generalized ```getObservation()``` and also add specific ```ObserveResponse.getSingleObservation()``` / ```ObserveCompositeResponse.getCompositeObservation()``` methods?

Introducing an interface for Observe*Response could nice logically group these classes together.

And what would be name for this interface? The ObserveResponse would be good one, but is already used :)

sbernard31: > And what would be name for this interface? The ObserveResponse would be good one, but is already used :)

I can not find a good name too ...
I agree that a good one would be `ObserveResponse`
But this means that we need to rename (again?) `ObserveResponse` in `SingleObserveResponse` and if we do that (This is a bit strange to keep the `ObserveRequest` name)

I can not find good solution for now :exploding_head:

Michal-Wadowski: No, I don't want go through renaming Observe* / SingleObserve* again :fearful: 

What do you think about HasObservation interface?

sbernard31: > No, I don't want go through renaming Observe* / SingleObserve* again fearful

I can understand :grimacing: 

> What do you think about HasObservation interface?

It sounds a bit strange to me.

Maybe the best move is to keep the code like this finish to implement the feature.
And we will see if we do some more refactoring like this in a second time.

sbernard31:
Integrated in #1083

PR #1062, opened 2021-07-29
Observe-composite client side feature

Michal-Wadowski:


leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/object/ObjectResource.java
```diff
@@ -434,6 +411,16 @@ public void resourceChanged(LwM2mObjectEnabler object, int instanceId, int... re
...
+
+        notifyRootResourceChanged();
+    }
+
+    private void notifyRootResourceChanged() {
+        Resource parent = getParent();
+        if (parent.getParent() == null && parent instanceof RootResource) {
+            RootResource rootResource = (RootResource) parent;
+            rootResource.changed();
+        }
```
sbernard31, line 423: (I talked about all changes in this class)

I don't think this is the right approach. 
I didn't check too much but I think this doesn't work as expected.

I write this test quickly (to add in `ObserveCompositeTest` from #1063) : 
```java
   @Test
    public void can_composite_observe_on_single_resource2() throws InterruptedException {
        ObserveCompositeResponse observeResponse = sendObserveCompose(examplePath2);

        assertObserveCompositeResponse(observeResponse);

        CompositeObservation observation = observeResponse.getObservation();
        assertObservationContainsPaths(observation, examplePath2);
        assertOneValidObservation(observation);

        writeSingleExampleValue();

        assertResponseContainsPaths(examplePath2);

        assertListenerResponseContainsValue(14, examplePath2, "+02");

        assertListenerResponse(listener);
    }
```
I observe 3/0/14 and modify 3/0/15 and I get a notification which is not expected. In other words this test should failed.
(Sry if this is confusing I should write a test which MUST succeed instead of a test which must failed but as I said I write this quickly :grimacing:)

sbernard31: So what should be the right approach : 

My feeling is that we should not add code in `ObjectResource`
`RootResource` should listen `ObjectTree`
And we need to create a `ObserveRelationFilter`, pretty much what I tried to explain here : https://github.com/eclipse/leshan/issues/1034#issuecomment-882653100

Michal-Wadowski: Now I have better understanding how Leshan events work, so I can more easily accept your proposition :) And I agree with your solution :)

sbernard31:
Integrated in #1083

PR #1063, opened 2021-07-29
Observe-Composite - Add integration tests

Michal-Wadowski:


leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/observe/ObserveCompositeTest.java
```diff
@@ -0,0 +1,253 @@
...
+    @After
+    public void stop() {
+        helper.client.destroy(false);
+        helper.server.destroy();
+        helper.dispose();
+    }
+
+    @Test
+    public void can_composite_observe_on_single_resource() throws InterruptedException {
+        ObserveCompositeResponse observeResponse = sendObserveCompose(examplePath1);
```
sbernard31, line 79: ~sendObserveCompose~  sendObserveComposite ?

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/observe/ObserveCompositeTest.java
```diff
@@ -0,0 +1,253 @@
...
+
+        assertObserveCompositeResponse(observeResponse);
```
sbernard31, line 81: assertObserveCompositeResponse is not so clear what is asserted ? 
maybe assertIsValidObserveCompositeResponse ?

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/observe/ObserveCompositeTest.java
```diff
@@ -0,0 +1,253 @@
...
+
+        CompositeObservation observation = observeResponse.getObservation();
+        assertObservationContainsPaths(observation, examplePath1);
+        assertOneValidObservation(observation);
+
+        writeSingleExampleValue();
```
sbernard31, line 87: we should see here the value you write and the path your write to make the test easier to read

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/observe/ObserveCompositeTest.java
```diff
@@ -0,0 +1,253 @@
...
+
+        assertResponseContainsPaths(examplePath1);
+
+        assertListenerResponseContainsValue(15, examplePath1, exampleValue1);
```
sbernard31, line 91: I feel those will be safer to check the whole content instead of just checking if the value is contained.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/observe/ObserveCompositeTest.java
```diff
@@ -0,0 +1,253 @@
...
+
+        assertListenerResponse(listener);
```
sbernard31, line 93: should be rename to be clearer too ?

sbernard31 (commented):
I think Tests readability could be improved here. 

Ideally, the test should be understandable just by reading the test code without reading methods called. (I know it's not so easy and probably Leshan Tests code base doesn't always respect this :grimacing: )

My remarks below are just example which should be applied to all tests. 

----------------------------

Another point, the code looks different than other integrations tests. In a general way I feel having "a code which is pretty much written in a same way" makes the whole code base easier to read. But I can understand if there is some part of the code you prefer to write in another way.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/observe/ObserveCompositeTest.java
Michal-Wadowski: Again, I tried to improve readability, but the result is questionable. And I agree that the code should be consistent. I'll try to improve the code.

sbernard31: No problem, you are doing well for a first contribution and you are open to discussion, I really appreciate this :+1:

sbernard31:
Integrated in #1083

Issue #568, opened 2018-09-20
Add some colors 🌈 to output logs


sbernard31:
It seems to be possible to print color to console output using logback : https://logback.qos.ch/manual/layouts.html#coloring

sbernard31:
The ANSI color issue will be fixed in logback 1.3.0 but for now there is only a 1.3.0-alpha5 available waiting for the v1.3.0  stable release.

sbernard31:
> The ANSI color issue will be fixed in logback 1.3.0 but for now there is only a 1.3.0-alpha5 available waiting for the v1.3.0 stable release.

Logback 1.3.0  need [java8 at runtime](http://mailman.qos.ch/pipermail/announce/2021/000156.html).

sbernard31:
I finally find a workaround about color issue to make it work with 1.2.x version of logback. (see #1068)

sbernard31:
This is now integrated in master (#1068, #1075, #1078)

PR #1068, opened 2021-08-03
Add color to logback log for leshan-client-demo

sbernard31:
This aims to implement #568.

To fix issue "using color with logback will add some crappy code color in eclipse console." (see https://github.com/eclipse/leshan/issues/568#issuecomment-429280289 for more details)

We use some workaround using a Custom  `PatternLayout` and using picocli Color Heuristic.


sbernard31:
It looks like this : 
Capture d’écran de 2021-08-03 12-27-24

PR #1083, opened 2021-08-19
Add Observe-Composite support. 

sbernard31:
This is a rebase/review/polish branch integrating : 
 - #1058 
 - #1060
 - #1061
 - #1062
 - #1066
 
 more details at https://github.com/eclipse/leshan/issues/1034#issuecomment-901152904


leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/ObserveCompositeRelationFilter.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private List<LwM2mPath> getObserveRequestPaths(Request request) {
+        ContentFormat contentFormat = ContentFormat.fromCode(request.getOptions().getContentFormat());
+
+        try {
+            return decoder.decodePaths(request.getPayload(), contentFormat);
+        } catch (CodecException e) {
+            return null;
+        }
+    }
```
sbernard31, line 94: I wonder if it's worth it to replace the decoder way and use  : 
 - `Request.setUserContext(Map<String, String>)` in RootResource. 
 - `Map<String, String> Request.getUserContext()` here.
 
 The idea is avoid to decode paths each time `ObserveRelationFilter` is called  but only first time we receive the request. 
 But the issue is that `UserContext` value string only and so we need to serialize / de-serialize it anyway.
 
 Or we need a way to store more than String at californium level.
 But If I remember well String was chosen to make serialization easier. 
 
 @boaks  maybe you have an idea for this "issue" ? (I can give more context if needed)

Michal-Wadowski: I created PR #1084 witch tries to solve this issue. I think that simple splitting values by character will end up with less overhead.

Michal-Wadowski:
@sbernard31 Yes, I think all is ok form me. I think you can merge it into master :smile:

PR #1084, opened 2021-08-31
Optimization in ObserveCompositeRelationFilter

Michal-Wadowski:
Reduce decoding overhead with passing path list via user context in ObserveCompositeRelationFilter


leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/ObserveUtil.java
```diff
@@ -133,17 +147,27 @@ public ObserveCommon(Request request) {
...
+        return context;
+    }
+
+    /**
+     * Update user context with encoded list of {@link LwM2mPath}.
+     * 
+     * @param context user context
+     * @param paths the list of {@link LwM2mPath}
+     */
+    public static void putPathsIntoContext(Map<String, String> context, List<LwM2mPath> paths) {
```
sbernard31, line 164: Maybe `addPathstoContext()` ? (if we keep `putPathsIntoContext` it's OK too)

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/ObserveUtil.java
```diff
@@ -133,17 +147,27 @@ public ObserveCommon(Request request) {
         context.put(CTX_ENDPOINT, endpoint);
         context.put(CTX_REGID, registrationId);
 
+        List<LwM2mPath> paths = request.getPaths();
+
+        putPathsIntoContext(context, paths);
```
sbernard31, line 152: WDYT about ? 

```suggestion
        putPathsIntoContext(context, request.getPaths());
```

Michal-Wadowski: No problem :smile:

sbernard31:
Thx @Michal-Wadowski  :pray:

PR #1093, opened 2021-09-10
Add TestObject to Leshan demos.

sbernard31:
This aims to implement new Test Object from OMA. (See https://github.com/OpenMobileAlliance/lwm2m-registry/pull/639)

Capture d’écran de 2021-09-10 17-37-06


sbernard31:
I added support of "multi instance" and test "execute operation with Arguments"

sbernard31:
(Implementing this I saw we do not have a very good support for Arguments of Execute Operation, so I created dedicated issue : https://github.com/eclipse/leshan/issues/1097)

Issue #1022, opened 2021-06-07
Enhance CoRE link format parsing.


sbernard31:
Currently the CoRE link format parsing is probably too permissive, see : 
- #1020 
- TODO in [Link](https://github.com/eclipse/leshan/blob/6aebf47543d4bd03e1cac70204749c94f2dfe8e1/leshan-core/src/main/java/org/eclipse/leshan/core/Link.java#L29) class

We should maybe refactor this part of the code. 

If we become more strict we probably need a way to let users provide a custom parser or any other way/HACK to support badly implemented client. 

This issue aims to centralize discussion/issue about this.

weinholt:
The current parser starts out by splitting on commas, which RFC 6690 makes the point to mention is not right:

> Note that commas can also occur in quoted strings and URIs but do not end a description.

sbernard31:
@weinholt thx for pointing this :pray: 

If you have some motivation about working on this, please do not hesitate.

Michal-Wadowski:
According to https://datatracker.ietf.org/doc/html/rfc6690#section-2 especially in "link-param" / "link-extension" I don't see possibility to create parameter without value. However the test ```LinkObjectTest.parse_with_some_attributes()``` creates link from ```</2/1> ;empty``` string and check for attribute "empty" with no value.
I missed something in spec or this part of test is wrong?

sbernard31:
I looked at the RFC quickly and I didn't see anything about creating parameter without value too.
So at first sight I would say that the test seems wrong.

weinholt:
The space in `</2/1> ;empty` seems wrong, there shouldn't be any whitespace. But isn't `empty` a link-extension? The link-extension rule is as follows:

```abnf
link-extension = ( parmname [ "=" ( ptoken / quoted-string ) ] )
               / ( ext-name-star "=" ext-value )
```

The part within `[]` is optional: <https://datatracker.ietf.org/doc/html/rfc5234#section-3.8>. I could be reading it wrong, but I think `empty` can go into the `parmname` part.

Michal-Wadowski:
@weinholt - I think you're right, square brackets are optional so the value is not mandatory. Thanks for help :smile:

sbernard31:
@weinholt thanks for putting us back on the right way :)

@Michal-Wadowski, I guess we should but as I said we should also : 

> If we become more strict we probably need a way to let users provide a custom parser or any other way/HACK to support badly implemented client.

[...]

PR #1098, opened 2021-09-24
Improved compliance with rfc6690.

Michal-Wadowski:
Improved tests and implementation to pass the new rules.
The code is not cleaned yet.


sbernard31:
Some more general point : 
1. I face some warning in eclipse about `helper.serializer` and  `helper.parser` which should be accessed in a static way.
2. about the grammar (https://github.com/eclipse/leshan/issues/1022#issuecomment-927734377) , I  understand that  URI-Reference can contains `;` or `,`
I'm not sure we want to support it. If we don't want we should maybe update the grammar we use as specification/documentation. 
3. I think I let a comment about that but I think we should create dedicated Exception like a `LinkException`. (A checked one if possible or a runtime if there is too much issue)

sbernard31:
> But I don't want to write new serializer now because it have to be written in GSON yet and needs rewrite in #1109. It's quite complicated task for now, so I choose simplest solution smile. But if you merge #1109 to master I can write proper serializer in Jacskon.

I understand,  but I have some concerns about merging #1109 now. I'm afraid we do that too hurriedly and I know with this kind of changes this is pretty common to introduce some regression.

So I need to think more about that : 
- either I keep your quick fix (maybe it's ok)
- or I tried to write a Serializer
- or we postpone this PR , and I integrate it just after the release of 2.0.0-M5.

sbernard31:
I tried to write a serializer as I feel more comfortable to not break this API for now. (8bfc8d2)
I think it will be OK for the 2.0.0-M5.

PR #1106, opened 2021-10-14
Some fix about "Adapt for new Californium Configuration."

sbernard31:
This aims to polish #1073.


sbernard31:
I discover a `Californium3.properties` file in **leshan-integration-tests**.
This was not expected, so I try to search why ?
I finally find this is because `CoapEndpoint.Builder` create by default a "Standard" Configuration.

My 2 cts, I think this is not a so good idea that the **default behavior** : 
- create a file,
- share the configuration content between several endpoint.

boaks:
> I finally find this is because CoapEndpoint.Builder create by default a "Standard" Configuration.

OK, it's not about the name. But I'm wondering, if this is new? For me this is the same as before in Californium 2.x.y?

sbernard31:
Integrated in squashed commit 64c0ba619498bba6db10797c3baac4749f82bc77

PR #1107, opened 2021-10-14
Try to add formatter-maven-plugin to ensure file are well formatted.

sbernard31:
I tried [formatter-maven-plugin](https://code.revelc.net/formatter-maven-plugin/).
But for now this is a failure as it does not exactly format as my eclipse IDE ... 

This is an interesting track but don't want to take more time one this for now.


sbernard31:
Done with #1281

PR #1104, opened 2021-10-13
Some fix about "Improved compliance with rfc6690"

sbernard31:
This is #1098 with some cosmetic fixes.

PR #1127, opened 2021-10-25
Make LinkParser extensible

Michal-Wadowski:


Michal-Wadowski:
@sbernard31 - what do you think about managing validation errors in this solution? Instead of chain of try/catch I added ```isValid()``` and ```getValidationErrorMessage()``` methods. So we can check if content is valid before parsing or we can throw ```LinkParseException``` exception if we call ```parse()``` method on invalid content (which internally call ```isValid()```/```getValidationErrorMessage()``` on sub-parser and optionally throw exception).

Michal-Wadowski:
> * drop  the "splitting first" parsing. (more details [here](https://github.com/eclipse/leshan/issues/1022#issuecomment-856117128) and [here](https://github.com/eclipse/leshan/pull/1104#issuecomment-945740962))
> 
> * better responsability handling about quoted sting/ excaping char / uri normalization (more [details](https://github.com/eclipse/leshan/issues/1022#issuecomment-938507807))

According the ```splitIgnoringEscaped``` method - the solution is written in PR #1127 and there is explanation what there happen:

If we have for example ```</foo>;param=",",</bar>``` content to parse, LinkParser have to split content by ```,``` character somehow. So at first try, at first ```,``` character LinkParser gets ```</foo>;param="```. This part is validated by subparser and if it's not valid, LinkParser continues its job. At the next ```,``` character LinkParser gets ```</foo>;param=","``` and it's validated by subparser. Now this part is valid, so LinkParser can create Link and continues its job.

The responsibility for validating link-value is moved from LinkParser to LinkValueParser, so LinkParser doesn't know anything about link-value format. The same thing is for LinkValueparser - it delegates validation of URI-Reference and link-extension to next sub-parsers.

---

According the LwM2mLinkParser and LwM2mLinks classes - the solution sounds good for me in general. But I have only one problem with understanding where it should be used. In bigger picture, where this solution will be used in our applicaiton? I don't see benefits of creating LwM2mLinkParser and LwM2mLinks yet.

Issue #1097, opened 2021-09-23
Add better support for Arguments of Execute Operation


sbernard31:
Currently we only handle argument as a String with providing any parser or validation. 

But the specification is pretty clear on this and so we should better handled it. 

See [LWM2M-v1.1@core§6.3.5-Execute-Operation](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#6-3-5-0-635-Execute-Operation)
```
arglist = arg *( "," arg )
arg = DIGIT / DIGIT "=" "'" *CHAR "'"
DIGIT = "0" / "1" / "2" / "3" / "4" / "5" / "6" / "7" / "8" / "9"
CHAR = "!" / %x23-26 / %x28-5B / %x5D-7E
```

   Examples of valid lists of arguments:
  -    5
  -    2='10.3'
  -    7, 0=' https://www.omaspecworks.org'
  -    0,1,2,3,4

sbernard31:
(This will bring some API break but maybe just the Parser/Serializer could be backported to leshan v1.x)

sbernard31:
@Michal-Wadowski I think this one is a good candidate for you ? 

Maybe you could propose an API / design for this. Either sharing some code or just explaining the idea with a comment ?

Michal-Wadowski:
Ok, I can look at this :smile:

Michal-Wadowski:
I created class that represents Arguments (one or many). I'm not sure how we should handle case that arguments are not present (they're not required). Should I create instance of Arguments that contains no arguments, or maybe should I put null in this place?
I'm not sure if Arguments without content would be valid - the format of Arguments (in ABNF syntax) doesn't allow no-content. However Arguments class is our implementation and might include case there are no argumets passed.

Michal-Wadowski:
I thought little about it and there are my conclusions: Avoiding null values are good idea because it simplifies code (skips null checking in many places). So we can put logic of handling arguments/no-arguments just into Arguments class.

sbernard31:
I'm not sure I guess most of the time you will probably replace the `== null` by `Arguments.isEmpty`(or something like this) :thinking: 
But If you feel empty Arguments is better let's go for it.

sbernard31:
A **very very small benefit** of using null as no arguments is maybe when you migrate from current API to the new one,  `== null` has same meaning ?

[...]

PR #1148, opened 2021-11-10
Initial implementation of Arguments API

Michal-Wadowski:
First design proposal for #1097.


leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/MyDevice.java
```diff
@@ -90,10 +91,10 @@ public ReadResponse read(ServerIdentity identity, int resourceid) {
     }
 
     @Override
-    public ExecuteResponse execute(ServerIdentity identity, int resourceid, String params) {
+    public ExecuteResponse execute(ServerIdentity identity, int resourceid, Arguments arguments) {
         String withParams = null;
-        if (params != null && params.length() != 0)
-            withParams = " with params " + params;
+        if (arguments.size() != 0)
+            withParams = " with params " + arguments;
```
sbernard31, line 97: Do not hesitate to replace "parameters` by `arguments` for Execute Request as this seems to be the terminology used in the specification.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/ArgumentParser.java
```diff
@@ -0,0 +1,5 @@
+package org.eclipse.leshan.core.request.execute;
+
+public interface ArgumentParser {
```
sbernard31, line 3: > I also created parser/serializer similar to Link solution. I'm not sure, maybe it's overkill for such simple format.

You maybe right about this is maybe overkill. I haven't strong opinion. 
If you simplify it what does it looks like ? 
- just removing the Interface class ? 
- moving the serialization/deserialization in  `Arguments` class ?

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/SingleArgument.java
```diff
@@ -0,0 +1,63 @@
...
+    private final int digit;
+
+    private final String value;
+
+    public SingleArgument(int digit, String value) {
+        this.digit = digit;
+        this.value = value;
+    }
+
+    public SingleArgument(int digit) {
```
sbernard31, line 14: > This is only draft, it doesn't have proper format validation yet. 

Yep I guess this is here where you will put validation and raise a kind of  `InvalidSingleArgumentException`  / `InvalidArgumentException`

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
```diff
@@ -0,0 +1,105 @@
+package org.eclipse.leshan.core.request.execute;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+public class Arguments implements Iterable<SingleArgument> {
```
sbernard31, line 9: First time, I see it. I'm asking if we really need an `Arguments` class and a `SingleArgument` one but maybe it's OK.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
```diff
@@ -0,0 +1,105 @@
...
+
+        @Override
+        public void remove() {
+            throw new UnsupportedOperationException("remove");
+        }
+    }
+
+    @Override
+    public Iterator<SingleArgument> iterator() {
+        return new ArgumentIterator(argumentList);
```
sbernard31, line 102: I'm not sure you need this ArgumentIterator list. You should just be able to return argumentList.iterator();

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private final List<SingleArgument> argumentList;
+
+    public Arguments(List<SingleArgument> argumentList) {
+        this.argumentList = argumentList;
+    }
+
+    public Arguments(Map<Integer, String> argumentsMap) {
+        this(buildSingleArguments(argumentsMap));
+    }
```
sbernard31, line 19: **maybe** we could add more convenient API like : 

Just some ideas : 
public Arguments(SingleArgument... argumentList);
public Arguments(String... values); // where digit in the index of the value.
public Arguments(Integer... digits); // for arguments without value only. 
public Arguments (Object... digitAndValues) where first element is first digit and second is first value and so on

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/MyDevice.java
```diff
@@ -90,10 +91,10 @@ public ReadResponse read(ServerIdentity identity, int resourceid) {
     }
 
     @Override
-    public ExecuteResponse execute(ServerIdentity identity, int resourceid, String params) {
+    public ExecuteResponse execute(ServerIdentity identity, int resourceid, Arguments arguments) {
         String withParams = null;
-        if (params != null && params.length() != 0)
-            withParams = " with params " + params;
+        if (arguments.size() != 0)
```
sbernard31, line 96: maybe we can add a isEmpty() ?

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/ArgumentParser.java
Michal-Wadowski: I try to follow by clean code rules. For example single-responsibility principle: parsing, serializing and holding arguments are different responsibilities. Introducing interfaces also helps with decoupling the code.

But that's only the theory - in practice often clean code means more code. Different developers have different opinions about how to organize the code, so I just try to balance.

In project we have for example Link with parsers, and LwM2mPath without parser. For LwM2mPath format is very simple, so creating parsers/serializers would be overkill. For Link the format is more complicated and by introducing parser we allow to customize the rules.

I think the arguments have quite simple format, so maybe creating parsers and serializers would introduce unnecessary noise to the code.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/SingleArgument.java
```diff
@@ -0,0 +1,63 @@
+package org.eclipse.leshan.core.request.execute;
+
+public class SingleArgument {
```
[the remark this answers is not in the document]
Michal-Wadowski: I used SingleArgument because Argument and Arguments have very similar names. I'll rename to Argument.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
Michal-Wadowski: At first glance Arguments are just list of arguments. But java list have many methods that should not be allowed, for example add() - we don't want to extend argument list once created. We could override add() method with UnsupportedOperationException, but the better is to disable this operation at compile time than runtime.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
Michal-Wadowski: I'm not sure about public ```Arguments(String... values)``` and ```Arguments(Object... digitAndValues)``` constructors. In the first one there is not clear how to use it (without refering into javadoc) the string values ```"1", "2", "3", "4"``` could be interpreted as pair digit-values (1: "2", 3: "4") or just digits (1, 2, 3, 4). How about odd number of arguments? In the second one - we can pass list of any objects, and nothing prevent us against errors. The argument validation will be performed at runtime so it's harder to detect errors.

I am believer of preventing errors by compiler as much as possible (I think everyone hate runtime errors). Parsing and validating text into arguments is reasonable in one case - when we receive content from external world. In case invalid content we should throw some validation exception. But creating object programmatically from fuzzy arguments reminds me script programming and we reject the the great advantage of strong-typed language.

I know that some other objects are created in similar way, so it's argument to continue this for consistency. But how about introducing builder pattern here? It should be also convenient and it prevents against mistakes at compile time.

Michal-Wadowski:
I'm asking myself if we should convert empty string of argument value (e.g. ```3=''```) into null or empty string. For now I don't cast "" into null at least for one reason - parsing and serialize will be reversible. And maybe empty string can mean something different than null, and specific implementation should determine the meaning, not our parser.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
sbernard31: No problem that was just idea, we will see later if we need some more constructors. 

> I know that some other objects are created in similar way, so it's argument to continue this for consistency. But how about introducing builder pattern here? It should be also convenient and it prevents against mistakes at compile time.

I don't maybe we should keep it simple and we will see if we need it at usage and so we could add it later ? 
(no strong opinion so if you feel we need it now, we can add it)

sbernard31:
> I'm asking myself if we should convert empty string of argument value (e.g. 3='') into null

I think we must not because nothing in the specification says us this must be interpreted in a same way.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
Michal-Wadowski: I added builder for Argument and Arguments. If you don't like it, I'll remove it and add constructors like you proposed. If you accept this solution, I'll cleanup the code, add some javadoc and change draft to :slightly_smiling_face:

sbernard31: I looked at this.
I understand the benefits about `Arguments` Builder but for `Argument` I'm not so sure ?

Michal-Wadowski: You're right, I created Argument builder but it isn't used anywhere. Even if I replaced new Argument() with Argument.builder() it seems to not be helpful. So I removed it.

For sake of simplify I also removed other ways of creating Arguments than with builder. If you think that creating from map or argument list is useful, I'll restore it.

sbernard31:
@Michal-Wadowski, I saw you put changes but maybe you plan more changes ? 
Tell me when you need review from me (and/or when you think this is in a mergeable state)

Michal-Wadowski:
@sbernard31 you can check PR when you have time :smile:

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
```diff
@@ -0,0 +1,260 @@
...
+     * @return {@link Arguments} represents list of arguments.
+     * @throws InvalidArgumentException if text has invalid format.
+     */
+    static Arguments parse(String arglist) throws InvalidArgumentException {
+        List<Argument> argumentList = new ArrayList<>();
+
+        if (arglist != null && !arglist.isEmpty()) {
+            int beginProbe = 0;
+            int validBegin = 0;
+            while (true) {
```
sbernard31, line 107: I'm not sure what is your opinion about this but generally I try to avoid `while(true)` loop ?

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
```diff
@@ -0,0 +1,260 @@
...
+
+            String argument = arglist.substring(beginProbe);
+            ArgumentParser argumentParser = new ArgumentParser(argument, arglist);
+            argumentList.add(argumentParser.parse());
+        }
+
+        return new Arguments(argumentList);
+    }
+
+    private static class ArgumentParser {
```
sbernard31, line 130: Maybe I'm wrong but the `ArgumentParser` class seems a bit over kill ?

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
```diff
@@ -0,0 +1,260 @@
...
+        private final String content;
+
+        public ArgumentParser(String argument, String content) {
+            this.content = content;
+            String[] keyValue = argument.split("=", 2);
+            digitPart = keyValue[0];
+            valueDecorated = keyValue.length == 1 ? null : keyValue[1];
+        }
+
+        private void validate() throws InvalidArgumentException {
```
sbernard31, line 142: I'm not sure we need a `validate` + `isValid` maybe we just raise InvalidArgumentException in `parse`

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
```diff
@@ -0,0 +1,260 @@
...
+            if (!isValid()) {
+                throw new InvalidArgumentException("Unable to parse Arguments [%s]", content);
+            }
+        }
+
+        private Argument parse() throws InvalidArgumentException {
+            validate();
+            int digit = Integer.parseInt(digitPart);
```
sbernard31, line 150: If digitPart is not a number we should raide an InvalidArgumentException

(maybe we can even create an Argument constructor which take a String as Digit and so the validation + raise exception is in the constructor ?)

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/MyDevice.java
```diff
@@ -90,12 +91,12 @@ public ReadResponse read(ServerIdentity identity, int resourceid) {
...
-        String withParams = null;
-        if (params != null && params.length() != 0)
-            withParams = " with params " + params;
+    public ExecuteResponse execute(ServerIdentity identity, int resourceid, Arguments arguments) {
+        String withArguments = null;
+        if (!arguments.isEmpty())
+            withArguments = " with arguments " + arguments;
         LOG.info("Execute on Device resource /{}/{}/{} {}", getModel().id, getId(), resourceid,
-                withParams != null ? withParams : "");
+                withArguments != null ? withArguments : "");
```
sbernard31, line 99: we should remove this test by initializing withArguments with ""

leshan-core/src/test/java/org/eclipse/leshan/core/request/execute/ArgumentValidationTest.java
```diff
@@ -0,0 +1,98 @@
...
+                        InvalidArgumentException.class), // " character
+                new TestParams(1, new String(new byte[] { 0x27 }, StandardCharsets.UTF_8),
+                        InvalidArgumentException.class), // ' character
+                new TestParams(1, new String(new byte[] { 0x5C }, StandardCharsets.UTF_8),
+                        InvalidArgumentException.class), // \ character
+                new TestParams(1, new String(new byte[] { 0x7F }, StandardCharsets.UTF_8),
+                        InvalidArgumentException.class), // DEL character
+                new TestParams(1, "`aAzZ190-=~!@#$%^&*()_+[]{}|;:<>/?,.", null) // more visualized character rules above
+        );
+    }
```
sbernard31, line 54: It's OK to me if we keep it like this but I'm not sure if you create TestParams class on purpose or because you don't know you can pass several Parameterized.parameters ? 

So just in case : https://www.tutorialspoint.com/junit/junit_parameterized_test.htm

sbernard31 (commented):
I would prefer to not move `ExecuteRequest` in execute package.

Maybe just move Arguments and all related class in a : `org.eclipse.leshan.core.argument` like we do for `Observation` ?

leshan-core/src/test/java/org/eclipse/leshan/core/request/execute/ArgumentsParsingTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        List<Argument> caughtArguments = new ArrayList<>();
+        for (Argument arg : arguments) {
+            caughtArguments.add(arg);
+        }
+
+        assertEquals(1, caughtArguments.size());
+        assertEquals(new Argument(3, "stringValue"), caughtArguments.get(0));
+        assertEquals(3, caughtArguments.get(0).getDigit());
+        assertEquals("stringValue", caughtArguments.get(0).getValue());
```
sbernard31, line 43: You need to create a list for most of the tests.

This makes me feel that maybe the API need to be improve to avoid that ?
We already have a size()
We maybe also need to getter by digit or index. (I guess digit is more appropriate)

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
```diff
@@ -0,0 +1,260 @@
...
+ */
+public class Arguments implements Iterable<Argument> {
+
+    private final List<Argument> argumentList;
+
+    /**
+     * Arguments should be build using {@link ArgumentsBuilder}.
+     */
+    private Arguments(List<Argument> argumentList) {
+        this.argumentList = argumentList;
```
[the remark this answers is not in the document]
Michal-Wadowski: I don't see any information about Argument should be unique. That is a reason why I used List instead of Map and it's a reason why I don't provided access argument by digit. I established that arguments are not unique and client implementation decide how interpret two arguments with same digit.

Or I'm wrong?

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
Michal-Wadowski: For me splitting logic into variable assigment (separatorIndex) and loop breaking looks simpler than put the two things together. And variable is used only inside loop so there is no need to declare one outside loop.

For me I don't see problem with ```while(true)``` but I can change this if you want.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
Michal-Wadowski: It's only for code logic separation. I'm not for very long classes with many methods. Usually I would put it into separated file but I don't want to generate too many files which would be harder to maintain.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
Michal-Wadowski: Ok, I can inline validation in parse.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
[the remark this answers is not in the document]
sbernard31: Good point.
My initial understanding was pretty clear that each arguments are defined by 1 digit key.
I double checked and you're right there is nothing clear about that. :grimacing: 

There is just maybe a kind of hint at [Table: D.1-3 Executable Resource Arguments Definition](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#Table-D1-3-Executable-Resource-Arguments-Definition)

ID | Resource Name | Key | Name | Type | Range or Enum | Unit | Description
-- | -- | -- | -- | -- | -- | -- | --
  |   | [0:9] | String | Data Types | If any | If any |  

As column are named **Key**. :thinking: 

Here the example in the spec about [portfolio](http://www.openmobilealliance.org/release/LWM2M_PortfolioObj/V1_0_1-20200616-A/OMA-TS-LWM2M_PortfolioObj-V1_0_1-20200616-A.pdf) but again this is only hints... :confused.

To be honest I'm not sure to see what would be the benefits to propose an arguments API like : `0='arg1',0='arg2',0='arg3'` instead of `0='arg1,arg2,arg3'`. But again this is only speculation.

So, I just asked at OMA https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/536 but it could be long to get answer (and some time we don't even get it) so we should find a short-term solution. 

So either : 
-  we keep the more open solution with maybe a no so good way to get a specific arguments.
- or we move to map and we bet that this is not allowed to have same digit twice.
- or a kind of mixed we keep the list and we provide a way to get argument by digit anyway.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
Michal-Wadowski: Oh, I missed NumberFormatException :anguished:

The second part: If the only place of Argument creation is while parsing I think it's not a good idea to introduce String-defined digit in costructor to reduce possibilities of producing error.

For sure I should improve validation while parsing.

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/MyDevice.java
Michal-Wadowski: You're right, It looks obvious now :smile:

leshan-core/src/test/java/org/eclipse/leshan/core/request/execute/ArgumentValidationTest.java
Michal-Wadowski: I know, but I wanted to use toString() to pretty print test case name. But now it looks like quite complicated so I make it in standard way.

sbernard31: You can also change the test name with : https://github.com/junit-team/junit4/wiki/Parameterized-tests#identify-individual-test-cases, but this will not be as flexible as a toString() method.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
sbernard31: Not a big deal, let's go like this. :slightly_smiling_face:

leshan-core/src/test/java/org/eclipse/leshan/core/request/execute/ArgumentsParsingTest.java
Michal-Wadowski: All the complications are because I assumed that arguments are list, not map. I'm afraid adding ```getByDigit()``` (or something like that) method in case of more items at same digit. Or maybe I just overthinking?

And what about ```getFirstByDigit()```? We can implement desired functionality and if we establish that we allow only one digit per value (it would be unique), then we can rename method to e.g. ```getByDigit()``` (or just ```get()```)

sbernard31: Yep this is related to : https://github.com/eclipse/leshan/pull/1148#discussion_r758270666

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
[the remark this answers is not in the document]
sbernard31: > I'm afraid adding getByDigit() (or something like that) method in case of more items at same digit. Or maybe I just overthinking?
>
> And what about getFirstByDigit()? We can implement desired functionality and if we establish that we allow only one digit per value (it would be unique), then we can rename method to e.g. getByDigit() (or just get())

Yep, I understand. 

Related to this issue there is also the `toMap` methods which should behave strangely if a digit is used more than once. 

My bet is that "reusing digit more than once" is either unexpected or will be rarely used. So I will go with either 2) or 3).
If we go with 3),  I propose to provide a straightforward API (pretty much the same than if we have chosen  2)) and add some javadoc to explain the behavior in case same digit is used twice (e.g. only first one is used).

The drawback with 3) the API will allow to create Arguments "reusing digit more than once", if we think we want to tolerate it but not encourage it maybe the Builder should not allow this (at all or by default) ?

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
```diff
@@ -0,0 +1,260 @@
...
+
+        if (sb.toString().length() == 0) {
+            return null;
+        }
+
+        return sb.toString();
+    }
+
+    /**
+     * Creates a map that represents arguments. Every argument's null value is converted to empty string "".
```
sbernard31, line 200: I missed that part at first review :  `Every argument's null value is converted to empty string "".`

I'm not sure we want to do that.

Michal-Wadowski: I applied this behaviour from LwM2mTestObject.storeParams() - there was ```arguments.put(id, "");``` if argument has no value. But indeed we should keep original value.

sbernard31: About `LwM2mTestObject`, I'm not sure why I did that ? :thinking:
We should probably retest this in leshan-server-demo to be sure this works with `null` instead of empty string.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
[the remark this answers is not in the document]
Michal-Wadowski: The solution 2) looks like the simplest. I'll add uniqueness validation and method access by digit.

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
Michal-Wadowski: After dropping conversion null into "" while executing demo I got ```The validated collection contains an element not of type java.lang.String at index: 1``` exception.

I think that Arguments API should be consistent according to null value, so I just remap value in LwM2mTestObject (client implementation should be responsible for handling null values).

leshan-core/src/main/java/org/eclipse/leshan/core/request/execute/Arguments.java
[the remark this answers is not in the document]
sbernard31: Let's go for 2) :slightly_smiling_face:

Michal-Wadowski:
I pushed the updates in current PR. I hope all be ok :smile:

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
```diff
@@ -0,0 +1,314 @@
...
+
+    /**
+     * Checks if  {@link Arguments} has {@link Argument} with digit.
+     */
+    public boolean hasDigit(int digit) {
+        for (Integer mapKey: argumentMap.keySet()) {
+            if (mapKey == digit) {
+                return true;
+            }
+        }
```
sbernard31, line 94: Is there any reason to not use `argumentMap.containsKey` ?

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
```diff
@@ -0,0 +1,314 @@
...
+
+    /**
+     * Gets {@link Argument} by digit.
+     *
+     * @throws IllegalArgumentException if digit doesn't exists in {@link Arguments}
+     */
+    public Argument get(int digit) {
+        if (!hasDigit(digit)) {
+            throw new IllegalArgumentException();
+        }
```
sbernard31, line 107: I'm not sure about throwing IllegalArgumentException if digit is not contained in map ?  I feel return `null` is better ? 

(Eventually we can raise an IllegalArgumentException if the digit as argument is not between 0-9; if we do that we should to it for containkey too)

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
```diff
@@ -0,0 +1,314 @@
...
+    public static ArgumentsBuilder builder() {
+        return new ArgumentsBuilder();
+    }
+
+    /**
+     * Builder for creating {$link Arguments}.
+     */
+    public static class ArgumentsBuilder {
+
+        private final List<Map<Integer, String>> keyValuePairs = new ArrayList<>();
```
sbernard31, line 268: I'm not sure to get why using a List of Map here ?

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
```diff
@@ -0,0 +1,314 @@
...
+                value = StringUtils.removeEnd(StringUtils.removeStart(valueDecorated, "'"), "'");
+            }
+
+            return new Argument(digit, value);
+        }
+
+        private boolean isValid() {
+            if (digitPart.length() != 1) {
+                return false;
+            }
```
sbernard31, line 207: Not sure to get the benefits of this test ?

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
```diff
@@ -0,0 +1,314 @@
...
+
+            if (valueDecorated != null) {
+                return valueDecorated.length() >= 2 && valueDecorated.charAt(0) == '\''
+                        && valueDecorated.charAt(valueDecorated.length() - 1) == '\'';
+            }
+            return true;
+        }
```
sbernard31, line 214: About isValid, I understood that you plan to inline the checks in `parse` (https://github.com/eclipse/leshan/pull/1148#discussion_r758264083)
Benefits will be you will be able to create more detailed error mesage

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
Michal-Wadowski: No reason, it just remainder of converting list to map. I wasn't focused enough.

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
Michal-Wadowski: I assumed that checking digit format is not necessary if we rely on existence of key - in map there is only valid keys. But you right, returning null is more similar to conventional get() method of map, so the developers should expect similar behaviour. I'll change it as you proposed.

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
Michal-Wadowski: I tried to change just into Map but I realized that we drop unique-digit checking then. The checking (of uniqueness) is made in build method, not while adding arguments. So I had to preserve multiple key structure.

Alternatively I could use List of POJO Argument (local version of Argument, without validation in constructor) instead of list of map.

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
Michal-Wadowski: You right. It was used in intermediate-developed code, but for now there is no sense to use it.

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
Michal-Wadowski: Method ArgumentParser.isValid() has two functions:
* standard checking while parsing (allow throwing an error while parsing)
* pre-validation part of text to detect if it's able to parse into argument (it helps to deal with commas inside quotes)

So I can't inline this method because it's used in both places. For more detailed error messages I could use return codes to differentiate between validation errors. In first usage I could customise error messages, second usage can ignore return code because it doesn't matter.

Michal-Wadowski:
I updated PR now

sbernard31:
I will look at this. I'm still not so sure about all the `ArgumentParser` I feel that sounds lot of code but maybe I just don't get some hidden issue. I will maybe try to write some codeabout this  to clarify my mind.

sbernard31:
#1171 (which contains this PR) is now integrated in`master`.
I guess we can close this one. 

@Michal-Wadowski thanks again for your contribution :pray:

Michal-Wadowski:
> @Michal-Wadowski thanks again for your contribution

I hope I didn't bother you too much trying to help :smile:

sbernard31:
> I hope I didn't bother you too much trying to help 

Not at all, I really appreciate your help. :pray:

sbernard31:
Good news, the OMA confirms that argument number must be unique. (https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/536#issuecomment-1038205875)

PR #1164, opened 2021-11-26
Fix bad request when creating X509 security info via REST API of `leshan-server-demo` 

### LINKED ISSUE #1163: Adding X509-based security information does not work anymore
I have tried to use the Leshan Sandbox to add security information using X.509 credentials. Only two fields need to be entered, namely the endpoint name and the security mode. 

When I click "Add", then I get an error saying "Invalid Request Body". 

This was working just recently and I assume a recent code change broke this functionality

sbernard31:
This aims to fix #1163.


sbernard31:
(Integrated in `master` commit 02c035927dddb727d902ffd3e91667ad60fc8337)

Issue #981, opened 2021-03-09
New API to handle LWM2M Data.


sbernard31:
The current Leshan API to handle LWM2M Data is based on those classes : 

- Interface [LwM2mNode](https://github.com/eclipse/leshan/blob/leshan-2.0.0-M2/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mNode.java) 
- [LwM2mObject](https://github.com/eclipse/leshan/blob/leshan-2.0.0-M2/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mObject.java) implements  LwM2mNode 
- [LwM2mObjectInstance](https://github.com/eclipse/leshan/blob/leshan-2.0.0-M2/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mObjectInstance.java)   implements LwM2mNode 
- Interface [LwM2mResource](https://github.com/eclipse/leshan/blob/leshan-2.0.0-M2/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mResource.java)   extends LwM2mNode 
- [LwM2mSingleResource](https://github.com/eclipse/leshan/blob/leshan-2.0.0-M2/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mSingleResource.java) implements  LwM2mNode 
- [LwM2mMultipleResource](https://github.com/eclipse/leshan/blob/leshan-2.0.0-M2/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mMultipleResource.java) implements  LwM2mNode 
- [LwM2mResourceInstance](https://github.com/eclipse/leshan/blob/master/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mResourceInstance.java) implements  LwM2mNode 

The data structure is a kind of Tree  : 
```
Root
    ├── Object
    │   └── ObjectInstance
    │       ├── MultipleResource 
    │       │   ├── ResourceInstance
    │       │   └── ResourceInstance
    │       └── SingleResource
    └── Object
        ├── ObjectInstance
        │   ├── SingleResource
        │   └── SingleResource
        └── ObjectInstance
            ├── SingleResource
            └── ...
    └── ...
```

[...]

PR #1171, opened 2021-12-02
Add new API/Parser/Serializer for Execute request Arguments

sbernard31:
This aims to polish #1148.
(and so implements #1097)


sbernard31:
@Michal-Wadowski you could look at this (I advice to do that commit by commit) 

Strangely there was a compilation issue that I fixed first.
Then, reading your parser, I feel that there was maybe a bug, so I create a new Unit test to be sure.
As, I have doubt about initial parser, I tried to rewrite it, I hope you find this OK ? 
Then, I move ExecuteRequest back in its initial request package.

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
```diff
@@ -0,0 +1,331 @@
...
+         */
+        public Arguments build() throws InvalidArgumentException {
+            Map<Integer, Argument> argumentMap = new LinkedHashMap<>();
+
+            for (PendingArgument argument : pendingArguments) {
+                if (argumentMap.containsKey(argument.digit)) {
+                    throw new InvalidArgumentException("Unable to build Arguments : digit [%d] exists more than once",
+                            argument.digit);
+                }
+                argumentMap.put(argument.digit, new Argument(argument.digit, argument.value));
```
sbernard31, line 316: I'm not sure if we should raise this exception on `build` like we currently do.
OR if we should raise the exception as soon as possible on `addArgument` ?

The second way seems to be better to because code will be simpler and we failed as soon as possible.

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
```diff
@@ -0,0 +1,331 @@
...
+                    cursor++;
+                } else {
+                    throw new InvalidArgumentException(
+                            "Unable to parse Arguments [%s] : [,] separator expected at index %d after [%s]", arglist,
+                            cursor, arglist.substring(0, cursor));
+                }
+            }
+        }
+        // Use builder capability to validate arguments uniqueness.
+        return builder.build();
```
sbernard31, line 169: Here there is a little issue, when build failed we don't have the whole `arglist` in the exception message.
I guess it's not a big deal and I'm not sure there is an elegant way to do that.

sbernard31:
I also think that ExecuteRequest API should be changed to take Arguments as constructor parameter.

sbernard31:
I added more minor changes (commit above :point_up:)

sbernard31:
some more minor modifications :point_up:

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
Michal-Wadowski: In my opinion in general failing on build() method instead of parameters method is better, because build() method has more context for validation. In ArgumentsBuilder it seems no matter, but to be more coherent with potential future builders it's better to make validation in build() method.

There is discussion about when to fail: https://softwareengineering.stackexchange.com/questions/241309/builder-pattern-when-to-fail

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
Michal-Wadowski: I think it's general problem with validation. If we use for validation other sub-validations we end up with more specific error message. We could catch and repackage exception to add more context. I also tried to replace throwing validation with test if object is correct - isValid() method which returns error code or error message.

leshan-core/src/test/java/org/eclipse/leshan/core/request/argument/ArgumentsParsingTest.java
```diff
@@ -112,4 +112,10 @@ public void should_not_parse_text_into_argument_with_value_without_equal() throw
     public void should_not_parse_text_into_argument_with_value_without_quotes() throws InvalidArgumentException {
         Arguments.parse("3=hello");
     }
+
+    @Test(expected = InvalidArgumentException.class)
+    public void should_not_parse_unquoted_value() throws InvalidArgumentException {
+        Arguments.parse("3=string,4='Value'");
+    }
+
```
Michal-Wadowski, line 120: That's great you find this issue. I make test for different behaviours separately to check if implementation is valid but I don't check this kind behaviours together :clap:

Michal-Wadowski:
What do you think for using builder pattern in places where there is many constructor variants (e.g. ExecuteRequest). Builder pattern can provide flexible, order-independent way to create request. It adds some complexity, but multiple constructors also are hard to maintain.

Michal-Wadowski:
And according to current PR - thank you for clean up my code and fix the issues. According to parser - we have different approach to this subject but your proposition is ok for me :+1:

Michal-Wadowski (approved):

leshan-core/src/test/java/org/eclipse/leshan/core/request/argument/ArgumentsParsingTest.java
sbernard31: I suspected it reading the code :slightly_smiling_face:

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
sbernard31: I read the discussion and I get all the pros&cons makes senses and this still hard to be to know which move in the best one.

I guess we can keep it like this.

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
sbernard31: In this particular case, do you think we should "catch and repackage exception" to add more context ? or this is not worth it ?

sbernard31:
> What do you think for using builder pattern in places where are many constructor variants (e.g. ExecuteRequest). Builder pattern can provide flexible, order-independent way to create request. It adds some complexity, but multiple constructors also are hard to maintain.

I'm not against Builder.  For request, I'm not sure if we really need it but I'm open to discuss about it. 
About Builder in a general way,  there is probably lot of place where Builder could be added in Leshan. 
One problem with builder is sometime, it does not works well with inheritance. 

In a general way, I guess for already existing API before to change it (just to be better/ not new feature) we must :
 - first identify where current API is an issue or not so good. 
 - how a new one could help. 

 (Because change the API could be an huge task and  could be annoying for users too)
 
 E.g. : https://github.com/eclipse/leshan/issues/981 I have this API changes in mind since a long time but didn't do that move for now because there is more priority task and the impact could be huge.

sbernard31:
>  thank you for clean up my code and fix the issues. 

Thx to you for the contribution, this really helps :pray: !

leshan-core/src/main/java/org/eclipse/leshan/core/request/argument/Arguments.java
Michal-Wadowski: For me repackaging exceptions are not big deal - it simplifies code. From performance point of view - invalid data should be considered as an exception rather as regular usage.

sbernard31: > From performance point of view - invalid data should be considered as an exception rather as regular usage.

I agree but I guess **in case where** there is really a big performance drawback which can be trigger by foreign peer. This could become an easy way to DoS. 

But in this particular case, I don't know and I even guess with the [default JVM behavior](https://www.oracle.com/java/technologies/javase/release-notes-introduction.html) about stacktrace this should not be an issue : 
```
The compiler in the server VM now provides correct stack backtraces for all "cold" built-in exceptions. For performance purposes, when such an exception is thrown a few times, the method may be recompiled. After recompilation, the compiler may choose a faster tactic using preallocated exceptions that do not provide a stack trace.
```

So I guess we can go this of rule: "if this is rather easy to avoid to repackaging we do it. If this is impossible or make the code really ugly in a design point of view, we repackage exception."

So in this case, I will repackage the builder exception in parse method to get a better exception message.

sbernard31:
Integrated in master (squashed in commit c461785c87934a5a6a039235b85d6264114d043e)

PR #1175, opened 2021-12-08
Proposition of OSCore Store

Michal-Wadowski:
Example implementation of OSCore Store to debate on

### DIFF
--- a/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapConfigSecurityStore.java
+++ b/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapConfigSecurityStore.java
@@ -97,20 +97,24 @@ public class BootstrapConfigSecurityStore implements BootstrapSecurityStore {
         // TODO this should be done via OSCORE store ?
         // Extract OSCORE security info
         if (bsConfig != null && bsConfig.oscore != null && !bsConfig.oscore.isEmpty()) {
-            LOG.trace("Extracting OSCORE security info for endpoint {}", endpoint);
+            LOG.trace("Looking for OSCORE security info for endpoint {}", endpoint);
 
             // First find the context for this endpoint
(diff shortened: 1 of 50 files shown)


Michal-Wadowski:
Note that I changed FileSecurityStore from instead extending of InMemorySecurityStore to composite with EditableSecurityStore. It allows me to inject OSCoreSecurityStore as a middleware.

I also have a question: do we need separate OSCore Security Store? Or maybe it should be integral part of InMemorySecurityStore/FileSecurityStore/RedisSecurityStore because it will be always OSCore support (it's not just an option).

Another question: what name we should use for classes? OSCoreXXX or OscoreXXX? Because in project we have mix of these names we should take one strategy name. Similar problem is with other classes like LwM2mXXX/Lwm2mXXX

sbernard31:
(I begin by the easy question)

>  LwM2mXXX/Lwm2mXXX

I don't like so much the `LwM2mXXX` prefix. I would prefer `Lwm2mXXX` but that was the chosen one when I integrate the project. So `Lwm2mXXX` should be not be used. Is there so class which uses this prefix ? 

> OSCoreXXX or OscoreXXX

I have no strong opinion. The "real" way to write oscore should be OSCORE (https://datatracker.ietf.org/doc/html/rfc8613)
So I guess maybe  OSCoreXXX does not make so much sense. so maybe OscoreXXX is better.

Michal-Wadowski:
> Is there so class which uses this prefix ?

There are LwM2MResourceTest and Lwm2mEndpointContextMatcher which doesn't fit into pattern. But most of classes are LwM2mXXX indeed. I just noticed Lwm2mEndpointContextMatcher class by accident but I didn't noticed that is rare case.

sbernard31:
@Michal-Wadowski, I think we can close this one ?

About the LwM2mXXX/Lwm2mXXX wording, please do not hesitate to create a dedicated issue where you list the problematic name.

sbernard31:
I close it as Oscore Store is now implemented with  #1212.

Some of your ideas was reused (like OscoreIdentity)

@Michal-Wadowski thx for that :pray:

Issue #725, opened 2019-08-12
OSCORE support in Leshan


sbernard31:
This issue aims to centralize all about OSCORE integration in Leshan 2.0.0(LWM2M 1.1).
Currently work is in progress leading by @rikard-sics, he also works on OSCORE integration in Californium.

The code will be available in a `oscore` branch, waiting we have a minimal viable feature which could be integrated in 2.0.0 branch. 

We could try to define a minimal viable feature, I see something like : 
- OSCORE over coap at client side  #726
- OSCORE over coap at server side #727

rikard-sics:
Tentative plan for next steps:

1. - Update code to use Californium 2.0.0-M17
2. - Implement an Identity and related things in SecurityChecker for OSCORE
3. - Work on the server side issues according to discussion in #727
4. - Add possibility of entering OSCORE context information as command line parameters for client

rikard-sics:
I just created PR https://github.com/eclipse/leshan/pull/749 to update the code to use Californium 2.0.0-M17 instead. I will create a new PR to change the OSCORE context database to not be a singleton in Leshan anymore.

rikard-sics:
I have now made a PR that adds support for setting OSCORE related security parameters in command line arguments to the client here https://github.com/eclipse/leshan/pull/755

sbernard31:
I add this link : https://omaspecworks.org/end-to-end-security-for-the-internet-of-things/ about this topic.

rikard-sics:
> I add this link : https://omaspecworks.org/end-to-end-security-for-the-internet-of-things/ about this topic.

Thanks for providing a link to that document. Nice to see it also brings up ACE and EDHOC. I am following the work in the IETF LAKE working group (for creating a lightweight authenticated key exchange for OSCORE). ACE is also something we are using in various use cases.

rikard-sics:
Continuing the discussion from the [last PR](https://github.com/eclipse/leshan/pull/857).

On a very high level the initial goal for me would be support for communicating with OSCORE between Client & Device Manager, communicating with OSCORE between Client & Bootstrap Server, and a Client being able receive bootstrapping information with OSCORE security material from the Bootstrap Server. Currently the first two points functionally work (but needs adjustments). 

As a bit more detailed plans these are points I can think of:

 - Rebase the current code on master
 - Remove the ID Context input on client, Device Manager and Bootstrap Server
 - Design and implement an OSCORE Store
 - Implement bootstrapping to receive OSCORE security material

[...]

PR #1180, opened 2021-12-13
OSCORE integration & unit tests

Michal-Wadowski:
I added few integration tests to check combinations:

- OSCORE client to OSCORE server
- client bootstrap via OSCORE to OSCORE server
- client bootstrap unsecured to OSCORE server
- client bootstrap via OSCORE to unsecured server

The last test is disabled because current implementation seems not working well with this combination.

In RedisSecurityTest the "registered_device_with_oscore_to_server_with_oscore" test fails because redis serializer is not adjusted to OSCORE yet.

By the way of the tests I had to split OscoreHandler to three classes (OscoreBootstrapHandler, OscoreServerHandler, OscoreClientHandler) because they use static field and while testing the client/server/bootstrap messed up each other's context.

### DIFF
--- a/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapConfigSecurityStore.java
+++ b/leshan-bsserver-demo/src/main/java/org/eclipse/leshan/server/bootstrap/demo/BootstrapConfigSecurityStore.java
@@ -28,7 +28,7 @@ import org.eclipse.californium.oscore.OSCoreCtx;
 import org.eclipse.leshan.core.SecurityMode;
 import org.eclipse.leshan.core.util.Hex;
 import org.eclipse.leshan.core.util.SecurityUtil;
-import org.eclipse.leshan.server.OscoreHandler;
+import org.eclipse.leshan.server.OscoreBootstrapHandler;
 import org.eclipse.leshan.server.bootstrap.BootstrapConfig;
 import org.eclipse.leshan.server.bootstrap.BootstrapConfig.OscoreObject;
 import org.eclipse.leshan.server.bootstrap.BootstrapConfig.ServerSecurity;
@@ -106,7 +106,7 @@ public class BootstrapConfigSecurityStore implements BootstrapSecurityStore {
                 if (security.bootstrapServer && oscoreInstanceId != null) {
                     OscoreObject oscoreObject = bsConfig.oscore.get(oscoreInstanceId);
 
-                    HashMapCtxDB db = OscoreHandler.getContextDB();
+                    HashMapCtxDB db = OscoreBootstrapHandler.getContextDB();
                     byte[] rid = Hex.decodeHex(oscoreObject.oscoreRecipientId.toCharArray());
                     OSCoreCtx ctx = db.getContext(rid);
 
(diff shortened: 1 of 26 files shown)


Michal-Wadowski:
I noticed that the many tests are broken now so I have to fix them before I add more tests.

Michal-Wadowski:
About pure unit tests - I'm not sure how add some because a lot of modified code don't have tests at all (e.g. ServersInfoExtractor, BootstrapUtil, ConfigurationChecker, InMemoryBootstrapConfigStore).

I think the better way is add unit tests while implementing new classes (e.g. OSCORE store) add functionality (OSCORE redis support).

Michal-Wadowski:
I created PR #1186 which fix OSCORE tests in redis variant.

sbernard31:
@Michal-Wadowski  sorry I didn't find time to look at this, I will be unavailable until 5th January. I will try to look at this when I'm back.

sbernard31:
I will try to look at this. 
Just to be sure the scope of this feature is to add some tests (to prepare a future rebase on master) , correct ?

sbernard31:
> In RedisSecurityTest the "registered_device_with_oscore_to_server_with_oscore" test fails because redis serializer is not adjusted to OSCORE yet.

I think just ignoring it by adding this :point_down:  in `RedisSecurityTest` could be enough for now : 
```java
    @Test
    @Ignore
    @Override
    public void registered_device_with_oscore_to_server_with_oscore()
            throws NonUniqueSecurityInfoException, InterruptedException {
        // Redis security store does support oscore for now.
    }
```

sbernard31:
Changing this in root pom.xml make the maven test OK. 
```diff
<plugin>
    <artifactId>maven-surefire-plugin</artifactId>
    <version>3.0.0-M4</version>
    <configuration>
        <systemPropertyVariables>
             <logback.configurationFile>logback-leshan-test.xml</logback.configurationFile>
        </systemPropertyVariables>
        <parallel>classes</parallel>
-        <threadCount>4</threadCount>
+        <threadCount>1</threadCount>
        <excludes>
            <exclude>**/*$*</exclude>
        </excludes>
    </configuration>
</plugin>
 ```
 
 I think it reinforces the idea that there is a shared state between the tests that creates the problem.

Michal-Wadowski:
> I think just ignoring it by adding this point_down in RedisSecurityTest could be enough for now :

Yes, temporarily is solve problem not passing tests. But I created #1186 that handles this issue.

> I think it reinforces the idea that there is a shared state between the tests that creates the problem.

I also got some problems with not passing tests randomly. But re-run often helped so I wasn't sure what was about. In current PR I fixed shared OSCORE handler at least.

sbernard31:
@Michal-Wadowski, I propose a plan at https://github.com/eclipse/leshan/issues/725#issuecomment-1007560037

Regarding 1, do you think that this :point_down: : 
- the current PR 
- the way to skip Redis Tests  : https://github.com/eclipse/leshan/pull/1180#issuecomment-1007516874
- changing temporary the number of threadcount https://github.com/eclipse/leshan/pull/1180#issuecomment-1007587985

do the job ? 
If yes I integrate this in oscore branch, and so you can start the step 2) (rebase on master)

Michal-Wadowski:
@sbernard31 - I totally agree with you. We can temporarily skip the Redis Tests and merge current work with oscore branch.

sbernard31:
I integrated this in `oscore` (commit f9df4f7889b0a67227aa7e0933451aad1e0ff07d)

You can try to do the rebase of the `oscore` branch on `master` now.

Issue #920, opened 2020-10-29
OSCORE store for holding security material


rikard-sics:
As discussed during the process of adding OSCORE support to Leshan it would be good to have a "OSCORE store", similar to the DTLS store, for holding OSCORE security information.

To start the discussion, here's a quote proposed from @sbernard31 from an earlier PR related to OSCORE. https://github.com/eclipse/leshan/pull/857
> Maybe some hints for this discussion. I think maybe we could inspired our-self of DTLS+PSK.
![Table](https://user-images.githubusercontent.com/17338658/97576212-29417a80-19ee-11eb-9871-ead0ba9e751f.png)

I will come back with further comments on considerations in a follow up post.

rikard-sics:
As we discussed the OscoreStore can be modeled on the PSKStore, but is there some class in Leshan itself which has similar properties (persistent, shared between cluster environments) which I could use as inspiration to see how these properties can be achieved?

The way I could imagine for the OscoreStore to function is as kind of a layer between the OSCORE Object and the creation of the OSCORE Context. So it would be something like:

OSCORE Object or Local configuration -> OscoreStore -> Oscore Context.

Meaning the information from the OSCORE Object received on the Client (or the local configuration input in the GUI/CLI on the Client/Server/Bootstrap server) will be input into the OscoreStore , then the information from the OscoreStore is used when building the actual OSCORE Context.

sbernard31:
> As we discussed the OscoreStore can be modeled on the PSKStore, but is there some class in Leshan itself which has similar properties (persistent, shared between cluster environments) which I could use as inspiration to see how these properties can be achieved?

You can have a look at  `SecurityStore`,  `LwM2mPskStore`, `InMemorySecurityStore`, `RedisSecurityStore`.

I'm back on Tuesday and can give you more information if needed.

rikard-sics:
> You can have a look at `SecurityStore`, `LwM2mPskStore`, `InMemorySecurityStore`, `RedisSecurityStore`.
> 
> I'm back on Tuesday and can give you more information if needed when I'll be back.

Thank you for the feedback. I will start having a look at those classes.

Michal-Wadowski:
I created draft #1175 to show my proposition

sbernard31:
About  the proposition, I feel this is not exactly the idea. 

I will try to better explain what I have in mind. 

First we should keep in mind that Leshan should be as much as possible independent to californium and californium is just a possible implementation. 

[...]

PR #1186, opened 2021-12-22
Add redis support to OSCORE

Michal-Wadowski:
I added redis serializer/deserializer to allow use OSCORE.


sbernard31:
I didn't look at the code but not sure having this kind of git tree is a good idea ? 

Capture d’écran de 2022-01-07 15-09-51

Is there any reason about this ?

sbernard31:
I'm not sure working on RedisSecurityStore is a good idea while we don't define a clear API : 
- removing static design : https://github.com/eclipse/leshan/pull/1180#issuecomment-1007524953
- and a clean oscoreStore : https://github.com/eclipse/leshan/issues/920#issuecomment-988722923

Michal-Wadowski:
> I didn't look at the code but not sure having this kind of git tree is a good idea ?

Do you mean about merging opl/oscore_work ? I borrowed some code from opl/oscore_work (which wasn't be intended to be pushed as PR) and looks what it looks like. If it would be finally squashed it shouldn't be a problem I think.

> I'm not sure working on RedisSecurityStore is a good idea while we don't define a clear API

You're right. But in current PR I mostly fixed the redis tests which are broken in PR #1180.

sbernard31:
Regarding plan at https://github.com/eclipse/leshan/issues/725#issuecomment-1007560037
This is rather tasks corresponding to cleaning and add redis support.
So maybe we can postpone it. 

> But in current PR I mostly fixed the redis tests which are broken in PR #1180.

Yep but maybe https://github.com/eclipse/leshan/pull/1180#issuecomment-1007516874, Is enough in a first time ? 

What do you think ?

Michal-Wadowski:
> Yep but maybe #1180 (comment), Is enough in a first time ?

I fixed the tests because I didn't think we can add ```@Ignore```. So we can postpone current PR :smile:

sbernard31:
I created an issue about that https://github.com/eclipse-leshan/leshan/issues/1707 so I close this obsolete PR)

PR #1188, opened 2021-12-23
Send data error handling

JaroslawLegierski:
We noticed the problem in Leshan with the data received by the server using send data. For instance if the declared model is a Boolean and the data sent is “42”. Currently, it leads to a log error and a stack trace (lwm2m o.e.l.core.californium.LwM2mCoapResource : CodecException: Invalid content X for type X for path X).
We suggest that in this case the server should respond with 400 Bad request instead of 500 server error.

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
@@ -40,8 +40,7 @@ public interface SendListener {
 
 
     /**
-     *  Called for any error e.g. if something wrong happened when we handle SendRequest,
-     *  data can't be decoded by server etc.
+     *  Called when data can't be decoded by server (because of e.g. unsupported content format or mistake in payload).
      *
      * @param registration Registration of the client which send the data.
      * @param e exception
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
@@ -19,6 +19,7 @@ import java.util.Map;
 
 import org.eclipse.leshan.core.node.LwM2mNode;
 import org.eclipse.leshan.core.request.SendRequest;
+import org.eclipse.leshan.core.request.exception.InvalidRequestException;
 import org.eclipse.leshan.server.registration.Registration;
 
 /**
@@ -38,4 +39,5 @@ public interface SendListener {
     void dataReceived(Registration registration, Map<String, LwM2mNode> data, SendRequest request);
 
     // TODO should we add a listener, if called if something wrong happened when we handle SendRequest ?
+    void onError(Registration registration, InvalidRequestException e);
 }
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
@@ -38,6 +38,6 @@ public interface SendListener {
      */
     void dataReceived(Registration registration, Map<String, LwM2mNode> data, SendRequest request);
 
-    // TODO should we add a listener, if called if something wrong happened when we handle SendRequest ?
+
     void onError(Registration registration, InvalidRequestException e);
 }
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
@@ -39,5 +39,12 @@ public interface SendListener {
     void dataReceived(Registration registration, Map<String, LwM2mNode> data, SendRequest request);
 
 
-    void onError(Registration registration, InvalidRequestException e);
+    /**
+     *  Called for any error e.g. if something wrong happened when we handle SendRequest,
+     *  data can't be decoded by server etc.
+     *
+     * @param registration Registration of the client which send the data.
+     * @param e exception
+     */
+    void onError(Registration registration, Exception e);
 }
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/SynchronousSendListener.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/SynchronousSendListener.java
@@ -39,7 +39,7 @@ public class SynchronousSendListener implements SendListener {
     }
 
     @Override
-    public void onError(Registration registration, InvalidRequestException e) {
+    public void onError(Registration registration, Exception e) {
 
     }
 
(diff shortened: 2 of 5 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
```diff
@@ -37,5 +38,6 @@
      */
     void dataReceived(Registration registration, Map<String, LwM2mNode> data, SendRequest request);
 
-    // TODO should we add a listener, if called if something wrong happened when we handle SendRequest ?
+
+    void onError(Registration registration, InvalidRequestException e);
```
sbernard31, line 42: Here we need some javadoc to explain when we get this error and this will help to better define the behavior.

I mean when should we call this  ? for any error or just when we think device send bad request ?

Maybe we should also replace `InvalidRequestException` by `Exception` ?

JaroslawLegierski: Replacing InvalidRequestException by Exception is good idea I also added a description in javadoc.

sbernard31: Reading your javadoc I understand that on Error should be called on any error.

So should we adapt the code for something llke :
```java
 @Override
    public void handlePOST(CoapExchange exchange) {
        Request coapRequest = exchange.advanced().getRequest();
        Identity sender = extractIdentity(coapRequest.getSourceContext());
        Registration registration = registrationStore.getRegistrationByIdentity(sender);

        // check we have a registration for this identity
        if (registration == null) {
            exchange.respond(ResponseCode.NOT_FOUND, "no registration found");
            return;
        }

        try {
            // Decode payload
            LwM2mModel model = modelProvider.getObjectModel(registration);
            byte[] payload = exchange.getRequestPayload();
            ContentFormat contentFormat = ContentFormat.fromCode(exchange.getRequestOptions().getContentFormat());
            if (!decoder.isSupported(contentFormat)) {
                exchange.respond(ResponseCode.BAD_REQUEST, "Unsupported content format");
                return;
            }
            Map<LwM2mPath, LwM2mNode> data = null;

            data = decoder.decodeNodes(payload, contentFormat, (List<LwM2mPath>) null, model);

            // Handle "send op request
            SendRequest sendRequest = new SendRequest(contentFormat, data, coapRequest);
            SendableResponse<SendResponse> sendableResponse = sendHandler.handleSend(registration, sendRequest);
            SendResponse response = sendableResponse.getResponse();

            // send reponse
            if (response.isSuccess()) {
                exchange.respond(toCoapResponseCode(response.getCode()));
                sendableResponse.sent();
                return;
            } else {
                exchange.respond(toCoapResponseCode(response.getCode()), response.getErrorMessage());
                sendableResponse.sent();
                return;
            }
        } catch (RuntimeException e) {
            if (e instanceof CodecException) {
                InvalidRequestException invalidreqexception = new InvalidRequestException(e);
                sendHandler.onError(registration, invalidreqexception);
                throw invalidreqexception;
            } else {
                sendHandler.onError(registration, e);
                throw e;
            }
        }
    }
```

But I'm not so sure... if we need to do that ... :thinking: 
E.g. Why do we currently call `onError` on "payload error" but not on "content format error" ? 

In your case, Do you need the `onError` ? or do you implement it because of the TODO you found in code ? 
If you need it could you describe how ? maybe this could help to better define which behavior we must implement.

JaroslawLegierski: We would like first of all to handle the situation when content format is correct (eg SENML_CBOR) but we have mistake in data e.g. Boolean data is sent as String or Integer. We'd like to detect such situations.  Why onError ? - because we were asked by colleagues who integrate Leshan with the external management system to implement this method in SendListener.

sbernard31: > Why onError ? - because we were asked by colleagues who integrate Leshan with the external management system to implement this method in SendListener.

Ok that does really help to better understand the need.

I guess, it makes sense to also raise en error about content format too.

(Just to let you know I will be unavailable until 5th January)

JaroslawLegierski: When you are mentioning about content format error do you think about something like this ? 

```java
        ContentFormat contentFormat = ContentFormat.fromCode(exchange.getRequestOptions().getContentFormat());
        if (!decoder.isSupported(contentFormat)) {
            exchange.respond(ResponseCode.BAD_REQUEST, "Unsupported content format");
            sendHandler.onError(registration, new InvalidRequestException("Unsupported content format"));
            return;
        }
```

sbernard31: Yep something like this.

sbernard31:
@JaroslawLegierski thx for the contribution :pray: 

I created a PR based on your work with some improvement.  #1195 
Could you review / test if that works for you ?

leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendListener.java
JaroslawLegierski: Thank You very much ! 
I tested it and recognized that sendHandler.onError method isn't always reached, for example when payload is incorrect: (byte[] payload = new byte[]{0x00, 0x10};) we got an error from org.eclipse.leshan.core.californium.LwM2mCoapResource - Exception while handling request because of exception during processing coapRequest.
When I removed coapRequest from InvalidRequestException :

```java
            ...
		} catch (CodecException e) {

            exchange.respond(ResponseCode.BAD_REQUEST, "Invalid Payload");
            sendHandler.onError(registration,new InvalidRequestException(e, "Invalid payload from %s",sender));
            return;
        } catch (RuntimeException e) {
            sendHandler.onError(registration, e);
            throw e;
        }
```
sendHandler.onError is called correctly.

sbernard31: I'm not sure I get it.

You face an issue using #1195 ? 
And onError is not called correctly because of usage of coapRequest in the log ? 
And so if we remove coapRequest from the log this is OK ? 
That means there is a bug in Request.toString() from `californium` ? 

Or I miss something ?

sbernard31: OK I got it this is a missing `s` after  `%` at https://github.com/eclipse/leshan/blob/56d89325302011065971a427636d932e991cd677/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/send/SendResource.java#L108

I will fix it. 

Thx you very much for testing this and reporting the error :pray:.

sbernard31: I pushed the fix and 1 integration tests about that at #1195.

If you can confirm it works for you. (Do not hesitate to comment at #1195 instead of here)

sbernard31:
This PR is included in #1195.

Issue #1187, opened 2021-12-22
Send data error handling


JaroslawLegierski:
We noticed the problem in Leshan with the data received by the server using send data. For instance if the declared model is a Boolean and the data sent is “42”. Currently, it leads to a log error and a stack trace (lwm2m o.e.l.core.californium.LwM2mCoapResource : CodecException: Invalid content X for type X for path X).

We suggest that in this case the server should respond with 400 Bad request instead of 500 server error. 

Please find below our proposal solving this problem:

[https://github.com/Przem83/leshan/tree/opl/send_data_error_handling] (https://github.com/Przem83/leshan/commit/57e8d0799ea90ebcd568c36582a97a12e49c6dcf)

@sbernard31 What is your opinion about this solution ?

sbernard31:
I think it makes totally sense. :+1:

sbernard31:
Do you plan to provide a PR ?

JaroslawLegierski:
Yes, I'll just clean the code and start the PR

sbernard31:
Great :slightly_smiling_face:

JaroslawLegierski:
PR created: Send data error handling #1188

sbernard31:
Implemented by #1195.

Thx again @JaroslawLegierski  :pray:

PR #1195, opened 2022-01-07
Better Send Operation Error handling at server side.

sbernard31:
This aims to polish #1188.
And so implement : #1187


JaroslawLegierski:
I tested error handling functionality using different payloads (correct and incorrect) and everything seems to working fine. Thank you very much for your help.

Issue #1118, opened 2021-10-20
Make LinkParser extensible.


sbernard31:
With #1022  we now have a better default parser and user is able to provide their own parser.

But the [LWM2M spec defines some specific grammar](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#5-1-2-0-512-Attributes-Classification) for its Attributes. This looks like some kind of `link-extension`.

So it seems that `DefaultLinkParser`/`LinkParser` should be extensible to allow to parse this kind of `LinkExtention`.
Ideally, Link class and Parser should not know about `Attribute` concept which is a pure LWM2M concept but I guess this is OK if `AttributeModel` contains a kind of `LinkExtensionParser`.

Implementing this will probably come with solving some raised issues like : 
 - drop  the "splitting first" parsing. (more details [here](https://github.com/eclipse/leshan/issues/1022#issuecomment-856117128) and [here](https://github.com/eclipse/leshan/pull/1104#issuecomment-945740962))
 - better responsability handling about quoted sting/ excaping char / uri normalization (more [details](https://github.com/eclipse/leshan/issues/1022#issuecomment-938507807))

(**Bonus question:** Not clear to me but Link Parsing is also related to #1042 which I didn't think too much for now)

Michal-Wadowski:
@sbernard31 - what do you think about managing validation errors in #1127 solution? Instead of chain of try/catch I added ```isValid()``` and ```getValidationErrorMessage()``` methods. So we can check if content is valid before parsing or we can throw ```LinkParseException``` exception if we call ```parse()``` method on invalid content (which internally call ```isValid()```/```getValidationErrorMessage()``` on sub-parser and optionally throw exception).

In my solution I replaced try/catch chain with ```isValid()``` / ```getValidationErrorMessage()``` chain. It reduces exceptions but adds more checks while validation. If you agree with this I can import this idea to next-approach solution.

sbernard31:
I don't know so much.  :thinking: 

Pros : 
- **a)**  Maybe one good point is that avoid the `LinkParseException` chain. (but we are not sure this is a real problem)

Cons : 
- **b)**  Maybe, a not so good point is that we need to valid first, then we parse. It looks like we get often 2 part of code which seems very similar but not exactly the same. 
- **c)** We also execute pretty much the same code twice (I don't know if this is a real issue too)

[...]

PR #1197, opened 2022-01-14
Refactoring about Link Format

sbernard31:
This is a refactoring to try to propose a API to bound `Attribute`  and `Link` world, proposing a kind of LWM2M flavored `Link`.
Some idea are described at : 
- https://github.com/eclipse/leshan/issues/1118#issuecomment-961184229
- https://github.com/eclipse/leshan/issues/1118#issuecomment-964337947

This is an huge PR ... I really don't like to do that :disappointed:...
However I tried to make it with a series of "small" consistent commit.

I will probably squash all of this in 1 commit when integrating this in `master`.

~There are renaming work but I think this should be done in some other PRs :~ 
~- Use LwM2mLink for discover~
~- Use Version instead of String for VERSION.~
~- Add notification/properties kind to model.~
~- Add range check~
~- Add applicability check~
~- quoted version invalid by default ?~
~- Make linkparser/attribute parser configurable.~
Finally done in this PR.


sbernard31:
@Michal-Wadowski I don't know maybe you want to have a look at this before I integrate it in `master` ?
(In the other hand, I could understand that you didn't want to dig in a huge PR like this :fearful:)

Michal-Wadowski:
@sbernard31 unfortunately I have limited time recently, so I can't look at this. I believe all the changes are OK :smile:

sbernard31:
I think this is in a mergeable state. 
I plan to integrate it in `master` on next Monday.

sbernard31:
This is now integrated in `master` (commit 30f79d333b3bf8cd2b4e602fc5bb797f481583b5)

nurely:
Noob here, noob to Java and everything. Somehow many stuff mentioned here is not available to the Leshan-Server-Demo. For example: 

org.eclipse.leshan.core.link.attributes does not exist. 
package org.eclipse.leshan.core.link.lwm2m.attributes does not exist

I tried clean installing.

Can somebody help me here?

When I go into details, checking out the external libraries. I do not see these files are available as it does in Master. Somehow the Maven builds are not new. If that's the case, how do I build jar directly from the master and use them in the demo server?

sbernard31:
> Somehow many stuff mentioned here is not available to the Leshan-Server-Demo

Which version are you talking about ? (where/how do you get your "leshan-server-demo" ?

Currently there is no (stable or milestone) release which integrate this. This is available in `master`.
To build from `master` : https://github.com/eclipse/leshan#compile-leshan--run-demos

HTH

sbernard31:
You can also download the already built jar of `master` looking at the https://github.com/eclipse/leshan#readme table (leshan v2.0.x)

Issue #1203, opened 2022-02-11
Add support of OSCORE in demo.


sbernard31:
We should add a way to test OSCORE with our demos (server/client/bootstrap server)

Some works was done in [`old_oscore`](https://github.com/eclipse/leshan/tree/old_oscore) but should be rewritten because most of demos code changed.

sbernard31:
#1204 aims to implement the client demo part.

sbernard31:
#1211 aims to implement the server demo part.

sbernard31:
I think this done and integrated in `master` with https://github.com/eclipse/leshan/pull/1277

So I close this one.

PR #1204, opened 2022-02-11
#1203: Add Oscore Support to leshan-client-demo

sbernard31:
This aims to implement the client part of #1203


sbernard31:
@rikard-sics, just to be sure sid, rid, msec and msalt are a byte array ? not an ascii String ?
So this arguments should be passed as an Hexadecimal form ?

rikard-sics:
> @rikard-sics, just to be sure sid, rid, msec and msalt are a byte array ? not an ascii String ?
> So this arguments should be passed as an Hexadecimal form ?

In the OSCORE RFC and its functionality these values are byte arrays. However in the [LWM2M OSCORE object](https://md2html-tool.com/docs/OpenMobileAlliance/LwM2M/master/c47850f/SUP/objects-lwm2m/master/21.html) they are defined as strings. So it depends if you, in the client input, want to consider what type OSCORE fundamentally uses, or match the type in the OSCORE object. I guess it also comes down to the prefered way of entering the values in on the bootstrap server and device management server.

There is an issue open here regarding this difference in type:
https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/521
I elaborate a bit more on this in my posts there.

sbernard31:
@rikard-sics thx for the explanation :pray:.

I have no strong skill on Security but I guess limiting the key as a String is not a good idea from a security point of view ?
And the UTF-8 way  sounds not so good as I understand that UTF-8 is just a subset of all possible byte arrays....

The OSCORE object you point is in version 2.0 and uses OPAQUE instead of STRING : [LWM2M OSCORE object](https://md2html-tool.com/docs/OpenMobileAlliance/LwM2M/master/c47850f/SUP/objects-lwm2m/master/21.html)
But when I looked at LWM2M registry the last version of OSCORE object seems to be 1.1 which still uses STRING. 
(See https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/21.xml#L68)
So not sure to get what is this 2.0 version ? is it something which will be integrated soon in LWM2M registry ? (I don't know this **md2html-tool** from where the 2.0 version comes from)

I guess this will not be so easy to support both byte and string version.
My current understanding is that using STRING was a kind of error/bug and the real good one should be to use OPAQUE.
So I would go to only support OPAQUE,  but  we need the OSCORE version 2.0 to be released. :grimacing: 

What is your opinion about this ?

rikard-sics:
> @rikard-sics thx for the explanation pray.

No problem!

> I have no strong skill in Security but I guess limiting the key as a String is not a good idea from a security point of view ? And the UTF-8 way sounds not so good as I understand that UTF-8 is just a subset of all possible byte arrays....

I think it's not a huge security problem, as long as the values have sufficient length. The big problem is more that using strings, only a subset of possible byte arrays are available as you mention.

> The OSCORE object you point is in version 2.0 and uses OPAQUE instead of STRING : [LWM2M OSCORE object](https://md2html-tool.com/docs/OpenMobileAlliance/LwM2M/master/c47850f/SUP/objects-lwm2m/master/21.html) But when I looked at LWM2M registry the last version of OSCORE object seems to be 1.1 which still uses STRING. (See https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/21.xml#L68) So not sure to get what is this 2.0 version ? is it something which will be integrated soon in LWM2M registry ? (I don't know this **md2html-tool** from where the 2.0 version comes from)

Yes, sorry, my intention was to link to the LWM2M 1.1 OSCORE object where these parameters are strings (as they are also in the 1.2 spec AFAIK). My understanding is that the link I provided you should be the latest working version of the spec (so intended to be released as the next version).

> I guess this will not be so easy to support both byte and string version. My current understanding is that using STRING was a kind of error/bug and the real good one should be to use OPAQUE. So I would go to only support OPAQUE, but we need the OSCORE version 2.0 to be released. grimacing
> 
> What is your opinion about this ?

Yes, I agree. Using OPAQUE makes more sense than STRING, and switching to OPAQUE is a kind of fix.  If possible I think it would make sense to support only OPAQUE, however if the current Leshan code is targeting LwM2M 1.1 we may need to go for STRING to be strictly following what the spec says. In such case we can go for only supporting STRING for now, and later when Leshan comes to implement the version where the OSCORE object uses OPAQUE (probably 2.0), we add OPAQUE at that stage.

sbernard31:
The v2.0 OSCORE version is compatible with LWM2M v1.1.
So I'm not sure, if we should support the STRING version one.

This is mainly use for bootstrap and there is no easy way to know the version used by the LWM2M client at bootstrap side.
(The only way is to do a Bootstrap Discover First)
Even if we get this information this will not be easy to provide a compatible way to use both version with the simple BootstrapConfigStore way. (But it is still possible to use the more complex way : BootstrapTaskProvider)


So I would rather go with supporting v2.0 by default and we will see if this is not enough. (following user feedback)
Unless you already know that you need to support the STRING version ?

sbernard31:
OSCORE v2.0 should be available soon : https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/521#issuecomment-1043048493

rikard-sics:
> The v2.0 OSCORE version is compatible with LWM2M v1.1. So I'm not sure, if we should support the STRING version one.
> 
> This is mainly use for bootstrap and there is no easy way to know the version used by the LWM2M client at bootstrap side. (The only way is to do a Bootstrap Discover First) Even if we get this information this will not be easy to provide a compatible way to use both version with the simple BootstrapConfigStore way. (But it is still possible to use the more complex way : BootstrapTaskProvider)
> 
> So I would rather go with supporting v2.0 by default and we will see if this is not enough. (following user feedback) Unless you already know that you need to support the STRING version ?

Yeah, I do not see any strong reason to support the STRING-version. So I agree it makes sense to go for only supporting the OPAQUE-version, and then possibly add support for STRING if requested by users.

sbernard31:
I change the CLI option to work with byte[] (hexadecimal should be provided)

sbernard31:
@rikard-sics I adapted this with last modification, I think this usable/testable now.

(To test with server demo, you should rather use #1211)

sbernard31:
@rikard-sics I also integrated this in `oscore` branch

PR #1208, opened 2022-02-21
Replace usage of OSCoreCtx(from cf) by OscoreSeting in SecurityInfo

sbernard31:
I cherry-pick some @Michal-Wadowski's improvement from #1175.
Which should allow to start to work on support of Oscore in server demos. (#1204)

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityInfo.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityInfo.java
@@ -12,6 +12,7 @@
  * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
+ *     Rikard Höglund (RISE SICS) - Additions to support OSCORE
  *******************************************************************************/
 package org.eclipse.leshan.server.security;
 
@@ -19,7 +20,11 @@ import java.io.Serializable;
 import java.security.PublicKey;
 import java.util.Arrays;
 
+import org.eclipse.californium.oscore.HashMapCtxDB;
+import org.eclipse.californium.oscore.OSCoreCtx;
+import org.eclipse.leshan.core.util.Hex;
 import org.eclipse.leshan.core.util.Validate;
+import org.eclipse.leshan.server.OscoreServerHandler;
 
 /**
  * The security info for a client.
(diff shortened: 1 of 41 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/security/oscore/OscoreSetting.java
```diff
@@ -0,0 +1,117 @@
...
+        return masterSecret;
+    }
+
+    public Integer getAeadAlgorithm() {
+        return aeadAlgorithm;
+    }
+
+    public Integer getHmacAlgorithm() {
+        return hmacAlgorithm;
+    }
```
sbernard31, line 61: Maybe better to replace `Integer` by dedicated class/enum, like I did in : https://github.com/eclipse/leshan/pull/1204/files#diff-593f0c45e56ce01047c78d607542064f008dda0112d69f3c3097a237c5a0e32cR26

leshan-server-core/src/main/java/org/eclipse/leshan/server/security/SecurityInfo.java
```diff
@@ -114,26 +121,42 @@ public static SecurityInfo newX509CertInfo(String endpoint) {
...
-        db.addContext(oscoreCtx);
+        db.addContext(getContext(oscoreSetting));
 
-        return new SecurityInfo(endpoint, null, null, null, false, oscoreCtx);
+        return new SecurityInfo(endpoint, null, null, null, false, oscoreSetting);
+    }
+
+    private static OSCoreCtx getContext(OscoreSetting oscoreSetting) {
+        try {
+            OSCoreCtx osCoreCtx = new OSCoreCtx(oscoreSetting.getMasterSecret(), true,
```
sbernard31, line 138: I'm not sure to understand the second argument (boolean) ?
The javadoc says: 
```
 * @param client is this originally the client's context
```
But still not clear to me, I don't know if I should use true or false. (with true or false integration tests seems to be happy :thinking:  )

Michal-Wadowski: According to the constructor of OSCoreCtx the ```boolean client``` only matters in the case if sender_id or recipient_id is not defined.

rikard-sics: Yes, indeed. That parameter is only relevant if the sender ID and recipient ID is not defined. I think the original intention was for an application the be able to create a very simple OSCORE context for testing.

sbernard31:
This is now integrated in `oscore` branch (commit 992b356)

PR #1211, opened 2022-02-23
Add Oscore SecurityInfo support to Server and BS Server demo

sbernard31:
This adds support to OSCORE for Server and Bootstrap Server demo. (https://github.com/eclipse/leshan/issues/1203).
With this PR a client should be able to connect to both server using OSCORE.
But this PR does not contain support to write OSCORE object during bootstrap session. (This will be done in another PR)

~This is work in progress, for now we are only able to provide Sender ID, Recipient ID, Master Secret.~

~:warning: Without the Oscore Store (#920), this does not really behave correctly :~
- ~it is only possible to connect to server demo (not bootstrap server). (because SecurityInfo only add context to `OscoreServerHandler.getContextDB()` not to `OscoreBootstrapHandler`.~
- ~context is not persisted on restart so you need to delete/recreate your security info to make it work again.~

~So no need to invest more time for now, let's focus on OscoreStore.~


sbernard31:
This looks like : 
--------------------------------------------------
Capture d’écran de 2022-02-23 15-14-51
----------------------------------------
Capture d’écran de 2022-02-23 15-15-38
-------------------------------------------
Capture d’écran de 2022-02-23 15-40-54

sbernard31:
@rikard-sics I adapted this with last modification, I think this usable/testable now.

(This contains all most recent changes about OSCORE for now)

sbernard31:
@rikard-sics I also integrated this in oscore branch

PR #1212, opened 2022-02-23
Some ideas about OSCORE Store

sbernard31:
This aims to clarify the ideas I try to explain at #920.

This is a kind of "fill-in-the-blank code" but the bootstrap server part is not done.

(I reuse the @Michal-Wadowski's `OscoreIdentity` from #1175 )

`OscoreSetting` and `OscoreParameters` looks like the same but one is intended to be pushed in californium and the other is  abstraction intended to be used in code without californium dependency.


leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/oscore/cf/InMemoryOscoreContextDB.java
```diff
@@ -0,0 +1,53 @@
...
+            OSCoreCtx osCoreCtx = new OSCoreCtx(oscoreSetting.getMasterSecret(), true, oscoreSetting.getAeadAlgorithm(),
+                    oscoreSetting.getSenderId(), oscoreSetting.getRecipientId(), oscoreSetting.getHmacAlgorithm(), 32,
+                    oscoreSetting.getMasterSalt(), null, 1000);
+            osCoreCtx.setContextRederivationEnabled(true);
+            return osCoreCtx;
+        } catch (OSException e) {
+            throw new IllegalStateException("Unable to create OSCoreContext", e);
+        }
+    }
+}
```
sbernard31, line 53: @rikard-sics this is the part that I'm not able to code without you :grin:

rikard-sics: So the intention is to attempt to retrieve the OSCORE context from the context DB, and if none exists to derive it? The derivation part should be done by using that constructor as you are in the getContext method. As for the context retrieval, you can check the DB for a specific context by using db.getContext(byte[] rid, byte[] IDContext), or db.getContext(byte[] rid).

Perhaps one way would be to extend HashMapCtxDB, and make sure that after getContext has been invoked, in the cases it ends up deriving a context, it is also added to the database?

Let us discuss more and let me know if I misunderstood something.

sbernard31: > So the intention is to attempt to retrieve the OSCORE context from the context DB, and if none exists to derive it?

Yes something like this.

> The derivation part should be done by using that constructor as you are in the getContext method.

Yes, context is created from OscoreParameters which can be found in the OscoreStore.

> Perhaps one way would be to extend HashMapCtxDB, and make sure that after getContext has been invoked, in the cases it ends up deriving a context, it is also added to the database?

I guess something like this but I didn't understand OSCORE enough to know how to do that in a right way.
(Note in the code above `InMemoryOscoreContextDB` already extends `HashMapCtxDB`)

rikard-sics: Right, yes I see now that InMemoryOscoreContextDB already extends HashMapCtxDB.

Hmm, so I can try to put something together where the functionality is such that if my method is called:

- if a context exists in the database matching the input argument OscoreParameters: It is retrieved and returned
- if no context exists in the database matching the input argument OscoreParameters: It is derived, added to the database and returned

rikard-sics: So to understand better, is the idea is to store and retrieve the OSCORE contexts from the OscoreStore, rather than the maps in the HashMapCtxDB? In such case it could be better to rather implement OSCoreCtxDB I believe.

If the above is not correct I have the following draft code. Note that the way things work now is that the ObjectSecurityLayer does the job of retrieving the correct context for a request or response based either on information in their OSCORE options, or URI. So currently the intent is to not control for each individual message what context to use (that can be changed if needed). This is also why when adding the context on the server and client side it needs to be associated with the URi of the client and server respectively. That is the way the ObjectSecurityLayer finds the context for outgoing requests. So this is why in the previous code on the server side there is the code to "re-add" the context to the DB after the first request from the client since it is only then the URI (hostname+port) of the client is known.

```
/*******************************************************************************
 * Copyright (c) 2022 Sierra Wireless and others.
 * 
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v2.0
 * and Eclipse Distribution License v1.0 which accompany this distribution.
 * 
 * The Eclipse Public License is available at
 *    http://www.eclipse.org/legal/epl-v20.html
 * and the Eclipse Distribution License is available at
 *    http://www.eclipse.org/org/documents/edl-v10.html.
 * 
 * Contributors:
 *     Sierra Wireless - initial API and implementation
package org.eclipse.leshan.core.californium.oscore.cf;

import org.eclipse.californium.oscore.CoapOSException;
import org.eclipse.californium.oscore.HashMapCtxDB;
import org.eclipse.californium.oscore.OSCoreCtx;
import org.eclipse.californium.oscore.OSCoreCtxDB;
import org.eclipse.californium.oscore.OSException;

/**
 * An {@link OSCoreCtxDB} which store context in memory and is able to derive context from data provided in
 * {@link OscoreStore}
 *
 */
// TODO OSCORE this should be moved in californium.
// TODO OSCORE I don't know if we need to extends HashMapCtxDB or implement OSCoreCtxDB
public class InMemoryOscoreContextDB extends HashMapCtxDB {

    private OscoreStore store;

    public InMemoryOscoreContextDB(OscoreStore oscoreStore) {
        this.store = oscoreStore;
    }

    // TODO OSCORE I guess here we should override some method
    // When there is no context in DB we should create/derive it from OscoreStore, right ?

    // TODO OSCORE I guess a method like this will be be useful
    private OSCoreCtx getContext(OscoreParameters oscoreSetting) {

        // Attempt to retrieve the context from the DB. Return it if found, otherwise derive and store it.
        OSCoreCtx osCoreCtx;
        try {
            osCoreCtx = getContext(oscoreSetting.getRecipientId(), null);
        } catch (CoapOSException e) {
            throw new IllegalStateException("Error when attempting to retrieve OSCoreContext", e);
        }
        if (osCoreCtx != null) {
            return osCoreCtx;
        }

        try {
            osCoreCtx = new OSCoreCtx(oscoreSetting.getMasterSecret(), true, oscoreSetting.getAeadAlgorithm(),
                    oscoreSetting.getSenderId(), oscoreSetting.getRecipientId(), oscoreSetting.getHmacAlgorithm(), 32,
                    oscoreSetting.getMasterSalt(), null, 1000);
            osCoreCtx.setContextRederivationEnabled(true);
            addContext(osCoreCtx); // May need to add associated to client URI too
            return osCoreCtx;
        } catch (OSException e) {
            throw new IllegalStateException("Unable to create OSCoreContext", e);
        }
    }
}

```

sbernard31: > So to understand better, is the idea is to store and retrieve the OSCORE contexts from the OscoreStore, rather than the maps in the HashMapCtxDB? In such case it could be better to rather implement OSCoreCtxDB I believe.

I don't think so. (Or maybe I don't get your point :sweat_smile:)

The OscoreScore contains user setting (OscoreParameters) which could be persisted and/or shared.
The OscoreContextDB contains the real context (OSCoreContext). 

OSCoreContext is what is really needed to decrypt payload. (it can be lost because it can be regenerate from OscoreParameters)
OscoreParameters is the User inputs which must not be lost. 

OscoreStore/OscoreParameters will be directly used by the user. 
OscoreContextDB is maybe more internal or advance API. 

This is pretty much what I try to explain at :
- https://github.com/eclipse/leshan/issues/920#issue-732289668
- https://github.com/eclipse/leshan/issues/920#issuecomment-988722923

sbernard31: About the code you share, I don't get how it could work ? 
Should you override some `HashMapCtxDB` or `OSCoreCtxDB` ?

rikard-sics: > I don't think so. (Or maybe I don't get your point sweat_smile)

I see, that may be good then if the second understanding that I based the code example on is more accurate.

> About the code you share, I don't get how it could work ?
> Should you override some `HashMapCtxDB` or `OSCoreCtxDB` methods?

The idea was to use _getContext(byte[] rid, byte[] IDContext)_ and _addContext(OSCoreCtx ctx)_ from the underlying HashMapCtxDB to store and retrieve contexts in its maps.

Thinking further, so what about overriding all the get* methods from HashMapCtxDB in such a way that it initially tries to retrieve the context from the DB, but if that fails derives and stores the context in the DB (based on info in OscoreStore). I believe that could work for _OSCoreCtx getContext(byte[] cid)_ and _getContext(byte[] cid, byte[] IDContext)_, if OscoreStore allows retrieving the information to build the context from based on only the Recipient ID. Overriding _getContext(String uri)_  and _getContextByToken(Token token)_ may be trickier as then we'd lack the information to try to retrieve any info from the OscoreStore about store configuration information (unless it associates them with the Token and URI). Let me think more on this and maybe come up with an updated code idea.

sbernard31: > Thinking further, so what about overriding all the get* methods from HashMapCtxDB in such a way that it initially tries to retrieve the context from the DB

I guess this should be something like this. 

> Overriding getContext(String uri) and getContextByToken(Token token) may be trickier as then we'd lack the information to try to retrieve any info from the OscoreStore

I think there is 2 questions.
- Is it doable ?
- do we want to do it ?

Maybe some of them should only search in OscoreCtxDB and not in OscoreStore  (maybe for security reason or other). 
Currently It's not crystal clear to me why there is all this getContext methods but I think you should better know when it makes sense to search in OscoreStore :slightly_smiling_face:

rikard-sics: > I think there is 2 questions.
> 
>     * Is it doable ?
> 
>     * do we want to do it ?
> 
> 
> Maybe some of them should only search in OscoreCtxDB and not in OscoreStore (maybe for security reason or other). Currently It's not crystal clear to me why there is all this getContext methods but I think you should better know when it makes sense to search in OscoreStore

Yes I think for instance that in the case of _getContextByToken(Token token)_ it can be simply left as it is and not overridden. As I mentioned I believe overriding _OSCoreCtx getContext(byte[] cid)_ and _getContext(byte[] cid, byte[] IDContext)_  should be fine. Again, trickier can be getContext(String uri) as the context needs to be associated with a URI to be found for outgoing requests. I will put together updated code with a proposal related to this.

As a general reflection, the point of this design and overriding these methods is to avoid splitting the addition of contexts and retrieval of contexts into different steps? So practically what is desired is that when attempting to retrieve a context it is automatically derived and added to the database right there, and thus does not need to be added to the DB in earlier code elsewhere?

As far as why there are multiple getContext methods it is partially due to OSCORE functionality and legacy things. For instance it was originally only possible to store a context using the Recipient ID (and not also the ID Context). Also, the OSCORE RFC states that a context should be retrieved using the URI for outgoing requests, the RID+ID Context for incoming requests, and the CoAP Token for incoming/outgoing responses. My long term plan for this is to reduce the amount of these getContext methods and maps in the HashMapCtxDB. For instance in an earlier PR to Californium I managed to remove a map that was used for mapping sequence numbers to Tokens. I also want to rely on the request/response binding in the CoAP exchange to retrieve contexts for responses (as I know the context for the corresponding request). So these things should become simplified in the future.

sbernard31: > So practically what is desired is that when attempting to retrieve a context it is automatically derived and added to the database right there, and thus does not need to be added to the DB in earlier code elsewhere?
 
 Yep I think that's the idea.

rikard-sics: Okay, I have worked on a new suggestion regarding this, please see the draft below. So basically it initially tries to get the context from the parent HashMapCtxDB, and if that fails derive it from material in the OscoreStore. The point about the IllegalArgumentException is that only internal Californium operations should ever use and ID Context (such as the context derivation in Appendix B.2.) Leshan itself should not add a context that is not already in the DB if it has an ID Context (as the ID Context is not part of the parameters in OscoreStore).

One possibly tricky detail is with regards to the URI. I believe having the URI of the other party should not be a problem on the client, as it is the one taking the initiative for connections and knows the URI of the other party. However ideally the information in OscoreStore would need to be updated with the hostname and port of a client after it has communicated with the server for the first time. This is so that the URI can be used to look up the context for outgoing requests from the server.

```
package org.eclipse.leshan.core.californium.oscore.cf;

import org.eclipse.californium.oscore.CoapOSException;
import org.eclipse.californium.oscore.HashMapCtxDB;
import org.eclipse.californium.oscore.OSCoreCtx;
import org.eclipse.californium.oscore.OSCoreCtxDB;
import org.eclipse.californium.oscore.OSException;

/**
 * An {@link OSCoreCtxDB} which store context in memory and is able to derive context from data provided in
 * {@link OscoreStore}
 *
 */
// TODO OSCORE this should be moved in californium.
// TODO OSCORE I don't know if we need to extends HashMapCtxDB or implement OSCoreCtxDB
public class InMemoryOscoreContextDB extends HashMapCtxDB {

    private OscoreStore store;

    public InMemoryOscoreContextDB(OscoreStore oscoreStore) {
        this.store = oscoreStore;
    }

    @Override
    public synchronized OSCoreCtx getContext(byte[] rid, byte[] IDContext) throws CoapOSException {
        OSCoreCtx osCoreCtx = super.getContext(rid, IDContext);

        if (osCoreCtx == null && IDContext != null) {
            throw new IllegalArgumentException("Internal Leshan operations should always use a null ID Context");
        }

        if (osCoreCtx == null) {
            OscoreParameters params = store.getOscoreParameters(rid);
            osCoreCtx = deriveContext(params);
            super.addContext(osCoreCtx);
        }

        return osCoreCtx;
    }

    @Override
    public synchronized OSCoreCtx getContext(byte[] rid) {
        OSCoreCtx osCoreCtx = super.getContext(rid);

        if (osCoreCtx == null) {
            OscoreParameters params = store.getOscoreParameters(rid);
            osCoreCtx = deriveContext(params);
            super.addContext(osCoreCtx);
        }

        return osCoreCtx;

    }

    @Override
    public synchronized OSCoreCtx getContext(String uri) throws OSException {
        OSCoreCtx osCoreCtx = super.getContext(uri);

        if (osCoreCtx == null) {
            OscoreParameters params = store.getOscoreParameters(uri);
            osCoreCtx = deriveContext(params);
            super.addContext(uri, osCoreCtx);
        }

        return osCoreCtx;
    }

    // TODO OSCORE I guess here we should override some method
    // When there is no context in DB we should create/derive it from OscoreStore, right ?

    // TODO OSCORE I guess a method like this will be be useful
    private static OSCoreCtx deriveContext(OscoreParameters oscoreSetting) {
        try {
            OSCoreCtx osCoreCtx = new OSCoreCtx(oscoreSetting.getMasterSecret(), true, oscoreSetting.getAeadAlgorithm(),
                    oscoreSetting.getSenderId(), oscoreSetting.getRecipientId(), oscoreSetting.getHmacAlgorithm(), 32,
                    oscoreSetting.getMasterSalt(), null, 1000);
            osCoreCtx.setContextRederivationEnabled(true);
            return osCoreCtx;
        } catch (OSException e) {
            throw new IllegalStateException("Unable to create OSCoreContext", e);
        }
    }
}
```

```
package org.eclipse.leshan.core.californium.oscore.cf;

/**
 * A store holding OSCORE pre-established inputs which will be used to derive security context.
 * <p>
 * See : https://datatracker.ietf.org/doc/html/rfc8613#section-3.2
 *
 */
// TODO OSCORE this should be moved in californium.
// TODO OSCORE don't know if we want an async API or a more simple sync one.
// Let's start with a sync API for now.
public interface OscoreStore {

    // TODO OSCORE I don't know if this is the right arguments but this is the idea.
    OscoreParameters getOscoreParameters(byte[] recipientID);

    OscoreParameters getOscoreParameters(String uri);
}
```

sbernard31: I  will try to have a working version of this using your code.

At first side, I don't understand how could be used the `OscoreStore.getOscoreParameters(String uri);` API ? :thinking: 

But I guess I totally missed the point about getting context by URI.  (I tried to read this discussion again and have look to OSCORE RFC and still not clear) 

So I will try a first version without this and I will see if this works or not. 
If this does'nt work I guess this will help me to understand why this is needed :crossed_fingers:

sbernard31: (I also read http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Transport-V1_1_1-20190617-A.html#5-5-0-55-OSCORE-based-Security but does not help about context by URI too)

sbernard31: OK I think I get it a little bit  more now.
From a LWM2M perspective, I don't think we need to get the context associated to "a target resource" but more by the foreign peer address but as you said : 

> Although in the current code not the full path but only the hostname + port is considered.

I guess this is more or less the case. 

This is pretty much the same thing than [getIdentity](https://github.com/eclipse/californium/blob/67ad63616de3f071d4f2643889994bf12b0947dc/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/AdvancedPskStore.java#L124-L141) from PskStore ? 

When you initiate an handshake you need to find (PSK + PSK Identity) by foreign peer address.
And when foreign peer initiate handshake you need to get PSK by PSK Identity. 
(PSK identity = user name / PSK = password)

sbernard31: Now I see a bit clearer about URI.
I face a new "missunderstanding" :sweat_smile: 
Until now we are using senderID+recipient ID to identify a foreign peer.  (see [OscoreIdentity](https://github.com/eclipse/leshan/blob/3c23f4baa5faa375cfb393af27797f441855f29c/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/oscore/OscoreIdentity.java#L28))
But with last code you shared,  I see that we can find context (and oscoreParameter) by recipient ID only.
That means that recipient Id is unique and so why not using recipient ID only (without sended ID) as identifier ?

sbernard31: (I changed identifier as recipient Id we will see if we need to change it again)

sbernard31:
@rikard-sics I implemented ideas we talked about at (https://github.com/eclipse/leshan/pull/1212#discussion_r813150717)

So I'm now able to execute  the test below successfully : 
`org.eclipse.leshan.integration.tests.SecurityTest.registered_device_with_oscore_to_server_with_oscore()`

I implement the get by uri at client side, easy as it's static. (See CaliforniumEndpointsManager changes)
I should be able to implement it at server side searching in RegistrationStore (I added a ref to it in LwM2mOscoreStore) but maybe we can do better with californium modification. 

The test start by a register request then the server send a read request.
The register seems to be OK using OSCORE.
As I didn't implement the get by uri at server side, I was expected that the read request failed BUT it is just done without usage of OSCORE (I mean in clear coap) 

I guess in term of security this is not what we want. Any thoughts about that ?

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/oscore/cf/InMemoryOscoreContextDB.java
rikard-sics: > I guess this is more or less the case.
> 
> This is pretty much the same thing than [getIdentity](https://github.com/eclipse/californium/blob/67ad63616de3f071d4f2643889994bf12b0947dc/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/AdvancedPskStore.java#L124-L141) from PskStore ?
> 
> When you initiate an handshake you need to find (PSK + PSK Identity) by foreign peer address. And when foreign peer initiate handshake you need to get PSK by PSK Identity.

True, it is similar to that as the context is associated to the foreign peer address (hostname/IP + port). Just that for OSCORE this kind of lookup is only needed for outgoing requests specifically.

rikard-sics: > Until now we are using senderID+recipient ID to identify a foreign peer. (see [OscoreIdentity](https://github.com/eclipse/leshan/blob/3c23f4baa5faa375cfb393af27797f441855f29c/leshan-server-core/src/main/java/org/eclipse/leshan/server/security/oscore/OscoreIdentity.java#L28))
> But with last code you shared, I see that we can find context (and oscoreParameter) by recipient ID only.
> That means that recipient Id is unique and so why not using recipient ID only (without sended ID) as identifier ?

Strictly speaking OSCORE only guarantees (local) uniqueness of the quartet Master Secret, Master Salt, ID Context, Sender ID. See https://datatracker.ietf.org/doc/html/rfc8613#section-3.3
(I should check also if the LWM2M spec says anything about the uniqueness requirements.)

Practically in my implementation in Californium the RID + ID Context is unique, and you cannot have 2 contexts with the exact same values for those parameters. As Leshan does not consider the ID Context, I think practically enforcing the restriction that no 2 clients can use the same Recipient ID is reasonable. (And either way using the Sender ID as a tie-breaker there would not really help since in incoming requests only the RID (+ID Context) is present.

rikard-sics:
> I implement the get by uri at client side, easy as it's static. (See CaliforniumEndpointsManager changes)
> I should be able to implement it at server side searching in RegistrationStore (I added a ref to it in LwM2mOscoreStore) but maybe we can do better with californium modification.

Sounds very nice, yes we can definitely consider modifications to Californium.



> The test starts by a register request then the server send a read request. The register seems to be OK using OSCORE. As I didn't implement the get by uri at server side, I was expected that the read request failed BUT it is just done without usage of OSCORE (I mean in clear coap successfully)
> 
> I guess in term of security this is not what we want. Any thoughts about that ?

I see, yes as the get by uri is not implemented on the server side it cannot find an OSCORE context to use for protecting the request. But I see your point, basically if a request has the OSCORE option but a context cannot be found, this request should not be sent at all. I will add that to the list of changes to do in Californium. This should be a quick fix.

One possibility for now is to utilize the information in an EndpointContext of an incoming request. See for instance:
https://github.com/eclipse/californium/blob/ee65f934ddb7d40eddd0b0feb91eb8155839fc5a/cf-oscore/src/test/java/org/eclipse/californium/oscore/EndpointContextInfoTest.java#L158
By using that you can see explicitly for a request what Sender ID, Recipient ID, ID Context and URI its associated OSCORE context has (if any). If this information does not exist in the EndpointContext you can tell that the request did not use OSCORE. Perhaps this is good either way for matching the expected security information for a client endpoint name with what it is actually using? I recall something like this was used in the code we had, for instance here:
https://github.com/eclipse/leshan/blob/61b5a12f861b38b416e5c9997019395968b004b7/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/EndpointContextUtil.java#L74

sbernard31:
> I see, yes as the get by uri is not implemented on the server side it cannot find an OSCORE context to use for protecting the request. But I see your point, basically if an outgoing request has the OSCORE option but a context cannot be found, this request should not be sent at all.

I didn't think about this but it makes sense.

My point  was more that the server send a ReadRequest without using OSCORE (this is the expected behavior because I didn't implement get by uri at server side)  **I was expected that the device doesn't answer to it.**

But maybe this behavior can not be done at californium side and so should be done at upper layer (so LWM2M in our case) :thinking: I will try to experiment a way with EndpointContext.

sbernard31:
> One possibility is using an [OSCoreResource](https://github.com/eclipse/californium/blob/master/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreResource.java) as opposed to a normal CoapResource

This will not work with Leshan Client use case, because we create the resource once for all server and some server use OSCORE and some other not. 

My last commit (48b1d60) use `EndpointContext` to check server identity and now the Request request failed :tada: 

I just share an idea for californium API : 
For leshan client we are using 1 Coap Endpoint by foreign peer.
So we are providing a static OscoreStore with only 1 oscore parameter to be sure we talk about the right server.
When using OSCORE, I would like to configure my endpoint in way that OSCORE is always used.
I mean if some try to send none OSCORE request on this endpoint just reject the request.
I guess this could be an option of oscore stack ? 
Do you think this kind of API make sense ?

(I will now try to implement get by uri at server side)

sbernard31:
I Implemented `LwM2mOscoreStore.getOscoreParameters()` (commit 2d9f63d)

But I think returning an OscoreParameters is not the right way because we will re-derive a new context from those parameters where we rather want to reuse an existent one (the one from registration for example)

So maybe we should rather get the recipientId by URI, then call `InMemoryOscoreContextDB.getContext(byte[] recipientID)` ?

I will create a commit about this (would be simple than this explanation :point_up:) and you let me know if this make sense.

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/oscore/cf/InMemoryOscoreContextDB.java
```diff
@@ -46,6 +46,7 @@ public synchronized OSCoreCtx getContext(byte[] rid, byte[] IDContext) throws Co
 
         if (osCoreCtx == null) {
             OscoreParameters params = store.getOscoreParameters(rid);
+            // TODO OSCORE we should handle if param is null ?
             osCoreCtx = deriveContext(params);
             super.addContext(osCoreCtx);
         }
```
sbernard31, line 52: If we have not oscore parameters for this RID.
We should not be able to decrypt the payload and reject the message.

What is the right way to handle this ? return `null` ?

sbernard31:
@rikard-sics, see my last commit (a788103) about : 
> But I think returning an OscoreParameters is not the right way because we will re-derive a new context from those parameters where we rather want to reuse an existent one (the one from registration for example)

rikard-sics:
> But I think returning an OscoreParameters is not the right way because we will re-derive a new context from those parameters where we rather want to reuse an existent one (the one from registration for example)
> 
> So maybe we should rather get the recipientId by URI, then call `InMemoryOscoreContextDB.getContext(byte[] recipientID)` ?

Hmm, maybe I don't get 100% in what situations a new context will be re-derived rather than using the existing one. But I agree that before trying to derive a context we should check the OSCORE DB to make sure a context with a matching RID doesn't already exist there. I will check the later commit.

rikard-sics:
> @rikard-sics, see my last commit ([a788103](https://github.com/eclipse/leshan/commit/a7881033b2780d29e98d0330a6a26aefd8f91489)) about :
> 
> > But I think returning an OscoreParameters is not the right way because we will re-derive a new context from those parameters where we rather want to reuse an existent one (the one from registration for example)

Yes, here I would add the context to the DB associated to the URI. As here the URI would be the "foreign peer address".
https://github.com/eclipse/leshan/blob/a7881033b2780d29e98d0330a6a26aefd8f91489/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/oscore/cf/InMemoryOscoreContextDB.java#L81-L82
Then this would allow the ObjectSecurityLayer to find the correct context for later outgoing requests to that peer.

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/oscore/cf/InMemoryOscoreContextDB.java
rikard-sics: > If we have not oscore parameters for this RID. We should not be able to decrypt the payload and reject the message.
> 
> What is the right way to handle this ? return `null` ?

In the context of OSCORE, if getContext returns null then the ObjectSecurityLayer will understand that no context was found for that peer and then react to that. So for instance for an incoming request it will respond with a 4.01 (Unauthorized) error message with payload string "Security context not found". So if the OSCORE code in Californium is allowed to handle things, that will be the result of a null from getContext().

But for Leshan, perhaps we would rather like a 4.00 (Bad Request) with payload "Unauthorized" as response?

rikard-sics:
> > One possibility is using an [OSCoreResource](https://github.com/eclipse/californium/blob/master/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreResource.java) as opposed to a normal CoapResource
> 
> This will not work with Leshan Client use case, because we create the resource once for all server and some server use OSCORE and some other not.

Yes, I suspected that it would be too course-grained for this use case.

> My last commit ([48b1d60](https://github.com/eclipse/leshan/commit/48b1d602abd9bc9b56b0bedb6be7b0a0fe904dcb)) use `EndpointContext` to check server identity and now the Request request failed tada

Right, very nice.

> I just share an idea for californium API : For leshan client we are using 1 Coap Endpoint by foreign peer. So we are providing a static OscoreStore with only 1 oscore parameter to be sure we talk about the right server. When using OSCORE, I would like to configure my endpoint in way that OSCORE is always used. I mean if some try to send none OSCORE request on this endpoint just reject the request. I guess this could be an option of oscore stack ? Do you think this kind of API make sense ?
> 
> (I will now try to implement get by uri at server side)

Okay I see. So instead of a resource that only accepts OSCORE requests it would be like an endpoint (or an OSCORE CoAP stack building endpoints) that would only accept incoming requests if they use OSCORE. Yes, I think that could be something to consider adding. I can think about how it fits in Californium.

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/oscore/cf/InMemoryOscoreContextDB.java
sbernard31: Ok I will handle this by returning null in Leshan. 

> So for instance for an incoming request it will respond with a 4.01 (Unauthorized) error message with payload string "Security context not found". So if the OSCORE code in Californium is allowed to handle things, that will be the result of a null from getContext(). 

I quickly looked at [OSCORE RFC](https://datatracker.ietf.org/doc/html/rfc8613#section-8.2) and it says : 
_"If the server fails to retrieve a Recipient Context with Recipient ID corresponding to the 'kid' parameter received, the server MAY respond with a 4.01 (Unauthorized) error message."_

So I think it's the right behavior.

> But for Leshan, perhaps we would rather like a 4.00 (Bad Request) with payload "Unauthorized" as response?

Except we see something like this in the LWM2M specification, I would prefer to strictly respect the OSCORE RFC ?

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/oscore/cf/InMemoryOscoreContextDB.java
```diff
@@ -0,0 +1,99 @@
...
+        OSCoreCtx osCoreCtx = super.getContext(uri);
+
+        if (osCoreCtx == null) {
+            byte[] rid = store.getRecipientId(uri);
+            if (rid != null) {
+                osCoreCtx = getContext(rid);
+                // TODO don't know if I should add by uri here ?
+                // super.addContext(uri, osCoreCtx);
+            }
+        }
```
sbernard31, line 84: (from @rikard-sics)
> Here I would add the context to the DB associated to the URI. As here the URI would be the "foreign peer address".
>  Then this would allow the ObjectSecurityLayer to find the correct context for later outgoing requests to that peer.

Thinking a bit more about this :   
In LWM2M the foreign peer address could change (e.g. because devices are behind NAT).
When address changes client should update its registration.

So to handle this
1. either we need to never cache the context by uri  in DB (but that means more call to registration store.
2. Or we cache in for uri and we need to update/clean DB on registration update. (still a bit afraid by this kind of house keeping)

There is another solution and this is something we will maybe need anyway. 
At LWM2M level when we send a request we can tell the "identity" of the destination. This should even be done to be sure that context does not change between the moment where user decide to send the request and the moment where we really send the request.
So when we send the request we could say the expected  OSCORE foreign peer RID.
(This is what we do for DTLS see : https://github.com/eclipse/leshan/blob/leshan-2.0.0-M6/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/request/CoapRequestBuilder.java#L304-L306)

If we do that I guess ObjectSecurityLayer can find RID in EndpointContext and so get by RID and not by URI, right ?

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/oscore/cf/InMemoryOscoreContextDB.java
rikard-sics: > I quickly looked at [OSCORE RFC](https://datatracker.ietf.org/doc/html/rfc8613#section-8.2) and it says : _"If the server fails to retrieve a Recipient Context with Recipient ID corresponding to the 'kid' parameter received, the server MAY respond with a 4.01 (Unauthorized) error message."_
> 
> So I think it's the right behavior.

Yes, this behaviour is based on the OSCORE RFC.

> Except we see something like this in the LWM2M specification, I would prefer to strictly respect the OSCORE RFC ?

Sure, that makes sense to me. Then we stick to the the current way it's done in the Californium OSCORE code, and returning null as you plan should result in the intended behaviour.

sbernard31: (this should be done in 6c2b4c3)

sbernard31:
Not directly linked to this PR.
But what is the "cost" of deriving a new context ?

leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/oscore/cf/InMemoryOscoreContextDB.java
rikard-sics: Right, so you mean that the ObjectSecurityLayer would get the RID from something like the following?
`endpointContext.get(OSCoreEndpointContextInfo.OSCORE_RECIPIENT_ID)`

That could be an addition to Californium, so if that value is set the ObjectSecurityLayer would use the RID there to get the context, rather than the request URI.

sbernard31: I just implemented oscore store at bs server side and this idea :point_up:  would help a lot.

rikard-sics: > I just implemented oscore store at bs server side and this idea point_up would help a lot.

I see, then that is something I can add to the Californium side.

rikard-sics:
> Not directly linked to this PR.
> But what is the "cost" of deriving a new context ?

Let me have a look at this and get back to you.

rikard-sics:
> > Not directly linked to this PR.
> > But what is the "cost" of deriving a new context ?
> 
> Let me have a look at this and get back to you.

Having a look at this this, the main operations taking time when deriving a new context would be running 3 times a HMAC-based Extract-and-Expand Key Derivation Function (HKDF). (For generating the Sender Key, Recipient Key and Common IV).
https://github.com/eclipse/californium/blob/05a828efded564593f15ad042261a07380a00aa8/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java#L810

sbernard31:
@rikard-sics  thx for you last answer :pray: 

I squashed all this commit and integrated this in `oscore` branch (commit  7fff92d)

Issue #1192, opened 2022-01-04
Unable to Send and receive multiple values with different timestamps of the same resource


Michal-Wadowski:
It seems to be no way to Send multiple data by client from one resource with different timestamps. Also I don't see possibility to receive this kind of data by server (by ```SendListener```).

However I can see ```LwM2mDecoder.decodeTimestampedData()``` and ```LwM2mEncoder.encodeTimestampedData()``` which implementations allow to send timestamped data.

I created POC code locally that uses already implemented ```decodeTimestampedData()``` and ```encodeTimestampedData()```, so technically it's possible to do that, but it needs more investigation.

Michal-Wadowski:
I tried to update ReadResponse to allow send list of timestamped nodes insead of one node but it turned out it requires a lot of changes in related classes. Finally, even if I resolved conflicts (and break a lot of code) I end up on firing LwM2mEncoder.encodeTimestampedData() method with data type ```Map<LwM2mPath, List<TimestampedLwM2mNode>>``` which is not supported. There is supported only encode nodes with different timestamps but one path.

sbernard31:
Send timestamped value is relative to `stored notification` feature. (and so relative to observe)
There is an issue about this : https://github.com/eclipse/leshan/issues/596. (following links you should get more details about this)
And today this is not implemented.

> However I can see LwM2mDecoder.decodeTimestampedData() and LwM2mEncoder.encodeTimestampedData() which implementations allow to send timestamped data.

Yep  `stored notification`  is supported at Server side and not a client side but some part of the work was done at encoder side. 
This helps to write some tests : [ObserveTimeStampTest.java](https://github.com/eclipse/leshan/blob/master/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/observe/ObserveTimeStampTest.java)

About implementing `stored notification` feature at client  this could be a bit complication as this is part of Observe feature which is not really in the spirit of CoAP Observe. (I don't know if this is doable with current Californium API)

> Also I don't see possibility to receive this kind of data by server (by SendListener).

You talk about Send Operation, right ? 
Reading the specification I don't understand this is allowed, I mean sending timestamped value with Send Operation. but maybe I missed something. Do you find anything about that in the specification ? :thinking:

Michal-Wadowski:
Yes, my main purpose is to allow to Send multiple timestamped values for some path. For example device can collect some of data and send this with historical values. I'm not sure if it's allowed with specification either.

[...]

PR #1217, opened 2022-03-08
Multi-timestamp Send implementation

Michal-Wadowski:
The solution for #1192


leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodesImpl.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+        return result;
+    }
+
+    @Override
+    public Set<Long> getTimestamps() {
+        return timestampedPathNodesMap.keySet();
+    }
+
+    public void put(Long timestamp, LwM2mPath path, LwM2mNode node) {
```
sbernard31, line 76: TimestampedLwM2mNodesImpl seems not to be immutable. 
See https://github.com/eclipse/leshan/issues/1192#issuecomment-1058059555


If possible I would prefer to make it immutable.

sbernard31, line 76: TimestampedLwM2mNodesImpl seems not to be immutable. 
See https://github.com/eclipse/leshan/issues/1192#issuecomment-1058059555


If possible I would prefer to make it immutable.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/send/SendTimestampedTest.java
```diff
@@ -0,0 +1,181 @@
...
+            Map<Long, LwM2mNode> timestampedNodes = getExampleTimestampedNodes();
+
+            TimestampedLwM2mNodesImpl data = new TimestampedLwM2mNodesImpl();
+            for (Map.Entry<Long, LwM2mNode> entry : timestampedNodes.entrySet()) {
+                data.put(entry.getKey(), getExamplePath(), entry.getValue());
+            }
+
+            return data;
+        }
+
```
sbernard31, line 159: I'm not sure I get the point here because it seems that the server will ignore the `byte[] content` ?
Should we rather hack the client instead of server ? 
I mean the purpose of this PR is to add the feature at server side without the client side.
So I can understand that we hack the client to make it send a timestampted payload but if we hack the server I'm not sure what we are testing here ? :thinking:

sbernard31, line 159: I'm not sure I get the point here because it seems that the server will ignore the `byte[] content` ?
Should we rather hack the client instead of server ? 
I mean the purpose of this PR is to add the feature at server side without the client side.
So I can understand that we hack the client to make it send a timestampted payload but if we hack the server I'm not sure what we are testing here ? :thinking:

sbernard31 (commented):
A first review without looking at small details.

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Get timestamp grouped map of {@link LwM2mPath}-{@link LwM2mNode} map.
+     */
+    Map<Long, Map<LwM2mPath, LwM2mNode>> getTimestampedNodes();
+
+    /**
+     * Get all collected nodes represented as {@link LwM2mPath}-{@link LwM2mNode} map.
+     */
+    Map<LwM2mPath, LwM2mNode> getNodes();
```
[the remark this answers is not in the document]
Michal-Wadowski: The TimestampedLwM2mNodes is collection of nodes with information about path and timestamp. This methods just returns nodes (path map). Should be called getAllNodes()? I'm not quite understand the problem.

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodesImpl.java
```diff
@@ -0,0 +1,131 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+
+/**
+ * The default implementation of {@link TimestampedLwM2mNodes}
+ */
+public class TimestampedLwM2mNodesImpl implements TimestampedLwM2mNodes {
```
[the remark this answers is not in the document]
Michal-Wadowski: The core functionality is extracted as interface and - as you ask before - there is read only. The concrete implementation have some options that allow to create and manipulate instance (add elements, create instance etc). There is functional separation where read only interface should be used in most cases and concrete implementation to create one.

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodesImpl.java
Michal-Wadowski: Immutable is TimestampedLwM2mNodes interface. Concrete implementation should not be exposed outside.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/send/SendTimestampedTest.java
Michal-Wadowski: It was the starting point for searching solution, checking what happen if we get already decoded nodes in SendResource.handlePOST(). Where is no direct tests for SendResource and SendHandler so I just used integration tests to check this layer. I didn't hack the client because there was no real implementation of encoder/decoder yet.
But for now this test seems obsolete and it looks like it's actually better to modify client side indeed.
I'll check if I can provide better solution.

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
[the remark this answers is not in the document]
sbernard31: My concern is that I'm not sure what means "all nodes".

Eg. : 
```
t1 => path1 => nodeValue1a
t1 => path2 => nodeValue2
t2 => path1 => nodeValue1b
t3 => path1 => nodeValue1c

```
In this case,  with a map from path to node, we will have only 1 value for path1 or I missed something ? so this is not really all node ? or ?

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodesImpl.java
sbernard31: Not a big deal but I think we don't understand Immutable in a same way. 

I understand that immutable means that an object can not be modified once it is created.
So this is not really applicable to an interface because an interface can not be created.
And in this case, the object behind the interface can be modified  by using cast. 

Anyway, you can let this point aside if you want :-)

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/send/SendTimestampedTest.java
Michal-Wadowski: In the meantime I modified test to hack client side rather than the server side

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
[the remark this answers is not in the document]
Michal-Wadowski: My intention in this case is provide way to get nodes as usual, ignoring timestamps (if you don't care about timestamps). In other case you can use getTimestampedNodes(), getNodesForTimestamp() to be time-aware. What name should I use to improve readability or give more context? And maybe javadoc is not enough decriptive?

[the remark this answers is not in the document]
sbernard31: Ok I get it.
So I guess if you don't care about timestamped you want to get more recent value right ? So maybe this should be added in javadoc and/or the method is renamed in getMostRecentNodes() (or something like this) ? 

Bonus quesiton : 
Is it possible to have timestamped value mixed with "not timestamped" value ?  
If it is possible, null timestamped value is more or less recent ?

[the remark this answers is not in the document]
Michal-Wadowski: I want to avoid the "most recent" phrase because it suggest to gets only ```"t3 => path1 => nodeValue1c"``` value from example above (because it's really most recent). My intention is to ignore timestamp segregation. Of course in example above we have conflicts and we have to choose one of "nodeValue1a", "nodeValue1b", "nodeValue1c" and the last one seems to be most appropriate indeed.

>  Is it possible to have timestamped value mixed with "not timestamped" value ?

If client send this kind of data and it's technically valid, the server should be able to handle it. So the answer is "yes"

> If it is possible, null timestamped value is more or less recent ?

It's undefined by definition, and until you use methods ```getTimestampedNodes()```, ```getNodesForTimestamp()``` and ```getTimestamps()``` it doesn't really matter (the consumer implementation have ability to manage the order). I had to provide some order for technical reasons (default TreeMap doesn't support null keys), so I recreated Comparator.nullsFirst() equivalent from java 8 (it's not copy paste!).

Maybe I should add some tests to valid order in cases pointed above?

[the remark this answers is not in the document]
sbernard31: > I want to avoid the "most recent" phrase because it suggest to gets only "t3 => path1 => nodeValue1c" value from example above (because it's really most recent). My intention is to ignore timestamp segregation. Of course in example above we have conflicts and we have to choose one of "nodeValue1a", "nodeValue1b", "nodeValue1c" and the last one seems to be most appropriate indeed.

I understand but this seems more confusing to me to not explained how returned nodes are selected.
Even if the wording "most recent" is not crystal clear it gives at least some hints about the behavior which is only most recent value of each node is selected. 

Anyway, not a big deal if the method name does not contains the "most recent" idea but I think at least the javadoc should clearly define the behavior.

> It's undefined by definition, and until you use methods getTimestampedNodes(), getNodesForTimestamp() and getTimestamps() it doesn't really matter (the consumer implementation have ability to manage the order). I had to provide some order for technical reasons (default TreeMap doesn't support null keys), so I recreated Comparator.nullsFirst() equivalent from java 8 (it's not copy paste!).

I'm a bit confused by the idea because I don't know how this could be undefined.
At implementation level we must decide which node we keep in this path=>node map when we have something like : 
E.g. : 
```
null => path 1 => nodeValue1a
t1 => path 1 => nodeValue1b
t2 => path 1 => nodeValue1c
```
I guess `null` can be consider as now and so probably the most recent one and timestamp is for past value. (future value does not really make sense :thinking: )

> Maybe I should add some tests to valid order in cases pointed above?

It sounds a good idea.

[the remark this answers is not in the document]
Michal-Wadowski: > I'm a bit confused by the idea because I don't know how this could be undefined.
I mean this order is nowhere directly described, so we have to establish how order null timestamp.
> I guess null can be consider as now and so probably the most recent one and timestamp is for past value.
Ok for me. I've updated implementation to take null timestamp as the latest.

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodesImpl.java
Michal-Wadowski: There are many ways to achieve the goal. I changed the implementation to make TimestampedLwM2mNodes immutable (without interface).

sbernard31:
@Michal-Wadowski  thx for last changes.
I reviewed it more in details and fix some littles issue/details at #1218.
Could you review if this is OK for you ?

sbernard31:
This is included in #1218.
So we can close this.

Issue #1089, opened 2021-09-06
Add support of timestamped value for Observe-Composite.


sbernard31:
This feature is needed to support clients which are using `Notification Storing When Disabled or Offline` feature. 

Currently,  
Leshan client does not support this feature (see #596)  
Leshan server is able to handle "classic" observe response from client using this feature.  
Leshan server is not able to handle "composite" observe response from client using this feature.

JaroslawLegierski:
I prepared following [PoC](https://github.com/eclipse-leshan/leshan/compare/master...JaroslawLegierski:leshan:opl/observe_composite_support_for_ts_values) with very limited support for Observe-Composite (only in leshan client) with timestamped data 
(mainly for better understunding observe-composite concept in Leshan)

What do you think about it - is this the correct direction? 
Currently we have 2 different methods in encoder:

-  encode - for classical nodes
-  encodeTimestampedNode - for timestamped nodes

Do we need 3'rd encoder for encoding mixed nodes (timestamped with classical into one notification) ?

sbernard31:
I looked at it and don't think this is the right direction.

1. There is no need to code new encoder/decoder `TimestampedMultiNodeEncoder`/`TimestampedMultiNodeDecoder` should be enough
2. Maybe `TimestampedMultiNodeDecoder.decodeTimestampedNodes(byte[], LwM2mModel)` should be adapted as expected (https://github.com/eclipse-leshan/leshan/pull/1218#discussion_r824550092)
3. No need to add client code to test it, just create a dedicated integration tests like `ObserveTimeStampTest`
4. (not directly linked I guess)  I didn't get why you add `.getId(`) to `LwM2mRoot()`  ?

HTH :slightly_smiling_face:

sbernard31:
> OK then I will focus on adapting TimestampedMultiNodeDecoder.decodeTimestampedNodes(byte[], LwM2mModel)

Let me know, if there is unclear point. 

> As I understand it, I should start by writing dedicated interation test instead of modifying the client/leshan-client-demo code ?

Yep I think this will be easier and adding a integration tests is a good idea anyway.

JaroslawLegierski:
Please find [here ](https://github.com/eclipse-leshan/leshan/compare/master...JaroslawLegierski:leshan:opl/observe_composite_timestamped_data_support_for_server) PoC of Observe Composite  with ts values support on server side.

The following points are not clear for me:

1) Currently leshan SENML_JSON encoder can encode following  records:

```json
[
[...]

PR #1218, opened 2022-03-10
Add Timestamped Nodes support to Send Operation at Server side.

sbernard31:
This PR includes #1217 with some fixes/cleaning.

This aims to implement #1192 at server side.


sbernard31:
(I stopped for today but maybe tomorrow I will look at this a bit more)

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
```diff
@@ -0,0 +1,164 @@
...
+                Map<LwM2mPath, LwM2mNode> pathNodeMap = entry.getValue();
+
+                for (Map.Entry<LwM2mPath, LwM2mNode> pathNodeEntry : pathNodeMap.entrySet()) {
+                    LwM2mPath path = pathNodeEntry.getKey();
+                    LwM2mNode node = pathNodeEntry.getValue();
+                    put(timestamp, path, node);
+                }
+            }
+            return this;
+        }
```
sbernard31, line 142: We should maybe add some check here and raise exception in case of duplicate data ?
(this exception should be catched in decoder)

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/LwM2mDecoder.java
```diff
@@ -97,6 +98,19 @@
...
+     *
+     * @param content the content
+     * @param format the content format
+     * @param model the collection of supported object models
+     * @return the decoded timestamped nodes represented by {@link TimestampedLwM2mNodes}
+     * @throws CodecException if content is malformed.
+     */
+    TimestampedLwM2mNodes decodeTimestampedNodes(byte[] content, ContentFormat format, LwM2mModel model)
+            throws CodecException;
+
```
sbernard31, line 113: Compared to decodeNodes API:
```java
Map<LwM2mPath, LwM2mNode> decodeNodes(
                   byte[] content,
                   ContentFormat format,
                   List<LwM2mPath> paths,
                   LwM2mModel model)  throws CodecException;
```
We removed `List<LwM2mPath> paths`but I guess this will be necessary again when we will implement  https://github.com/eclipse/leshan/issues/1089 :thinking:

Michal-Wadowski: What if the ```List<LwM2mPath> paths``` parameter would be implemented as filter of result? It could simplify decoder implementation a little bit.

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
sbernard31: I'm not sure I was clear.
My point is when you put new data in the `Builer.timestampedPathNodesMap` structure,  overrinding already existing data seems to be not valid to me ?

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/LwM2mDecoder.java
sbernard31: The idea is exaclly the same as decodeNodes.

For observe we need the list of paths for send we don't. 

But maybe we add it later when/if we implement timestamp for composite observe.

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
Michal-Wadowski: I'm understand and I'm still think that should not be forbidden, it doesn't sound bad for me.

sbernard31: I wanted to be sure before to go in deeper explanation. :slightly_smiling_face:

Let me explain why I think that this should **maybe** not be allowed.

I feel there two usage about this: 
1. timestampedNodes created in a decoder
2. timestampedNode created at client side by the library itself.

In first case, I consider if that duplicate value means payload we try to decode in invalid and so I think we should raise an exception. 

In second case, I'm not sure because we don't implement anything about this but I also bet that trying to store 2 value for same path and same timestamp looks like a bug. 

Did I missed something ? 

In case, we really want to allow to override value in timestampedNodes, maybe we can adapt the API (no more fluent builder) and return erased nodes (when duplicate) but I'm not sure if this worth it  because I mainly see use case where we don't want duplicate :thinking:

Michal-Wadowski: Ok, the argument about validation in decoder is ok for me - we should not allow to decode invalid data. But my intention for TimestampedLwM2mNodes.Builder was not only decoder usage, so maybe we should allow to add validation optionally e.g. ```TimestampedLwM2mNodes.builder(boolean forbidDuplicates)``` or ```builder.forbidDuplicates(true)``` (it's only proposition).

> maybe we can adapt the API (no more fluent builder)

The reason to create builder is to make TimestampedLwM2mNodes immutable. You can technically create many constructors that allow to create TimestampedLwM2mNodes in many ways, but it's much easier using builder.

sbernard31: > The reason to create builder is to make TimestampedLwM2mNodes immutable. You can technically create many constructors that allow to create TimestampedLwM2mNodes in many ways, but it's much easier using builder.

I mean we keep the builder but not more fluent anymore (as we use returned value for something else than returning the builder itself)

Michal-Wadowski: I don't want to spoil builder fluent. Maybe better option would be add getter to get/check the value.

sbernard31: > Maybe better option would be add getter to get/check the value

I'm not sure to see how ? could you detailed this ?

sbernard31: (I just try to work on this, I will propose a solution soon)

sbernard31:
@Michal-Wadowski I added 2 more commits (not related to validation): 
- Fix LwM2mNodeDecoderTest.senml_multiple_timestamped_nodes
- Make TimestampedLwM2mNodes really immutable

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
sbernard31: See last commit about validation.
Is there something like this you have in mind ?

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/LwM2mDecoder.java
Michal-Wadowski: I looked at builder update and the solution is great for me :+1:

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
Michal-Wadowski: I looked at builder update and the solution is great for me :+1:

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
```diff
@@ -1277,24 +1277,12 @@ public void senml_multiple_timestamped_nodes() throws CodecException {
...
         // then
-        Map<Long, Map<LwM2mPath, LwM2mNode>> expectedResult = new HashMap<>();
-        Map<LwM2mPath, LwM2mNode> first = new HashMap<>();
-        first.put(new LwM2mPath("/3/0/7/0"), LwM2mResourceInstance.newIntegerInstance(0, 3800));
-        expectedResult.put(268600000L, first);
+        TimestampedLwM2mNodes.Builder expectedResult = new TimestampedLwM2mNodes.Builder();
+        expectedResult.put(268600000L, new LwM2mPath("/3/0/7/0"), LwM2mResourceInstance.newIntegerInstance(0, 3800));
+        expectedResult.put(268600001L, new LwM2mPath("/4/0/0"), LwM2mSingleResource.newIntegerResource(0, 1));
+        expectedResult.put(268600002L, new LwM2mPath("/4/0/1"), LwM2mSingleResource.newIntegerResource(1, 2));
+        expectedResult.put(268600003L, new LwM2mPath("/4/0/2"), LwM2mSingleResource.newIntegerResource(2, 3));
```
Michal-Wadowski, line 1284: You can use fluency of builder eg:
```java
   TimestampedLwM2mNodes expectedResult = new TimestampedLwM2mNodes.Builder()
      .put(268600000L, new LwM2mPath("/3/0/7/0"), LwM2mResourceInstance.newIntegerInstance(0, 3800))
      .put(268600001L, new LwM2mPath("/4/0/0"), LwM2mSingleResource.newIntegerResource(0, 1))
      .put(268600002L, new LwM2mPath("/4/0/1"), LwM2mSingleResource.newIntegerResource(1, 2))
      .put(268600003L, new LwM2mPath("/4/0/2"), LwM2mSingleResource.newIntegerResource(2, 3))
      .build()
```

leshan-core/src/test/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodesTest.java
```diff
@@ -103,11 +103,11 @@ public void should_getNodesForTimestamp_returns_ascending_ordered_nodes() {
         TimestampedLwM2mNodes tsNodes = getExampleTimestampedLwM2mNodes();
 
         // when
-        Map<Long, Map<LwM2mPath, LwM2mNode>> tsNodesMap = tsNodes.getTimestampedNodes();
```
Michal-Wadowski: After change usage of method ```getTimestampedNodes()``` you should rename the test to e.g. ```should_getTimestamps_returns_in_ascending_order```

leshan-core/src/test/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodesTest.java
```diff
@@ -118,11 +118,11 @@ public void should_null_timestamp_be_considered_as_latest_for_getNodesForTimesta
         TimestampedLwM2mNodes tsNodes = getExampleMixedTimestampLwM2mNodes();
 
         // when
-        Map<Long, Map<LwM2mPath, LwM2mNode>> tsNodesMap = tsNodes.getTimestampedNodes();
```
Michal-Wadowski: After change usage of method ```getTimestampedNodes()``` you should rename the test to e.g. ```should_null_timestamp_be_considered_as_latest_for_getTimestamps```

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
```diff
@@ -1268,30 +1268,30 @@ public void senml_multiple_timestamped_nodes() throws CodecException {
         StringBuilder b = new StringBuilder();
         b.append("[{\"bn\":\"/4/0/\",\"bt\":268600000,\"n\":\"0\",\"v\":1,\"t\":1},");
         b.append("{\"n\":\"1\",\"v\":2,\"t\":2},");
-        b.append("{\"n\":\"1\",\"v\":2,\"t\":3},");
+        b.append("{\"n\":\"2\",\"v\":3,\"t\":3},");
```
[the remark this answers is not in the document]
sbernard31: I guess the main reason is that I refer to LwM2mPath just below ("/4/0/0", "/4/0/1","/4/0/2")
I must confess that I maybe missed what you try to test here.  :grimacing: 

I saw several inconsistencies  : 
- all nodes are resource instance (LwM2mResourceInstance) where some path targeted single resource (e.g /4/0/0).
- some resource id  ( LwM2mResourceInstance.newIntegerInstance(**[0]**, 2))) doesn't match the path (/4/0/**[1]**) or/and the senML path.
- some SenML value (e.g : "v":3800) does not match node value (e.g 3000)
- despite this kind of inconsistency the final equals check is :heavy_check_mark: 

As this looks like bugs to me, I tried to fix it in something which sounds consistent to me, but I guess I maybe missed the point :grimacing: 

So to fix the test I should use "your" senML payload as "source" right ?

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
```diff
@@ -59,21 +55,20 @@ private TimestampedLwM2mNodes(Map<Long, Map<LwM2mPath, LwM2mNode>> timestampedPa
...
     public Set<Long> getTimestamps() {
-        return timestampedPathNodesMap.keySet();
+        return Collections.unmodifiableSet(timestampedPathNodesMap.keySet());
     }
 
     @Override
     public String toString() {
-        return String.format("TimestampedLwM2mNodes [timestampedPathNodesMap=%s, timestampedNodes=%s]",
-                timestampedPathNodesMap);
+        return String.format("TimestampedLwM2mNodes [timestampedNodes=%s]", timestampedPathNodesMap);
```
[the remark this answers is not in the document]
sbernard31: I think maybe I just keep : TimestampedLwM2mNodes [%s] ?

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
sbernard31: By the past I have some formatting issue with this kind of notation, so I get out the habit to use it.
But I just test it and it sounds good this time, so I will change it.

leshan-core/src/main/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodes.java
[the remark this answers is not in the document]
Michal-Wadowski: For one parameter it sounds ok

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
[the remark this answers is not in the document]
Michal-Wadowski: The reason that it looks like that is I created POC first, and while creating PR I didn't clean this test enough. I tried put too much ideas to one test.

You fixed the most of the inconsistency I created but it was missed one most important thing - multiple values for one path but different timestamp. So yes, if you restore original payload it would be ok.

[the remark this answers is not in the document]
sbernard31: > The reason that it looks like that is I created POC first, and while creating PR I didn't clean this test enough. I tried put too much ideas to one test.

No problem, mistakes happens :slightly_smiling_face: 

> So yes, if you restore original payload it would be ok.

OK I will change it in that way.  :+1:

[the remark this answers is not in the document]
sbernard31: I will keep same path (n=1) for different timestamp but change the value (v:2 & v:3), I feel there is little more chance to detect issue.
```diff
b.append("{\"n\":\"1\",\"v\":2,\"t\":2},");
- b.append("{\"n\":\"1\",\"v\":2,\"t\":3},");
+ b.append("{\"n\":\"1\",\"v\":3,\"t\":3},");
```

sbernard31:
@Michal-Wadowski, I do some modification to fix issue you raised and I also add some path/node consistency check.

From my point of view we are good to integrate this in `master` just let me know if you find more issue and/or if something is missing.  :)

Michal-Wadowski:
> @Michal-Wadowski, I do some modification to fix issue you raised and I also add some path/node consistency check.
> 
> From my point of view we are good to integrate this in `master` just let me know if you find more issue and/or if something is missing. :)

I checked the commits and all is OK for me except in the test ```should_raise_exception_id_path_does_not_match_node``` - there is unnecessary empty line at the end.

Good job, I'm very happy we worked out the functionality together :smile:

sbernard31:
>  the test should_raise_exception_id_path_does_not_match_node - there is unnecessary empty line at the end.

I fixed it. 

> Good job, I'm very happy we worked out the solution together smile

Thx to you. I very appreciate your contributions :pray:

sbernard31:
This is now integrated in `master` (commit 633e3e2)

PR #1226, opened 2022-03-24
`oscore` branch cleaning 

sbernard31:
This PR aims to do a check about all TODO OSCORE + re-review the code to polish the current `oscore` branch.

(This is part of the plan : https://github.com/eclipse/leshan/issues/920#issuecomment-1074919691)

This is work in progress.


rikard-sics:
> @rikard-sics, I will try to push small commit regulary on this PR. This way you can follow/review my work day to day.

Okay, sounds great. Then I will keep track of commits being added here. Let us discuss if there is anything unclear, or about additional things to add in Californium if any more come up.

leshan-server-core/src/main/java/org/eclipse/leshan/server/security/oscore/OscoreSetting.java
```diff
@@ -18,27 +18,43 @@
...
+        // TODO OSCORE add maximum length check
         this.senderId = senderId;
         this.recipientId = recipientId;
         this.masterSecret = masterSecret;
-        this.aeadAlgorithm = aeadAlgorithm;
-        this.hmacAlgorithm = hmacAlgorithm;
-        this.masterSalt = masterSalt;
+        this.aeadAlgorithm = aeadAlgorithm == null ? AeadAlgorithm.AES_CCM_16_64_128 : aeadAlgorithm;
+        this.hkdfAlgorithm = hkdfAlgorithm == null ? HkdfAlgorithm.HKDF_HMAC_SHA_256 : hkdfAlgorithm;
+        this.masterSalt = masterSalt == null ? new byte[0] : masterSalt;
```
sbernard31, line 57: @rikard-sics 

Is there some validation we can do here ? 

I see at https://datatracker.ietf.org/doc/html/rfc8613#section-3.1
> Sender ID.  Byte string used to identify the Sender Context, to
      derive AEAD keys and Common IV, and to contribute to the
      uniqueness of AEAD nonces.  Maximum length is determined by the
      AEAD Algorithm.
      
and 

> Recipient ID.  Byte string used to identify the Recipient Context,
      to derive AEAD keys and Common IV, and to contribute to the
      uniqueness of AEAD nonces.  Maximum length is determined by the
      AEAD Algorithm.
      
I don't know which data can be used from https://datatracker.ietf.org/doc/html/rfc8152#section-10 to define max length.

Is there limitation for `masterSalt` and/or `masterKey`, I don't think so but I prefer to be sure :thinking:

rikard-sics: In the OSCORE code in Californium there are calculations related to this. See the following:
https://github.com/eclipse/californium/blob/0459aa8c21eb59137cb6460fe352d0b73a1173d4/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java#L686-L688
https://github.com/eclipse/californium/blob/0459aa8c21eb59137cb6460fe352d0b73a1173d4/cf-oscore/src/main/java/org/eclipse/californium/cose/EncryptCommon.java#L217

See also the following sections in the OSCORE RFC:
https://datatracker.ietf.org/doc/html/rfc8613#section-5.2
https://datatracker.ietf.org/doc/html/rfc8613#section-3.3

> The maximum length of Sender ID in
> bytes equals the length of the AEAD nonce minus 6
> For AES-CCM-16-64-128 the maximum length of Sender ID is 7 bytes.

As for the master secret and master salt there is no maximum length enforced by OSCORE.

sbernard31: Great ! thx for the answers :pray: 
I added this kind of validation in the last commit 08690ee

leshan-core/src/main/java/org/eclipse/leshan/core/oscore/AeadAlgorithm.java
```diff
@@ -10,17 +10,17 @@
...
-    public static final AeadAlgorithm AES_CCM_16_64_256 = new AeadAlgorithm("AES-CCM-16-64-256", 11); //
-    public static final AeadAlgorithm AES_CCM_64_64_128 = new AeadAlgorithm("AES-CCM-64-64-128", 12); //
-    public static final AeadAlgorithm AES_CCM_64_64_256 = new AeadAlgorithm("AES-CCM-64-64-256", 13); //
-    public static final AeadAlgorithm AES_CCM_16_128_128 = new AeadAlgorithm("AES-CCM-16-128-128", 30); //
-    public static final AeadAlgorithm AES_CCM_16_128_256 = new AeadAlgorithm("AES-CCM-16-128-256", 31); //
-    public static final AeadAlgorithm AES_CCM_64_128_128 = new AeadAlgorithm("AES-CCM-64-128-128", 32); //
-    public static final AeadAlgorithm AES_CCM_64_128_256 = new AeadAlgorithm("AES-CCM-64-128-256", 33); //
+    public static final AeadAlgorithm AES_GCM_128 = new AeadAlgorithm("A128GCM", 1, 12); //
+    public static final AeadAlgorithm AES_GCM_192 = new AeadAlgorithm("A192GCM", 2, 12); //
+    public static final AeadAlgorithm AES_GCM_256 = new AeadAlgorithm("A256GCM", 3, 12); //
```
sbernard31, line 15: About nonce size for GCM, I'm not 100% sure.
The [rfc8152#section-10.1](https://datatracker.ietf.org/doc/html/rfc8152#section-10.1) says : 
>  The GCM mode is parameterized by the size of the authentication tag
   and the size of the nonce.  This document fixes the size of the nonce
   at 96 bits.  The size of the authentication tag is limited to a small
   set of values.  For this document however, the size of the
   authentication tag is fixed at 128 bits.
   
@rikard-sics , that sounds good to you ?

leshan-core/src/main/java/org/eclipse/leshan/core/oscore/OscoreSetting.java
```diff
@@ -45,13 +45,34 @@ public OscoreSetting(byte[] senderId, byte[] recipientId, byte[] masterSecret, A
...
+        }
+        if (this.recipientId.length > maxLength) {
+            throw new IllegalArgumentException(
+                    String.format("Invalid Recipient ID (%s) : max length for % algorithm is %s",
+                            Hex.encodeHexString(this.recipientId), this.aeadAlgorithm, maxLength));
+        }
+        // Validate master key.
+        if (this.masterSecret.length == 0) {
+            throw new IllegalArgumentException("Invalid Master Secret : can not be an empty String");
+        }
```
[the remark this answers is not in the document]
rikard-sics: True, at least in the current code in the Californium implementation it is not allowed.

As for the OSCORE RFC, it says the following:
https://datatracker.ietf.org/doc/html/rfc8613#section-12.3
> Therefore, the main requirement for the OSCORE Master Secret, in
> addition to being secret, is that it have a good amount of randomness.
> [...] For pre-shared key deployments and key transport solutions such
> as [RFC8613](https://datatracker.ietf.org/doc/html/rfc8613), the Master Secret can be generated offline using a good
> random number generator.  Randomness requirements for security are
> described in [RFC4086](https://datatracker.ietf.org/doc/html/rfc4086).

I believe the OSCORE RFC never explicitly forbids using a 0 length Master Secret, but it seems a reasonable restriction also considering the section I quoted.

rikard-sics:
> Which aead and hkdf algorithm are supported by current implementation ?
> Which ones do we want for the minimal viable feature ?
> Only the default one ?

Currently the following AEAD algorithms are supported:

- AES_CCM_16_64_128
- AES_CCM_16_128_128
- AES_CCM_64_64_128
- AES_CCM_64_128_128

And for HKDF:

- HKDF_HMAC_SHA_256
- HKDF_HMAC_SHA_512

I think for a minimal viable feature only supporting the defaults are sufficient.

leshan-core/src/main/java/org/eclipse/leshan/core/oscore/AeadAlgorithm.java
rikard-sics: The underlying COSE library that we use (but modified) uses a nonce size of 96 bits for GCM. This can be seen here:
https://github.com/cose-wg/COSE-JAVA/blob/e461e5e9f87e18f88ff33d5344c7512b4c3b4bdd/src/main/java/COSE/EncryptCommon.java#L30

So 96 bits for the nonce sounds good to me.

sbernard31:
I did most of what I wanted about cleaning.

The last thing I plan to add is to deactivate OSCORE by default and add an option to activate it (as this is still an experimental feature)

Then, I will do the of possible modification of californium API.

sbernard31:
I fall on this  OSCORE code :
https://github.com/eclipse/leshan/blob/7fff92d8b8a23e9cd088f37bce310468c79711a1/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/request/LwM2mResponseBuilder.java#L255-L267

@rikard-sics, I feel this should not be in Leshan code. Could you help me to better understand this ?

rikard-sics:
> @rikard-sics, I feel this should not be in Leshan code. Could you help me to better understand this ?

I need to dig a bit deeper into this to recall the exact problem.

From my memory right now the issue was that the coapRequest object here sometimes lacked the Token (even though it was sent over the wire with a Token), this in turn caused issues with later OSCORE processing. It is not clear to me why this was happening.

One simple thing to test would be to simply remove these lines and see if observations with OSCORE still works. I will try to have a look at this today or tomorrow to understand how things can be fixed (if still an issue) in the OSCORE code in Californium.

rikard-sics:
> I did most of what I wanted about cleaning.
> 
> The last thing I plan to add is to deactivate OSCORE by default and add an option to activate it (as this is still an experimental feature)
> 
> Then, I will do the list of possible modification of californium API.

Sounds great. Yeah, having it enabled with a flag for now can make sense.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java
```diff
@@ -271,9 +271,14 @@ public synchronized ServerIdentity createEndpoint(ServerInfo serverInfo, boolean
                         "Unable to create endpoint for %s using OSCORE : Unable to decode OSCORE AEAD from %s.",
                         serverInfo, serverInfo.oscoreSetting), e);
             }
+
+            // TODO OSCORE kind of hack because californium doesn't support an empty byte[] array for salt ?
+            byte[] masterSalt = serverInfo.oscoreSetting.getMasterSalt().length == 0 ? null
+                    : serverInfo.oscoreSetting.getMasterSalt();
```
sbernard31, line 277: I currently testing to update my previous OSCORE demo PRs with our last modification.

And I just saw that current californium API does not like if masterSalt is an empty byte array.
I guess this comes from that codes from : [org.eclipse.californium.oscore.OSCoreCtx.OSCoreCtx` line 261](https://github.com/eclipse/californium/blob/75e72b207eb927e914ac0062faac8009b7e4c55c/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java#L261-L266)
```java
if (master_salt == null) {
	// Default value. Automatically initialized with 0-es.
	this.common_master_salt = new byte[this.kdf.getKeySize() / Byte.SIZE];
} else {
	this.common_master_salt = master_salt.clone();
}
```
But reading the OSCORE [RFC8613](https://datatracker.ietf.org/doc/html/rfc8613#section-3.2) : 
>  Master Salt
>      *  Default is the empty byte string

Is it a kind of bug or is it the expected behavoir ?

sbernard31:
About https://github.com/eclipse/leshan/pull/1226#issuecomment-1086024736,

> One simple thing to test would be to simply remove these lines and see if observations with OSCORE still works. I will try to have a look at this today or tomorrow to understand how things can be fixed (if still an issue) in the OSCORE code in Californium.

I tested it on my side and it seems observe works with or without this code.
Testing this I saw that passive cancel observation seems to not work with or without this code. (so probably not related)

rikard-sics:
> I tested it on my side and it seems observe works with or without this code.
> Testing this I saw that passive cancel observation seems to not work with or without this code. (so probably not related)

Very good. That code may very well have been a workaround for something which is fixed now in later releases.

I will make a note to look into the passive observation cancellations.

leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java
rikard-sics: Looking into things a bit this is my understanding. The default Master Salt in OSCORE should indeed be the empty byte string. However, later when that is fed into the HKDF, if the salt is zero length, it should be padded:
https://datatracker.ietf.org/doc/html/rfc5869#section-2.2
```
Inputs:
      salt     optional salt value (a non-secret random value);
               if not provided, it is set to a string of HashLen zeros.
```

So that is why if the master salt is null that padding occurs. But yeah what seems to be missing is considering if the developer sets the empty byte string as salt in the context constructor. So it would be better if it was:
```
if (master_salt == null || master_salt.length == 0) {
	// Default value. Automatically initialized with 0-es.
	this.common_master_salt = new byte[this.kdf.getKeySize() / Byte.SIZE];
} else {
	this.common_master_salt = master_salt.clone();
}
```
(Or waiting to do the padding in the deriveKey method may be even clearer). I will add this to the list of updates to do.

sbernard31:
I think we can integrate this in `oscore` branch ? Tell me if you think this is not ready.

rikard-sics:
> I think we can integrate this in `oscore` branch ? Tell me if you think this is not ready.

Yes I believe this can be integrated now.

sbernard31:
I rebased `oscore` branch on `master` then integrated this PR `oscore` branch in the squashed commit 11d4053a28dbcdfab41aef99e38319c4c3843f0e

PR #1255, opened 2022-05-13
Experiment about DataCollector Idea

sbernard31:
Here are some idea about DataCollector.

This is very crappy code just to share : 
- ideas
- detect and begin to discuss about some issue.

I think maybe a first version of the code could contains only : 
- ManualDataCollector
- maybe remove the support of several DataCollector.
- I don't know if we keep DataCollectorManager ?

Then with this ManualDataCollector this should be enough for : 
- write unit test
- add some CLI commands to leshan client demo

About CLI, we can imagine to add a `collect` interactive command (e.g. : collect /3/0/1 /3/0/2)
And also add an option to `send`  interactive command which is `send --collected-data`.


sbernard31:
Data Collector was replaced by Data Sender. (See #1274)

PR #1258, opened 2022-05-27
New leshan-client-demo CLI feature list nodes

JaroslawLegierski:
New leshan-client-demo feature responsible for listing existing objects/object instances/resources

### DIFF
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/cli/interactive/InteractiveCommands.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/cli/interactive/InteractiveCommands.java
@@ -12,7 +12,11 @@ import org.eclipse.leshan.client.demo.cli.interactive.InteractiveCommands.MoveCo
 import org.eclipse.leshan.client.demo.cli.interactive.InteractiveCommands.SendCommand;
 import org.eclipse.leshan.client.demo.cli.interactive.InteractiveCommands.UpdateCommand;
 import org.eclipse.leshan.client.demo.cli.interactive.InteractiveCommands.ListCommand;
-import org.eclipse.leshan.client.resource.*;
+import org.eclipse.leshan.client.resource.LwM2mObjectTree;
+import org.eclipse.leshan.client.resource.ObjectsInitializer;
+import org.eclipse.leshan.client.resource.LwM2mObjectEnabler;
+import org.eclipse.leshan.client.resource.ObjectEnabler;
+import org.eclipse.leshan.client.resource.LwM2mInstanceEnabler;
 import org.eclipse.leshan.client.servers.ServerIdentity;
 import org.eclipse.leshan.core.LwM2m.Version;
 import org.eclipse.leshan.core.LwM2mId;
@@ -77,10 +81,10 @@ public class InteractiveCommands implements Runnable, JLineInteractiveCommands {
     /**
      * A command to list objects.
      */
-    @Command(name = "list", description = "List Objects", headerHeading = "%n", footer = "")
+    @Command(name = "list", description = "List available Objects, Instances and Resources", headerHeading = "%n", footer = "")
     static class ListCommand implements Runnable {
 
-        @Parameters(description = "Id of the LWM2M object to enable", index = "0", defaultValue = "-1")
+        @Parameters(description = "List of resources for given LwM2M object id", index = "0", arity = "0..1")
         private Integer id;
 
         @ParentCommand
@@ -95,7 +99,7 @@ public class InteractiveCommands implements Runnable, JLineInteractiveCommands {
                 return;
             }
             objectTree.getObjectEnablers().forEach((objectId, objectValue) -> {
-                if (id != -1 && !id.equals(objectId)) {
+                if (id != null && !id.equals(objectId)) {
                     return;
                 }
                 ObjectModel objectModel = objectValue.getObjectModel();
@@ -104,7 +108,7 @@ public class InteractiveCommands implements Runnable, JLineInteractiveCommands {
                     List<Integer> availableResources = objectValue.getAvailableResourceIds(instance);
                     availableResources.forEach(resourceId -> {
                         ResourceModel resourceModel = objectModel.resources.get(resourceId);
-                        parent.out.printf("  * %d : %s%n", resourceId, resourceModel.name);
+                        parent.out.printf("  /%d : %s%n", resourceId, resourceModel.name);
                     });
                 });
             });
--- a/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/cli/interactive/InteractiveCommands.java
+++ b/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/cli/interactive/InteractiveCommands.java
@@ -94,30 +94,23 @@ public class InteractiveCommands implements Runnable, JLineInteractiveCommands {
                 parent.out.flush();
                 return;
             }
-            objectTree.getObjectEnablers().forEach(
-                    (objectId, objectValue) -> {
-                        if (id != -1 && !id.equals(objectId) ) {
-                            return;
-                        }
-                        ObjectModel objectModel = objectValue.getObjectModel();
-                        objectValue.getAvailableInstanceIds().forEach(
-                                instance -> {
-                                    parent.out.printf("/%d/%d : %s%n", objectId, instance, objectModel.name);
-                                    List<Integer> availableResources = objectValue.getAvailableResourceIds(instance);
-                                    availableResources.forEach(
-                                            resourceId -> {
-                                                ResourceModel resourceModel = objectModel.resources.get(resourceId);
-                                                parent.out.printf("  * %d : %s%n", resourceId, resourceModel.name);
-                                            }
-                                    );
-                                }
-                        );
-                    }
-            );
+            objectTree.getObjectEnablers().forEach((objectId, objectValue) -> {
+                if (id != -1 && !id.equals(objectId)) {
+                    return;
+                }
+                ObjectModel objectModel = objectValue.getObjectModel();
+                objectValue.getAvailableInstanceIds().forEach(instance -> {
+                    parent.out.printf("/%d/%d : %s%n", objectId, instance, objectModel.name);
+                    List<Integer> availableResources = objectValue.getAvailableResourceIds(instance);
+                    availableResources.forEach(resourceId -> {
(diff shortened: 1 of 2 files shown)


leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/cli/interactive/InteractiveCommands.java
```diff
@@ -75,6 +74,44 @@ public void run() {
...
+                if (id != -1 && !id.equals(objectId)) {
+                    return;
+                }
+                ObjectModel objectModel = objectValue.getObjectModel();
+                objectValue.getAvailableInstanceIds().forEach(instance -> {
+                    parent.out.printf("/%d/%d : %s%n", objectId, instance, objectModel.name);
+                    List<Integer> availableResources = objectValue.getAvailableResourceIds(instance);
+                    availableResources.forEach(resourceId -> {
+                        ResourceModel resourceModel = objectModel.resources.get(resourceId);
+                        parent.out.printf("  * %d : %s%n", resourceId, resourceModel.name);
```
sbernard31, line 107: Should we use / instead of * ? 

E.g. : 
```
/6/0 : Location 
  /0 : Latitude
  /1 : Longitude
  /5 : Timestamp
```

sbernard31:
Some tiny advices : 

**1.** Do not create unnecessary commit, I mean here you can just squash your 2 commits in one. I could make sense to have seperate commits if you split in it atomic changes that make reviewing more easy. 

*2.** Before to push a PR, it could be a good idea to fetch the target branch and so rebase your work on last commit of this branch.

sbernard31:
Clearly optionnal, If you want we can add some ANSI color/formatting to the list command output like this : 

```java
    static class ListCommand implements Runnable {

        @Parameters(description = "Id of the LWM2M object to enable", arity = "0..1")
        private Integer id;

        @Spec
        CommandSpec spec;

        @ParentCommand
        InteractiveCommands parent;

        @Override
        public void run() {
             ... 
                objectValue.getAvailableInstanceIds().forEach(instance -> {
                    // existing pattern are available at  https://picocli.info/#_usage_help_with_styles_and_colors
                    String colorpattern = spec.commandLine().getColorScheme().ansi()
                            .string("/%d/%d : @|bold,underline %s|@ %n");
                    parent.out.printf(colorpattern, objectId, instance, objectModel.name);
                   ...
                });
             ...
        }
```
(This is just an example)

Maybe it could be good to add a more elegant API to output ANSI styled string in InteractiveCommands :thinking:  
but probably out of topic for this PR.

sbernard31:
@JaroslawLegierski, I created PR #1261 about styled output, so if you want to use styled output you should base your work on [`ansi_cli`](https://github.com/eclipse/leshan/tree/ansi_cli) branch.

In any case, you can review the PR #1261, if you want.

JaroslawLegierski:
I fixed all mentioned issues in a new commit. [5f74535](https://github.com/eclipse/leshan/pull/1258/commits/5f74535827a6ff0d06542e93b18d2cddcdc30a77)

Regarding ANSI color/formatting I'm working on it using ansi_cli branch https://github.com/eclipse/leshan/tree/ansi_cli . Unfortunately I have some problems with my test environment. I'm using Windows 10 and after adding AnsiConsole.systemInstall(); in main()  ANSI formatting works however, it looks some strange in Powershell:
obraz
in cmd I also don't see underline or bold characters:
obraz
After adding AnsiConsole.systemInstall() in IntelliJ Run window ANSI color/formatting has stopped working completely. :worried:

Interesting that in  Linux (Ubuntu) everything is ok
obraz

Are we taking into account MS Windows as the operating system for leshan-client-demo or should I focus only on Linux?

sbernard31:
About cmd, I guess the terminal doesn't support ANSI style so I think you get the expected behavior.
About powershell, my guess is that its bold/underline style just looks like this :shrug: . That sounds strange but sometime ANSI color theme are surprising. I know with some theme even color is not respected (e.g. red is not red)

So from my point of view it behave as expected.

About the style I was using, it's just an example to show you how it works, feel free to style the output as you wish.

sbernard31:
@JaroslawLegierski tell me if you think I should integrate #1261 in `master` ?

JaroslawLegierski:
@sbernard31 I think it might be a good idea - it will be easier to work on ANSI color/formatting with mater branch

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/cli/interactive/InteractiveCommands.java
```diff
@@ -77,10 +81,10 @@ public void run() {
     /**
      * A command to list objects.
      */
-    @Command(name = "list", description = "List Objects", headerHeading = "%n", footer = "")
+    @Command(name = "list", description = "List available Objects, Instances and Resources", headerHeading = "%n", footer = "")
     static class ListCommand implements Runnable {
 
-        @Parameters(description = "Id of the LWM2M object to enable", index = "0", defaultValue = "-1")
+        @Parameters(description = "List of resources for given LwM2M object id", index = "0", arity = "0..1")
         private Integer id;
```
sbernard31, line 88: maybe should be rename in objectId;

JaroslawLegierski: OK - done

leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/cli/interactive/InteractiveCommands.java
```diff
@@ -77,10 +81,10 @@ public void run() {
     /**
      * A command to list objects.
      */
-    @Command(name = "list", description = "List Objects", headerHeading = "%n", footer = "")
+    @Command(name = "list", description = "List available Objects, Instances and Resources", headerHeading = "%n", footer = "")
     static class ListCommand implements Runnable {
 
-        @Parameters(description = "Id of the LWM2M object to enable", index = "0", defaultValue = "-1")
+        @Parameters(description = "List of resources for given LwM2M object id", index = "0", arity = "0..1")
```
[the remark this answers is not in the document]
JaroslawLegierski: Maybe we'll use a shorter form:  "Id of the object, if no value is specified all available objects will be listed." ?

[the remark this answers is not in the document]
sbernard31: Yep maybe it's enough :)

JaroslawLegierski:
@sbernard31  Please find in commit https://github.com/eclipse/leshan/pull/1258/commits/8f92a8b00eb86f8e37fb40c2ba38fe2ea3399ea7 my proposal for formatting the list nodes function. (I'm sorry, I did it through merge not rebase - next time I will remember your recommendations)

obraz

I added AnsiConsole.systemInstall(); in main() Of course, if we do not plan to support ANSI color/formatting on Windows, it can be removed. But if we want to colorize leshan-client-demo in Windows I propose to remove style underline from leshan-client-demo because this is the source of the problems (in Powershell underline is interpreted as yellow letters on a pink background)

obraz

Of course, everything looks fine under Linux as well:

obraz

sbernard31:
I will look at that this afternoon.

About rebase, you can **eventually** fix your PR using a rebase then just push --force your banch again.

About, `AnsiConsole.systemInstall()` I must confess that I'm a bit hesitant because I haven't windows machine to test and I don't want to support issue about color on windows in the future. By curiosity do you have tested it with old `cmd` tools  ? 

> But if we want to colorize leshan-client-demo in Windows I propose to remove style underline from leshan-client-demo because this is the source of the problems (in Powershell underline is interpreted as yellow letters on a pink background)

I think it's not a good idea to take care of this bad windows behavior. I'm not even sure we can generalized it to windows, I bet this just about your terminal application with that particular theme. Do you try another terminal ?  maybe like this [windows-terminal](https://apps.microsoft.com/store/detail/windows-terminal/9N0DX20HK701?hl=fr-fr&gl=FR) (:warning: I never use windows so I don't know so much about it)

JaroslawLegierski:
> About, AnsiConsole.systemInstall() I must confess that I'm a bit hesitant because I haven't windows machine to test and I don't want to support issue about color on windows in the future. By curiosity do you have tested it with old cmd tools ?

Yes -in cmd we have similar problem (but in black & white version)
with underline style:
obraz
without underline:
obraz

> I think it's not a good idea to take care of this bad windows behavior. I'm not even sure we can generalized it to windows, I bet this just about your terminal application with that particular theme. Do you try another terminal ? maybe like this [windows-terminal](https://apps.microsoft.com/store/detail/windows-terminal/9N0DX20HK701?hl=fr-fr&gl=FR) (⚠️ I never use windows so I don't know so much about it)

In Windows Terminal https://apps.microsoft.com/store/detail/windows-terminal/9N0DX20HK701?hl=pl-pl&gl=PL  , leshan-demo-client looks similar to cmd
obraz
In general, underline doesn't look as bad as in Powershell in both cmd and Windows Terminal (but because it is displayed in black and white)

sbernard31:
> Yes -in cmd we have similar problem (but in black & white version) with underline style

I was more curious to know if AnsiConsole.install add any strange character in the console output but it seems it works fine.


So I integrate your work in `master` (commit 30bad682481e97e956025a016b934d5264d0b36c) : 
- I cleaned some imports
- better handle if given id is not available. 
- change parameter name id to object id (https://github.com/eclipse/leshan/pull/1258#discussion_r888915975) 

About underline, I prefer to keep it.  So I changed it back. 
Looking at your screenshot, this seems usable anyway.  Maybe only powershell could be consider as uggly but too bad if Microsoft is not able to provide a proper terminal. I don't think we should adapt the code just for one tool with a strange behavior.

About `AnsiConsole.install()`, I didn't integrate it for now. I think this should be added to all demo and so maybe in a sperated commit / PR.

About this last point do you really want to add it ?

JaroslawLegierski:
>About AnsiConsole.install(), I didn't integrate it for now. I think this should be added to all demo and so maybe in a sperated commit / PR.

>About this last point do you really want to add it ?

As for AnsiConsole.install (), in my opinion we do not necessarily have to support colors in Windows. Leshan-client-demo looks good also in the B&W version.

sbernard31:
> As for AnsiConsole.install (), in my opinion we do not necessarily have to support colors in Windows. Leshan-client-demo looks good also in the B&W version.

Ok, so I will not added for now. But if you change your mind feel free to open a new issue about it.

PR #1259, opened 2022-05-29
Core Link resource type support

jvermillard:

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mSingleResource.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mSingleResource.java
@@ -68,6 +68,9 @@ public class LwM2mSingleResource implements LwM2mResource {
         if (value instanceof ObjectLink) {
             return new LwM2mSingleResource(id, value, Type.OBJLNK);
         }
+        if (value instanceof Link[]) {
+            return new LwM2mSingleResource(id, value, Type.CORELINK);
+        }
         if (value instanceof ULong) {
             return new LwM2mSingleResource(id, value, Type.UNSIGNED_INTEGER);
         }
@@ -146,7 +149,7 @@ public class LwM2mSingleResource implements LwM2mResource {
         return new LwM2mSingleResource(id, objlink, Type.OBJLNK);
     }
 
-    public static LwM2mSingleResource newObjectLinkResource(int id, Link[] coreLinks) {
+    public static LwM2mSingleResource newCoreLinkResource(int id, Link[] coreLinks) {
         return new LwM2mSingleResource(id, coreLinks, Type.CORELINK);
     }
 
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mSingleResource.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mSingleResource.java
@@ -1,15 +1,15 @@
 /*******************************************************************************
  * Copyright (c) 2015 Sierra Wireless and others.
- * 
+ *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * and Eclipse Distribution License v1.0 which accompany this distribution.
- * 
+ *
  * The Eclipse Public License is available at
  *    http://www.eclipse.org/legal/epl-v20.html
  * and the Eclipse Distribution License is available at
  *    http://www.eclipse.org/org/documents/edl-v10.html.
- * 
+ *
  * Contributors:
  *     Sierra Wireless - initial API and implementation
  *******************************************************************************/
@@ -20,6 +20,7 @@ import java.util.Date;
 import java.util.Map;
 import java.util.NoSuchElementException;
 
+import org.eclipse.leshan.core.link.Link;
 import org.eclipse.leshan.core.model.ResourceModel.Type;
 import org.eclipse.leshan.core.util.datatype.ULong;
 
@@ -116,6 +117,11 @@ public class LwM2mSingleResource implements LwM2mResource {
                 throw new LwM2mNodeException(
                         String.format(doesNotMatchMessage, value.getClass().getSimpleName(), type));
             break;
+        case CORELINK:
+            if (!(value instanceof Link[]))
+                throw new LwM2mNodeException(
+                        String.format(doesNotMatchMessage, value.getClass().getSimpleName(), type));
+            break;
         case UNSIGNED_INTEGER:
             if (!(value instanceof ULong)) {
                 throw new LwM2mNodeException(
@@ -138,6 +144,10 @@ public class LwM2mSingleResource implements LwM2mResource {
 
     public static LwM2mSingleResource newObjectLinkResource(int id, ObjectLink objlink) {
         return new LwM2mSingleResource(id, objlink, Type.OBJLNK);
+    }
+
+    public static LwM2mSingleResource newObjectLinkResource(int id, Link[] coreLinks) {
+        return new LwM2mSingleResource(id, coreLinks, Type.CORELINK);
     }
 
     public static LwM2mSingleResource newBooleanResource(int id, boolean value) {
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mSingleResource.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mSingleResource.java
@@ -1,15 +1,15 @@
 /*******************************************************************************
  * Copyright (c) 2015 Sierra Wireless and others.
- *
+ * 
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * and Eclipse Distribution License v1.0 which accompany this distribution.
- *
+ * 
  * The Eclipse Public License is available at
  *    http://www.eclipse.org/legal/epl-v20.html
  * and the Eclipse Distribution License is available at
  *    http://www.eclipse.org/org/documents/edl-v10.html.
- *
+ * 
  * Contributors:
  *     Sierra Wireless - initial API and implementation
  *******************************************************************************/
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvDecoder.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvDecoder.java
@@ -1,15 +1,15 @@
 /*******************************************************************************
  * Copyright (c) 2015 Sierra Wireless and others.
- *
+ * 
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * and Eclipse Distribution License v1.0 which accompany this distribution.
- *
+ * 
  * The Eclipse Public License is available at
  *    http://www.eclipse.org/legal/epl-v20.html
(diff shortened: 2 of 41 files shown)


leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvDecoder.java
```diff
@@ -60,6 +79,21 @@ public <T extends LwM2mNode> T decode(byte[] content, LwM2mPath path, LwM2mModel
...
+        return linkParser;
+    }
+
+    /**
+     * Set the Link parser used for parsing core links. If not set the
+     * {@link DefaultLwM2mLinkParser} is used.
+     */
+    public void setLinkParser(LinkParser linkParser) {
+        this.linkParser = linkParser;
+    }
```
sbernard31, line 95: This doesn't compile because of : `The final field LwM2mNodeTlvDecoder.linkParser cannot be assigned`

I think we can remove the getter and setter.

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvDecoder.java
```diff
@@ -49,6 +52,22 @@ public class LwM2mNodeTlvDecoder implements NodeDecoder {
 
     private static final Logger LOG = LoggerFactory.getLogger(LwM2mNodeTlvDecoder.class);
 
+    // parser used for core link parser
```
sbernard31, line 55: `// parser used for corelink data type` ?

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvDecoder.java
```diff
@@ -49,6 +52,22 @@ public class LwM2mNodeTlvDecoder implements NodeDecoder {
...
+
+    public LwM2mNodeTlvDecoder() {
+        this(new DefaultLwM2mLinkParser());
+    }
+
+    /**
+     * Create a new LwM2mNodeTlvDecoder with a custom link parser.
+     * 
+     * @param linkParser the link parser for core link format resources.
+     */
```
sbernard31, line 66: cosmetic remark but you can use link when you refer class : 
E.g. `Create a new {@link LwM2mNodeTlvDecoder} with a custom {@link LinkParser}.`

leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mSingleResource.java
```diff
@@ -116,6 +117,11 @@ public static LwM2mSingleResource newResource(int id, Object value, Type type) {
                 throw new LwM2mNodeException(
                         String.format(doesNotMatchMessage, value.getClass().getSimpleName(), type));
             break;
+        case CORELINK:
+            if (!(value instanceof Link[]))
+                throw new LwM2mNodeException(
+                        String.format(doesNotMatchMessage, value.getClass().getSimpleName(), type));
+            break;
```
sbernard31, line 124: I guess a new static constructor should be added for corelink in this class ?
And probably this should be also be implemented in `LwM2mResourceInstance` and `LwM2mMultipleResource`

leshan-server-demo/webapp/src/js/valueutils.js
```diff
@@ -22,6 +22,9 @@ function valueToString(value, type) {
   } else if (type == "time") {
     let date = new Date(value);
     return date.toDateString()+ " - " +  date.toLocaleTimeString() + " (" + date.getTime()/1000 + "s)";
+  } else if (type == "corelink") {
+    // for now display the corelink as a json structure
+    return JSON.stringify(value);
```
sbernard31, line 27: we should probably implement this ?
Or eventually change the JSON serialization and just return a string for corelink for now.

(As you prefer)

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvEncoder.java
```diff
@@ -66,6 +69,8 @@ public byte[] encode(LwM2mNode node, LwM2mPath path, LwM2mModel model, LwM2mValu
     }
 
     private static class InternalEncoder implements LwM2mNodeVisitor {
+    
+        private LinkSerializer linkSerializer = new DefaultLinkSerializer();
```
sbernard31, line 73: should be moved at LwM2mNodeTlvEncoder level with a custom constructor

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
```diff
@@ -83,6 +84,14 @@ public static void loadModel() {
...
         objects.add(
                 new ObjectModel(66, "object link tests 66", "", ObjectModel.DEFAULT_VERSION, true, false, resForObj66));
+        
+        // add object 25 from the LWM2M gateway specification
+        List<ResourceModel> resForObj25 = new ArrayList<>();
+        resForObj25.add(new ResourceModel(0, "Device ID", Operations.R, true, false, Type.STRING, null, null, null));
+        resForObj25.add(new ResourceModel(1, "Prefix", Operations.R, false, false, Type.STRING, null, null, null));
+        resForObj25.add(new ResourceModel(3, "IoT Device Objects", Operations.R, false, false, Type.CORELINK, null, null, null));
+        objects.add(
+                new ObjectModel(25, "Gateway", "", ObjectModel.DEFAULT_VERSION, true, false, resForObj25));
```
sbernard31, line 94: I would prefer we use the object [3342.xml](https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/3342.xml)

It is currently not used because it was added recently but we should use it in our tests.

Let's this aside for now, I will think a bit more how to make it available for test more easily.

sbernard31 (commented):
There is lot of formatting issue :grimacing:  but this is common issue from contribution where eclipse is not used :disappointed: 
(Tthis is probably a topic to address but hard to put it in priority list)

If I didn't missed something for now you only target TEXT and TLV correct ? any reason to not support other content format ? 

You should also add coreLink support to LwM2mTestObject class
So you can easily test it with **leshan-client-demo** and **leshan-server-demo**.

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvEncoder.java
jvermillard: done, but InternalEncoder is not static any more (to access the non static parent field)

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvDecoder.java
jvermillard: whoops, done

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/tlv/LwM2mNodeTlvDecoder.java
jvermillard: done

leshan-core/src/main/java/org/eclipse/leshan/core/node/LwM2mSingleResource.java
jvermillard: done

leshan-server-demo/webapp/src/js/valueutils.js
jvermillard: I like the fact the link is presented in the API a way you don't have to parse again, but I have no idea how to do a clean presentation in the front code

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
jvermillard: > There is lot of formatting issue but this is common issue from contribution where eclipse is not used 

And I feed vscode with the eclipse formatter parameters.. since it's supposed to be the same code under the hood..

> (This is probably a topic to address but hard to put it in priority list)

Or just ignore it and review with whitespace diff off :hand_over_mouth: 

> If I didn't missed something for now you only target TEXT and TLV correct ? any reason to not support other content format ?

Yes for now I want to make TEXT and TLV working and see for the other content type in a second step

> You should also add coreLink support to LwM2mTestObject class So you can easily test it with leshan-client-demo and leshan-server-demo.

Does mean we need to publish a new revision of the object @ OMA?

sbernard31:
> Does mean we need to publish a new revision of the object @ OMA?

Oops my mistake, it's not 3342 but  [3442](https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/3442.xml), and it already contains a coreLink resource (single and multi instance)

leshan-server-demo/webapp/src/js/valueutils.js
sbernard31: > I like the fact the link is presented in the API a way you don't have to parse again,

Yep but as we don't use it in the UI this could be OK to have value as string in the UI waiting we need something better.

>  I have no idea how to do a clean presentation in the front code

I think the classic string representation of Link is OK. (it should not be so hard to implement it in javascript I guess :thinking: )

and alternative (probably not so good) solution having both representation in the REST API. 
e.g. ` { str : "</3>;</4>...", val: { ... }}`

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
sbernard31: > Does mean we need to publish a new revision of the object @ OMA?

Oops my mistake, it's not 3342 but  [3442](https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/3442.xml), and it already contains a coreLink resource (single and multi instance)

leshan-server-demo/webapp/src/js/valueutils.js
jvermillard: done, I serialize it back to corelnk string and display as-is:
image

sbernard31:
@jvermillard I push some more code. You can review it if you want. 

With my last commit, you  should be able to test coreLink with leshan-client-demo, try to play with object 3442 and Randomize, Reset, Clear, CoreLnk Value and Multiple CoreLink value resources.

About tests, we should add integration tests, at least in WriteSingleValueTest, this will create a test for all content format.
(I don't know if keep the unit test LwM2mNodeDecoderTest :thinking:)

We should also add support of write coreLink in leshan-server-demo. 

And also support for other content format.

sbernard31:
(I found a kind of bug with serialization with rootpath but as this is not related to this PR, I try to fix this in another commit or PR)

sbernard31:
I finally added : 
- write support
- WriteSingleValueTest
- and support for other decoder/encoder

sbernard31:
Integrated in `master` (commit d5e5a7d6662f82d6c4dd5f0a857028b7f9994dc3)
(I updated supported feature page : https://github.com/eclipse/leshan/wiki/LWM2M-1.1-supported-features)

Thx @jvermillard  :pray:

PR #1263, opened 2022-06-02
Forbid overlapped paths in Observe-Composite request

adamsero:


leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/RootEnabler.java
```diff
@@ -212,6 +213,12 @@ public WriteCompositeResponse write(ServerIdentity identity, WriteCompositeReque
     public synchronized ObserveCompositeResponse observe(ServerIdentity identity, ObserveCompositeRequest request) {
         List<LwM2mPath> paths = request.getPaths();
 
+        try {
+            validatePathsNotOverlapping(paths);
+        } catch (InvalidRequestException exception) {
+            return ObserveCompositeResponse.badRequest(exception.getMessage());
+        }
+
```
sbernard31, line 221: I think this is not needed because `ObserveCompositeResponse.getPaths();` can not contain overlapped paths, because if you try to do that an `InvalidRequestException` will be raised.

Maybe it worth to double check but I think in client `RootResource` we will try to create the compositeRequest, the exception will be raised, and then catch and return a bad request. (see `LwM2mCoapResource.handleRequest`)

leshan-core/src/main/java/org/eclipse/leshan/core/request/ObserveCompositeRequest.java
```diff
@@ -88,13 +88,31 @@ public ObserveCompositeRequest(ContentFormat requestContentFormat, ContentFormat
...
+        for (int i = 0; i < paths.size(); i++) {
+            LwM2mPath firstPath = paths.get(i);
+            for (int j = i + 1; j < paths.size(); j++) {
+                LwM2mPath secondPath = paths.get(j);
+                if (firstPath.startWith(secondPath) || secondPath.startWith(firstPath)) {
+                    throw new InvalidRequestException("Invalid path list :  %s and %s are overlapped paths", firstPath,
+                            secondPath);
+                }
+            }
+        }
```
sbernard31, line 113: Maybe we should factorize that code ? 
I don't know where to put it :thinking:

Maybe we add a static method in LwM2mPath like : 
```java
public static boolean containOverlappedPath(List<LwM2mPath> paths)
```

leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/RootEnabler.java
adamsero: You're right, but there's no way to be 100% sure, since server and client check share the same validator in `ObserveCompositeResponse`'s constructor. That being seid, if I remove the check in `RootEnabler`, it should still work.

leshan-core/src/main/java/org/eclipse/leshan/core/request/ObserveCompositeRequest.java
adamsero: I added something of the like, I think making it void and throw an exception is a better idea - what do you think?

leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/RootEnabler.java
sbernard31: By double check, I meant that it worth you read this code too to be sure you understand like me.

If you understand like me, I think we can remove the code in `RootEnabler`.

leshan-core/src/main/java/org/eclipse/leshan/core/request/ObserveCompositeRequest.java
sbernard31: I'm not sure that `LwM2mPath` class should  raise `InvalidRequestException`, I think this is maybe not so reusable. :thinking:

adamsero: You're right, though I still don't like the validation method returning boolean - then if it fails we don't have error message and we have to duplicate it every single time, so that kind of defeats the purpose. Maybe we can do something like:

```java
public static void validateNotOverlapping(List<LwM2mPath> paths) throws RuntimeException {
    ...
}
public static void validateNotEmpty(List<LwM2mPath> paths) throws RuntimeException {
    ...
}

...

try {
    LwM2mPath.validateNotEmpty(paths);
    LwM2mPath.validateNotOverlapping(paths);
} catch(RuntimeException exception) {
    throw new InvalidRequestException(exception.getMessage());
}
```

leshan-client-core/src/main/java/org/eclipse/leshan/client/resource/RootEnabler.java
adamsero: Yup, I understand it and I did remove the code from `RootEnabler`

leshan-core/src/main/java/org/eclipse/leshan/core/request/ObserveCompositeRequest.java
sbernard31: Yep you're right my with my way we lost the error message.
You can see in `LwM2mNodeUtil`, there is way to solve this return a String errorMessage instead of boolean.
(I don't know if that sounds good to you?)

You're way works too but raising RuntimeException sounds strange.
I'm not sure `InvalidLwM2mPathException` is adapted as path are not really invalid ?
or IllegalArgumentException like in `Validate` class ?

adamsero: With String return it would be:
```java
public static String validateNotOverlapping(List<LwM2mPath> paths) {
    //returns error message string or null if everything is good
}
public static String validateNotEmpty(List<LwM2mPath> paths) {
    //same as above
}

...

String validateNotEmptyMessage = LwM2mPath.validateNotEmpty(paths);
if(validateNotEmptyMessage != null) {
    throw new InvalidRequestException(validateNotEmptyMessage);
}
String validateNotOverlappingMessage = LwM2mPath.validateNotOverlapping(paths);
if(validateNotOverlappingMessage != null) {
    throw new InvalidRequestException(validateNotOverlappingMessage);
}
```

So it's pretty long and awkward IMO. We could also just use `IllegalArgumentException` or `InvalidLwM2mPathException` instead of `RuntimeException` in my last concept - that's up to you how you want it.

sbernard31: Ok let's go for throwing an exception probably IllegalArgumentException is better.
Note that org.eclipse.leshan.core.util.Validate.notEmpty(Collection) already to this kind of check

adamsero: OK, I changed the exception class and removed empty check from `LwM2mPath` and used the one from `Validate`. Hope that's finally good.

PR #1271, opened 2022-06-23
Enhance Maven Build configuration.

sbernard31:
This doesn't change so much the behavior.
But pom.xml are better organized and some goals are no more called when this is not necessary. 

The old deprecated `mvn eclipse:eclipse` way to configure eclipse was removed.


sbernard31:
Unfortenately I discover that `mvn versions:set ...` doesn't work anymore for this structure. :grimacing: 
(Need to investigate : https://github.com/mojohaus/versions-maven-plugin/issues/605)

sbernard31:
Integrated in `master` (commit c560074a508174aee0f0f77d386fedd3bfefff87)

PR #1288, opened 2022-07-18
fix #1287 : log uncaught callback exceptions

### LINKED ISSUE #1287: Why not logging uncaught exception ?
https://github.com/eclipse/leshan/blob/master/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/CoapAsyncRequestObserver.java#L101

using something like: `LOG.warn("Uncaught exception during onResponse callback", e);`  would simplify the debug, WDYT?

jvermillard:

### DIFF
--- a/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/CoapAsyncRequestObserver.java
+++ b/leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/CoapAsyncRequestObserver.java
@@ -98,7 +98,7 @@ public class CoapAsyncRequestObserver extends AbstractRequestObserver {
             try {
                 responseCallback.onResponse(coapResponse);
             } catch (RuntimeException e) {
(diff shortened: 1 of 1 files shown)


leshan-core-cf/src/main/java/org/eclipse/leshan/core/californium/CoapAsyncRequestObserver.java
```diff
@@ -192,4 +192,4 @@ private synchronized void cancelCleaningTask() {
         }
         cancelled = true;
     }
-}
\ No newline at end of file
+}
```
sbernard31: Sounds to related to the fix.

AFAIK, there is no formatting rules about that currently if you think we need to add one let me know.

sbernard31:
Integrated in `master` (commit 50a719bad11fc66f2878e8e5926f929d09b5f094)

PR #1289, opened 2022-07-18
Visual studio code workspace settings for java

jvermillard:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,9 +24,6 @@ bin/
 .idea/
 *.iml
 
-# Visual Studio Code files #
-.vscode/
-
 # Package Files #
 *.jar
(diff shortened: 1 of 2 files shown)


jvermillard:
you need the official java extension (IIRC there is a popup proposing to install the first time you open a java file)

jvermillard:
I need to re-check but the last time I tried that it was not working. Apparently it a bug which was fixed in vscode long time ago :thinking:

sbernard31:
What do we do with that ? 
Should we finalize it OR should I close the PR ?

sbernard31:
@jvermillard what should we do with that ?

We close it ? or we make it works ? :wink:

Issue #1304, opened 2022-09-01
Timestamps in seconds or milliseconds ?


JaroslawLegierski:
We got following feedback in timestamps topic which needs some clarifications. Timestamps in Leshan client in ManualDataSender class [https://github.com/eclipse/leshan/blob/master/leshan-client-core/src/main/java/org/eclipse/leshan/client/send/ManualDataSender.java](https://github.com/eclipse/leshan/blob/master/leshan-client-core/src/main/java/org/eclipse/leshan/client/send/ManualDataSender.java) in collectData are curently calculated in miliseconds (ManualDataSender line 47). Maybe they are supposed to be calculated in seconds (as in SenML) ? The problem is wider, should be the timestamps in TimestampedLwM2mNode and TimestampedLwM2mNodes in milliseconds or in seconds only? 
Currently only SenML manage timestamped values (in seconds) byt maybe in a future data format they could be in ms. What is your opinion on this topic ?

sbernard31:
Lot of Good questions ! and you probably find something which smell not so good ... 

Actually, SenML is no the only one to support timestamped values, JSON content format (application/vnd.oma.lwm2m+json, 11543) support it too (ok this is also a kind of SenML format but not the official one). 

I quickly look at the [LWM2M specification](http://openmobilealliance.org/release/LightweightM2M/V1_0_2-20180209-A/OMA-TS-LightweightM2M-V1_0_2-20180209-A.pdf) and I understand that timestamp for JSON is also in seconds.

So conclusion is the same because all current format are using seconds (and not milliseconds).

> Currently only SenML manage timestamped values (in seconds) byt maybe in a future data format they could be in ms. What is your opinion on this topic ?

I was thinking this too but  reading more deeper about [SenML RFC](https://www.rfc-editor.org/rfc/rfc8428) and LWM2M spec about JSON format, I finally understand that time are in seconds but could be float number, and so you could have milli or ever micro seconds (or even maybe more)... 

> The problem is wider, should be the timestamps in TimestampedLwM2mNode and TimestampedLwM2mNodes in milliseconds or in seconds only?

Reading most of the encoder / decoder I understand current code are using /expecting seconds.

But what should it be is less clear... especially with what I found above.  :point_up: 
 
So my feelling about this : 
**Short term :** It seems that almost all the code is thought with timestamp in seconds in mind, so we should fix code which does not respect this and document most of the code about timestamp node (we need to precise that seconds is expected)

**Mid/long term :**
[...]

PR #1306, opened 2022-09-07
Timestamps in seconds instead of milliseconds in ManualDataSender.

JaroslawLegierski:
Timestamps in ManualDataSender correction mentioned in issue: [https://github.com/eclipse/leshan/issues/1304](https://github.com/eclipse/leshan/issues/1304)


sbernard31:
Integrated in `master`. Thx for the fix :pray:  !

PR #1311, opened 2022-09-12
Override toString(), hashCode() and equals() methods in ObserveCompositeRequest

JaroslawLegierski:
Added toString, hashCode and equals to ObserveCompositeRequest

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/request/ObserveCompositeRequest.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/request/ObserveCompositeRequest.java
@@ -136,4 +136,46 @@ public class ObserveCompositeRequest extends AbstractLwM2mRequest<ObserveComposi
     public Map<String, String> getContext() {
         return context;
     }
+
+    @Override
+    public final String toString() {
+        return String.format(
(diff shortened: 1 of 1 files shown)


adamsero:
Fixed toString(), should be good to merge

sbernard31:
On so small PR/changes, do  not hesitate squash all you commit in one.

leshan-core/src/main/java/org/eclipse/leshan/core/request/ObserveCompositeRequest.java
```diff
@@ -136,4 +136,46 @@ public ContentFormat getResponseContentFormat() {
     public Map<String, String> getContext() {
         return context;
     }
+
+    @Override
+    public final String toString() {
+        return String.format(
+                "ObserveCompositeRequest [paths=%s requestContentFormat=%s responseContentFormat=%s, context=%s]",
```
sbernard31, line 143: I see there is a `,` just after `responseContentFormat=%s` but not after the others.

So I checked in code base and there is case where `,` is used and so other where it is not used.
But when you generate `toString` with eclipse, `,` is used.

So I will add it

sbernard31:
Integrated in master (commit ecc113c4351fc351b9d300030e24b9c801be2737)

PR #1313, opened 2022-09-19
New timestamp format using BigDecimal and Instant

adamsero:


sbernard31:
@adamsero, looking at your PR, I see that github action does not work as expected.
When error is detected, comments should be added to PR to help you to resolve it. This should like this https://github.com/sbernard31/leshan/pull/12.

I create #1314  to try to resolve it.

adamsero:
@sbernard31 regarding your [comment](https://github.com/eclipse/leshan/issues/1314#issue-1377980113), specifically 
> Could you not fix the last one about `Check Android API Compliance` ?

I'm getting a bunch of errors caused by the use of `Instant` class, since it's not compatible with Android API v19 as defined in `build-config/lib-build-config/pom.xml`:
```xml
<signature>
    <groupId>net.sf.androidscents.signature</groupId>
    <artifactId>android-api-level-19</artifactId>
    <version>4.4.2_r4</version>
</signature>
```
The fact that we check against API v19 released in 2013 means that we can't use any feature of Java 1.8 (from 2014), so we either have to bump the API version (I don't know what issues could arise from that) or this PR stays useless since I can't use the `Instant` class.

sbernard31:
After discussion at https://github.com/eclipse/leshan/issues/1304#issuecomment-1253939281.

We decide to bump android API level from 19 to 26.
So from 
```
<signature>
    <groupId>net.sf.androidscents.signature</groupId>
    <artifactId>android-api-level-19</artifactId>
    <version>4.4.2_r4</version>
</signature>
```
to something like 
```
<signature>
    <groupId>net.sf.androidscents.signature</groupId>
    <artifactId>android-api-level-26</artifactId>
    <version>8.0.0_r2</version>
</signature>
```

leshan-core/src/test/java/org/eclipse/leshan/core/node/TimestampedLwM2mNodesTest.java
```diff
@@ -31,13 +32,16 @@
 
 public class TimestampedLwM2mNodesTest {
 
+    private final Instant sampleTimestamp1 = Instant.ofEpochMilli(1_000_000_000);
+    private final Instant sampleTimestamp2 = Instant.ofEpochMilli(2_000_000_000);
```
sbernard31, line 36: I prefer the previous usage of number instead of variable as you can easily read which one is the most recent timestamp.

But I understand that now this is too verbose.

So maybe we should rename  : 
 - sampleTimestamp1 to ts_1_000_000_000_ms
 - sampleTimestamp2 to ts_2_000_000_000_ms
 
 Or something like this.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
```diff
@@ -219,17 +220,17 @@ public void json_encode_device_object_instance() {
...
         byte[] encoded = encoder.encodeTimestampedData(data, ContentFormat.JSON, new LwM2mPath(1024, 0, 1), model);
 
         StringBuilder b = new StringBuilder();
         b.append("{\"bn\":\"/1024/0/1\",\"e\":[");
-        b.append("{\"v\":22.9,\"t\":500},");
-        b.append("{\"v\":22.4,\"t\":510},");
-        b.append("{\"v\":24.1,\"t\":520}]}");
+        b.append("{\"v\":22.9,\"t\":5E+2},");
+        b.append("{\"v\":22.4,\"t\":5.1E+2},");
+        b.append("{\"v\":24.1,\"t\":5.2E+2}]}");
```
sbernard31, line 233: I don't like so much this way to serialize this number ?
We should maybe investigate a bit about that ?

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
```diff
@@ -402,14 +408,17 @@ public void senml_json_encode_timestamped_nodes() throws CodecException {
...
+                .put(timestamp.plusSeconds(4), new LwM2mPath(0, 0, 0),
+                        LwM2mSingleResource.newStringResource(0, "SampleString"))
+                .put(timestamp.plusSeconds(5), new LwM2mPath(0, 0, 1), LwM2mSingleResource.newBooleanResource(1, false))
+                .put(timestamp.plusSeconds(6), new LwM2mPath(0, 0, 2), LwM2mSingleResource.newIntegerResource(2, 456))
+                .build();
 
         byte[] encoded = encoder.encodeTimestampedNodes(timestampedLwM2mNodes, ContentFormat.SENML_CBOR, model);
 
-        String expectedString = "83a321662f302f302f30221a1dcd6504036c53616d706c65537472696e67a321662f302f302f31221a1dcd650504f4a321662f302f302f32221a1dcd6506021901c8";
+        String expectedString = "83a321662f302f302f3022c482001a1dcd6504036c53616d706c65537472696e67a321662f302f302f3122c482001a1dcd650504f4a321662f302f302f3222c482001a1dcd6506021901c8";
```
sbernard31, line 421: I'm not sure to understand why the serialization change here ?
At first sight I would say this is not expected.

leshan-core/src/main/java/org/eclipse/leshan/core/util/TimestampUtil.java
```diff
@@ -0,0 +1,35 @@
...
+        return BigDecimal.valueOf(timestamp.getEpochSecond()).add(decimalPoints);
+    }
+
+    public static BigDecimal fromMilliseconds(long millis) {
+        return BigDecimal.valueOf(millis).divide(BigDecimal.valueOf(1000));
+    }
+
+    public static BigDecimal fromNanoseconds(long nanos) {
+        return BigDecimal.valueOf(nanos).divide(BigDecimal.valueOf(1_000_000_000));
+    }
```
sbernard31, line 23: If this is not used now (or after all modifications), I think we should remove this 2 methods ?

sbernard31 (changes requested):

leshan-core/src/main/java/org/eclipse/leshan/core/util/TimestampUtil.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public static Instant fromSeconds(BigDecimal timestamp) {
+        if (timestamp == null) {
+            return null;
+        }
+
+        long seconds = timestamp.longValue();
```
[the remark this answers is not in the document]
adamsero: Well, the overflow will only happen for `BigDecimal` with integer part larger than 2^63, which is a huge value for a timestamp in seconds. I'm not sure how to detect that overflow though, if you want to throw an exception here. We also must extract the value to `long` in order to construct an `Instant`

leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/senml/LwM2mNodeSenMLEncoder.java
```diff
@@ -147,7 +150,9 @@ public byte[] encodeTimestampedData(List<TimestampedLwM2mNode> timestampedNodes,
             internalEncoder.converter = converter;
             internalEncoder.records = new ArrayList<>();
             timestampedLwM2mNode.getNode().accept(internalEncoder);
-            internalEncoder.records.get(0).setBaseTime(timestampedLwM2mNode.getTimestamp());
+            // TODO: change after refactoring TimestampedLwM2mNode
```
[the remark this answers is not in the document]
adamsero: That should have been removed, sorry for that

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
adamsero: I believe it's because `BigDecimal` is serialized differently than `long`, as you noticed [here](https://github.com/eclipse/leshan/pull/1313#discussion_r981181837) for JSON.

sbernard31: mmhh but here this is not about JSON but CBOR, so ?  (maybe I missed something)

adamsero: `CBORObject` encodes `long` and `BigDecimal` differently, similarly to JSON encoding. Let's look at how it encodes the number 100 for both cases:

- For long
`CBORObject.NewMap().Add(1, 100).EncodeToBytes()` returns `[-95, 1, 24, 100]`

- For BigDecimal
`CBORObject.NewMap().Add(1, BigDecimal.valueOf(100)).EncodeToBytes()` returns `[-95, 1, -60, -126, 0, 24, 100]`

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
adamsero: I tired to fix it using some suggestions from [this SO thread](https://stackoverflow.com/questions/11319445/java-to-jackson-json-serialization-money-fields), but none of them seemed to have fixed the issue. Maybe I'm doing something wrong, so you could confirm if it's actually doable.

sbernard31: let's this aside for now.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
sbernard31: Yep you're right. I don't know if this is an issue...
By default the library we used encode Bigdecimal in [CBOR Decimal Fractions and Bigfloats](https://www.rfc-editor.org/rfc/rfc7049#section-2.4.3)

What do you thing about that ?
Considering https://github.com/eclipse/leshan/issues/1304#issuecomment-1260676527, how should we encode this value for our first version ?

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
sbernard31: If there is no obvious/easy way to change this, I think we can live with it.

leshan-core/src/main/java/org/eclipse/leshan/core/util/TimestampUtil.java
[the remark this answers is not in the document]
sbernard31: Something like this : `long seconds = timestamp.setScale(0,RoundingMode.DOWN).longValueExact();` ?

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
adamsero: Using double means that we lose precision beyond 0.1 microsecond for real epoch time in seconds, since the integer part is large enough so that it can't fit all the decimal points:
- double value `1664455172.123456789` is printed out like `1.6644551721234567E9`, so we lose the two digits, 8 and 9 at the end.

Currently however, there aren't very many devices (especially the client ones used in IoT) that can produce such precise timestamps, at least not using Java. As an example:
- `Instant.now().getNano()` would return something like `989921500`
- `System.nanoTime()` would return something like `86050945362600`

As you can see, in both cases I was _de facto_ only able to obtain 7 decimal points (so up to 0.1 microsecond).

Now, I don't know if that's a good enough argument, especially since it's not future-proof. My take is that we should use what we're using right now - whether it be Bigfloat or anything granting more precision than a double as a way to enable true nanosecond precision.

adamsero:
We should be good now @sbernard31, waiting for you to bump the API version if you don't have any more issues to fix.

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
sbernard31: Do you try  : https://fasterxml.github.io/jackson-core/javadoc/2.7/com/fasterxml/jackson/core/JsonGenerator.Feature.html#WRITE_BIGDECIMAL_AS_PLAIN ?

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
sbernard31: Ok let's using this for now.
I will create a dedicated issue about all of this to centralize all idea in one place.

leshan-core/src/main/java/org/eclipse/leshan/core/util/TimestampUtil.java
```diff
@@ -0,0 +1,54 @@
...
+import java.math.RoundingMode;
+import java.time.Instant;
+
+/**
+ * <p>
+ * This class contains methods for converting timestamps between <code>BigDecimal</code> and <code>Instant</code>
+ * format.
+ * </p>
+ */
+public class TimestampUtil {
```
sbernard31, line 13: Maybe we can add some small unit test about it ? 

(especially to test the case of out of range bigdecimal ?)

sbernard31 (changes requested):

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
sbernard31: @adamsero maybe you missed that comment ?

leshan-core/src/test/java/org/eclipse/leshan/core/util/TimestampUtilTest.java
```diff
@@ -0,0 +1,72 @@
+package org.eclipse.leshan.core.util;
```
sbernard31, line 1: It misses header in this file.
This is maybe something we should check at build time, I don't know if this is possible :thinking:

leshan-core/src/test/java/org/eclipse/leshan/core/util/TimestampUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+        BigDecimal tooPreciseValue = BigDecimal.valueOf(seconds)
+                .add(BigDecimal.valueOf(picoseconds).divide(BigDecimal.valueOf(1E12)));
+
+        Instant convertedValue = TimestampUtil.fromSeconds(tooPreciseValue);
+
+        Assert.assertEquals(seconds, convertedValue.getEpochSecond());
+        Assert.assertNotEquals(picoseconds, convertedValue.getNano());
+        Assert.assertEquals(nanoseconds, convertedValue.getNano());
+    }
```
sbernard31, line 60: Should we raise an exception instead ?

sbernard31:
> My question is, can we possibly be done with this PR and M9 by the end of the week? It would be really good timing if we could.

We can try.

leshan-core/src/test/java/org/eclipse/leshan/core/util/TimestampUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    public void test_instant_to_big_decimal_conversion() {
+        List<Instant> inputList = Arrays.asList(null, Instant.EPOCH, Instant.MIN, Instant.MAX,
+                Instant.ofEpochSecond(1664832119, 123456789));
+
+        List<BigDecimal> expectedOutputList = Arrays.asList(null, BigDecimal.ZERO,
+                BigDecimal.valueOf(-31557014167219200L),
+                BigDecimal.valueOf(31556889864403199L).add(BigDecimal.valueOf(0.999999999)),
+                BigDecimal.valueOf(1664832119L).add(BigDecimal.valueOf(0.123456789)));
```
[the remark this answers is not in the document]
adamsero: I wish I could, but due to `double` precision limitations, a `BigDecimal` constructed like that gets rounded to `31556889864403200` so it has to stay that way.

leshan-core/src/test/java/org/eclipse/leshan/core/util/TimestampUtilTest.java
sbernard31: I understand your point of view.
But I also think this could also hide issue, and so generally I prefer to just raise exception by default to show there is something not expected and then let user decide, how to handle this : 
 - stop using too much precision at client side.
 - lost precision at server side.
 - or anything else...

leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeEncoderTest.java
adamsero: I managed to finally get it working, it was a case of setting `JsonNodeFactory` to use the `withExactBigDecimals` flag.

sbernard31: > I managed to finally get it working, it was a case of setting JsonNodeFactory to use the withExactBigDecimals flag.

Good to know :)


I see you change every JSON serializer, I feel this is maybe a bit out of topic.
I think it should be better to limit the change to SenML-JSON and old JSON serializer for this PR.
(I'm not even sure we want to change it for other serializer)

adamsero: Sure, I reverted the changes for all but 3 serializers

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/servlet/json/JacksonLwM2mNodeSerializer.java
```diff
@@ -85,6 +85,7 @@ public void serialize(LwM2mNode src, JsonGenerator gen, SerializerProvider provi
 
         }
 
+        gen.enable(JsonGenerator.Feature.WRITE_BIGDECIMAL_AS_PLAIN);
```
sbernard31, line 88: I guess this one is not expected ?

sbernard31 (changes requested):

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/servlet/json/JacksonLwM2mNodeSerializer.java
adamsero: Sorry, there were a lot of those serializer classes so I must have missed one. I'll reorganize the commits so we have 3 instead of 10 once everything's fine.

sbernard31:
I think we are good now ?

I let you deal with `Maven / Code Check` github action and upgrade the Android level API.

Then maybe you could rebase on `master and, squash all commits in one ? (then push force)

**Eventually** you can Android level API change in a separated commit, if you do that then this android level api change commit should be before the modification. (but again it's ok if you squash all in 1)

Last point, maybe you can refer #1304 issue in your commit header.

adamsero:
Should be good to go. Good to finally see a green checkmark 🙂

sbernard31:
Thx for the contribution :pray: !

Issue #1325, opened 2022-10-18
Data Value in SenML format is not encoded in URL-safe way


adamsero:
Paragraph _4.3. SenML Labels_ of [SenML specification](https://www.rfc-editor.org/rfc/rfc8428.html#section-4.3) that describes all possible SenML fields tells us that: 

> Data Value is a base64-encoded string with the URL-safe alphabet as defined in Section 5 of [RFC4648], with padding omitted.

In Leshan we only use Data Value (`vd`) field for representing OPAQUE format, specifically in `SenMLJsonRecordSerDes` class, but we encode the data into a URL-unsafe string. The fix is really simple and involves changing one line of code:

`jsonObj.put("vd", Base64.encodeBase64String(record.getOpaqueValue()));`

into

`jsonObj.put("vd", Base64.encodeBase64URLSafeString(record.getOpaqueValue()));`

sbernard31:
Good catch, It seems you totally right. 
We should fix this, do you plan to create a PR ?

sbernard31:
I didn't look at this but I guess this only affect decoding ?

adamsero:
> I didn't look at this but I guess this also affect decoding ?

There's only one method for decoding, namely `Base64.decodeBase64`, but it takes care of both URL safe and unsafe format. Just to be sure I wrote a little test that encodes every possible byte value:

```java
byte[] sourceData = new byte[256];

int index = 0;
for(int val = -128; val < 128; val++) {
    sourceData[index++] = (byte) val;
}

String base64UrlUnsafe = Base64.encodeBase64String(sourceData);
String base64UrlSafe = Base64.encodeBase64URLSafeString(sourceData);

byte[] base64UrlUnsafeDecoded = Base64.decodeBase64(base64UrlUnsafe);
byte[] base64UrlSafeDecoded = Base64.decodeBase64(base64UrlSafe);

assertArrayEquals(sourceData, base64UrlUnsafeDecoded);
assertArrayEquals(sourceData, base64UrlSafeDecoded);
```

and decoding indeed works fine for both cases.

sbernard31:
> There's only one method for decoding, namely Base64.decodeBase64, but it takes care of both URL safe and unsafe format.

I would prefer that  we didn't accept unsafe format by default.
But not the first time we face limitation about current Base64 implementation. 
See :  https://github.com/eclipse/leshan/issues/678 and especially https://github.com/eclipse/leshan/pull/676#issuecomment-481188756

[...]

Issue #1319, opened 2022-10-06
Plan for  2.0.0-M9 release.


sbernard31:
A 2.0.0-M9 released is required by @adamsero. 
Ideally for the end of the week. (https://github.com/eclipse/leshan/pull/1313#issuecomment-1266648467)

Current modification are : 
```
140f8b893: Remove Unused local variable and add checkstyle rule about that.
2fdf9f17f: #1304 Refactored timestamps to use BigDecimal and Instant
d4939f091: Updates demo frontend dependencies
84a055bba: Upgrade jackson depndencies (2.13.3 => 2.13.4)
e6af61d53: Upgrade maven plugins dependencies
f5accbcf1: Integration of Californium 3.7.0
ecc113c43: #1305 : Added toString, hashCode and equals to ObserveCompositeRequest
d3cf1a2aa: Timestamps in seconds instead of milliseconds in ManualDataSender
b14a7406f: #1301: Fix relative timestamp issue of SenML decoder.
a9e1741a3: #1298: fix issue about EC Key Pair with java 15 or higher
eb6f51c35: Add "write checks" permissions for test reports github action
f0d72633d: Update SECURITY.md to add CVE-2022-2576 issue.
009aab986: no checkstyle or java import check if code doesn't compile
d31212d36: No github action on Draft PR.
fbc9f75f7: add requireUpperBoundDeps rules to maven-enforcer-plugin.
c0a73763d: Revapi should not alway check API on v1.0.0 but on the last released.
```

@adamsero :
- is there something more we need to add ?
- do you (or orange) need time to test current master in your product to be sure it's fine to you ?

One more details, I see a **dependabot bot** altert about current Jackson library, I think we are not concerned as we don't use `UNWRAP_SINGLE_VALUE_ARRAYS` setting but maybe better to upgrade anyway. The _"problem"_ : there is only a rc1 version available with the fix. So I don't know if we should wait, or  ? (See : https://github.com/FasterXML/jackson-databind/issues/3590#issuecomment-1269625267)

adamsero:
> - is there something more we need to add ?

I think we're good to go

> - do you (or orange) need time to test current master in your product to be sure it's fine to you ?

The thing is we need the release to test it in production 🙂

About the library, if it's not anything serious or critical, I'd say we can wait for 2.14.0 and bump it in the next release but for now if we could pull off M9 by the end of the week it would be fantastic.

Thanks again for being helpful with the quick release 🙏

[...]

PR #1326, opened 2022-10-18
Change data value encoding to Base64 URL-safe

adamsero:


adamsero:
PR for #1325

sbernard31:
1. some concern about decoding, we started to discuss about it at : https://github.com/eclipse/leshan/issues/1325#issuecomment-1282437770

2. I think we should add tests about this (integration or unit test)

3. About issue reference in commit message, until recently I was using `#1325` but it is sometime painful because if you start the commit header with # sometime it can be consider as a commented line (see https://stackoverflow.com/questions/2788092/start-a-git-commit-message-with-a-hashmark#2788642) So recently I decide to use `GH-1325` instead of `#1325` in commit message. (see https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#issues-and-pull-requests).
If you think this is a good idea, please replace `eclipse#1325` by `GH-1325`.

leshan-core/src/main/java/org/eclipse/leshan/core/util/base64/DefaultBase64Encoder.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public String encode(byte[] dataToEncode) {
+        Base64.Encoder encoder = urlSafe ? Base64.getUrlEncoder() : Base64.getEncoder();
+        if (withoutPadding) {
+            encoder = encoder.withoutPadding();
+        }
+
+        return encoder.encodeToString(dataToEncode);
+    }
```
sbernard31, line 37: maybe better to create the encoder internally at constructor time ?
(If you agree this must be also change for Decoder)

leshan-core/src/main/java/org/eclipse/leshan/senml/json/jackson/SenMLJsonRecordSerDes.java
```diff
@@ -105,7 +106,7 @@ else if (value instanceof Float) {
                 jsonObj.put("vlo", record.getObjectLinkValue());
                 break;
             case OPAQUE:
-                jsonObj.put("vd", Base64.encodeBase64String(record.getOpaqueValue()));
+                jsonObj.put("vd", new DefaultBase64Encoder(true, true).encode(record.getOpaqueValue()));
```
sbernard31, line 109: Should we store the encoder in an attribute ?
(same for decoder ?)

leshan-core/src/main/java/org/eclipse/leshan/senml/json/jackson/SenMLJsonRecordSerDes.java
```diff
@@ -171,7 +172,7 @@ public SenMLRecord deserialize(JsonNode o) throws JsonException {
 
         JsonNode vd = o.get("vd");
         if (vd != null && vd.isTextual()) {
-            record.setOpaqueValue(Base64.decodeBase64(vd.asText()));
+            record.setOpaqueValue(new DefaultBase64Decoder(true, true).decode(vd.asText()));
```
sbernard31, line 175: Should we catch the `IllegalArgumentException` and raise a `JsonException` instead ?

leshan-core/src/main/java/org/eclipse/leshan/senml/json/jackson/SenMLJsonJacksonEncoderDecoder.java
```diff
@@ -75,7 +75,7 @@ public SenMLPack fromSenML(byte[] jsonString) throws SenMLException {
             }
             return new SenMLPack(serDes.deserialize(node.iterator()));
 
-        } catch (JsonException | IOException e) {
+        } catch (JsonException | IllegalArgumentException | IOException e) {
```
sbernard31, line 78: (If we encapsulate `IllegalArgumentException` in  `JsonException`   we should revert this change)

sbernard31 (commented):
I think we should add some test about base64 decoder /encoder.

Maybe with case raised at : https://github.com/eclipse/leshan/pull/676#issuecomment-481188756

leshan-core/src/main/java/org/eclipse/leshan/senml/json/jackson/SenMLJsonRecordSerDes.java
adamsero: We actually discussed it in #1348 in [this comment](https://github.com/eclipse/leshan/issues/1348#issue-1447767894):

>And finally in our custom version of SenMLJsonRecordSerDes, we basically have 2 lines of code to change with respect to the Leshan version (related to OPAQUE fields encoding and decoding):

```java
case OPAQUE:
    jsonObj.put("vd", urlSafeEncoder.encode(record.getOpaqueValue()));
    break;
```
>and
```java
JsonNode vd = o.get("vd");
if (vd != null && vd.isTextual()) {
    record.setOpaqueValue(urlSafeDecoder.decode(vd.asText()));
    hasValue = true;
}
```

So that would require better customization of `SenMLJsonRecordSerDes` class and others.

sbernard31: I'm not sure I get you :thinking: 

My point was just to avoid a decoder(or encoder) object creation by value to encode or decode.

I'm not sure this is related to better customization ? 

Or maybe you mean : "I will change this later anyway so we don't need to clean this now ?"

sbernard31:
You didn't answer to : 

> I think we should add some test about base64 decoder /encoder.
> Maybe with case raised at : https://github.com/eclipse/leshan/pull/676#issuecomment-481188756

So I  don't know if you plan to do it or not ?  :slightly_smiling_face:

leshan-core/src/main/java/org/eclipse/leshan/core/util/base64/DefaultBase64Encoder.java
```diff
@@ -12,27 +12,28 @@
...
  *     Orange Polska S.A. - initial API and implementation
  *******************************************************************************/
-
 package org.eclipse.leshan.core.util.base64;
 
 import java.util.Base64;
 
 public class DefaultBase64Encoder implements Base64Encoder {
     private final boolean urlSafe;
     private final boolean withoutPadding;
```
sbernard31, line 21: attributes seems useless now ?

leshan-core/src/main/java/org/eclipse/leshan/core/util/base64/DefaultBase64Decoder.java
```diff
@@ -19,10 +19,12 @@
 public class DefaultBase64Decoder implements Base64Decoder {
     private final boolean urlSafe;
     private final boolean withoutPadding;
+    private final Base64.Decoder decoder;
 
     public DefaultBase64Decoder(boolean urlSafe, boolean withoutPadding) {
         this.urlSafe = urlSafe;
```
sbernard31, line 25: urlSafe attribute seems useless now.

leshan-core/src/main/java/org/eclipse/leshan/core/util/base64/DefaultBase64Decoder.java
```diff
@@ -39,8 +41,6 @@ public String decode(byte[] encoded) {
         if (!this.withoutPadding) {
```
sbernard31, line 41: you still create a new decoder on decode.

sbernard31 (changes requested):

adamsero:
@sbernard31 Just want to let you know that @Warmek will finish this pull request as well as any future tasks as I'm leaving Orange and the project. Thank you for your time spent on helping me learn the code and the project overall 🙂.

sbernard31:
@adamsero, never a good news to see contributor go. I hope this is mainly a desired choice from you.
Thanks for all your contributions and good luck for the future. :pray: 

If you have any feedback about how we worked together on the project (good or bad things) please do not hesitate to share. This could help to improve.

sbernard31:
@Warmek do no hesitate to ask if you face any issue / difficulty on this :wink:

leshan-core/src/main/java/org/eclipse/leshan/core/util/base64/DefaultBase64Decoder.java
Warmek: I don't understand where do i create new decoder on decode?

sbernard31: Comment review seems to not be attached to the right line, so this doesn't help to understand :grimacing: 

In this class, there is 2 `decode` methods.
1 method reuses decoder created in constructor, the other recreate a new decoder on each `decode` call.

Decoder should be create only once in constructor for both methods.

Tell me if this is not clearer enough.

Warmek: As far as i can tell the only new instances are two new strings that are created in this method. I can't see where in this method could new decoder be created 🤔
```java
    @Override
    public String decode(byte[] encoded) {
        if (!this.withoutPadding) {
            validateEncodedData(new String(encoded));
        }
        byte[] decoded = decoder.decode(encoded);
        return new String(decoded);
    }
```

sbernard31: This is in the other method. 
```diff
public byte[] decode(String encoded) {
        if (!this.withoutPadding) {
            validateEncodedData(encoded);
        }
--        Base64.Decoder decoder = urlSafe ? Base64.getUrlDecoder() : Base64.getDecoder();
        return decoder.decode(encoded);
}
```
(note that I was thinking that  `Base64.getUrlDecoder()` and` Base64.getDecoder() ` created a new decoder but this is a static reference. Anyway, I think it's more consistent to reuse the attribute create in constructor.

Maybe could even be replaced by : 
```diff
public byte[] decode(String encoded) {
--        if (!this.withoutPadding) {
--            validateEncodedData(encoded);
--        }
--       Base64.Decoder decoder = urlSafe ? Base64.getUrlDecoder() : Base64.getDecoder();
--        return decoder.decode(encoded);
++   return decode(encoded.getByte());
}
```

Not directly linked but I just see that Base64.decode(String) consider that String is encoded in ISO_8859_1 which is very strange to me. I will try to understand why...

sbernard31:
@Warmek let's me know when this is in a reviewable state :)

Warmek:
@sbernard31 I think i've done it and it's ready for your review

leshan-core/src/main/java/org/eclipse/leshan/senml/json/jackson/SenMLJsonRecordSerDes.java
Warmek: I think it depends if we will call decoders/encoders with different options often or not. I probably would lean towards:
```java
jsonObj.put("vd", Base64.encodeBase64String(record.getOpaqueValue()));
```
But if we aim to add more options for  decoders/encoders or change those options often, we should probably stick to creating new decoders/encoders.

sbernard31: We want to be able to let user creates a `SenMLJsonRecordSerDes` with a defirent base64 decoder than the one by default. (discussion at #1348)

But I don't think we want to be able to change `SenMLJsonRecordSerDes` behavior dynamically.

I mean once we create a `SenMLJsonRecordSerDes` with a given base64 decoder we don't change it.


So my idea was to create an attribute initialized with `new DefaultBase64Encoder(true, true)` then use it here.

Warmek: In that case i would initialize  decoder once and then reuse it

sbernard31 (approved):

sbernard31:
@Warmek, I will take some time to get deeper about the charset encoding question. (just to be sure I well understand the problem)

Then I plan to work a bit more on this : 
- removing `org.eclipse.leshan.core.util.Base64` class
- maybe add more test and validation in base64 codec.
- maybe add a decoder which is able to support all format for LWM2M old-json  Content Format. 

I let you know when I have code to review.

Warmek (approved):
👍

Warmek:
@sbernard31 Do you think it will be possibile to share leshan M10 in first half of december?

sbernard31:
> Do you think it will be possibile to share leshan M10 in first half of december?

I'm not sure what means "share" ? 
If you talk about a release, we can try but this means we have just 1 week more left.

And after my proposition at : https://github.com/eclipse/leshan/issues/1319#issuecomment-1271378134
You (at Orange) asked in an email for more time to test before the release : 
> We as Orange would prefer to know a week ahead before a milestone is released so we can test all the features for possible bugs.

You (Orange) also proposes : 
>  Maybe we could create a separate issue for that purpose or organize it differently if you'd like.

So I created an issue talk about 2.0.0-M10 release (as I did for 2.0.0-M9) : #1370

Warmek (approved):
👍

Warmek:
@sbernard31 I reviewed your changes and i wait for merge

sbernard31:
This is now integrated in #1371, so I close this PR.

Thx @adamsero and @Warmek  for your contributions :pray: !

JaroslawLegierski:
> You (at Orange) asked in an email for more time to test before the release :
> 
> > We as Orange would prefer to know a week ahead before a milestone is released so we can test all the features for possible bugs.

Yes our colleagues are ready to test all new features - as soon as they are available in the master branch.

> Thx @adamsero and @Warmek for your contributions 🙏 !

First of all, we thank You for Your help !

sbernard31:
> First of all, we thank You for Your help !

Glad to help :slightly_smiling_face:

sbernard31:
@Warmek  @JaroslawLegierski 

I fall on this last week : https://www.eclipse.org/projects/handbook/#resources-commit

> The email address of the author must match the email address on the Eclipse Foundation Account, the name "Some Body" must be replaced by the real name of the person.

:point_up:  this is not checked by any eclipse validation tools  but would be better if you change your git configuration to match the eclipse requirement. 

Currently @Warmek  `Author` looks like  `Author: Warmek <b***********@orange.com>` (which is not so good)
and @JaroslawLegierski  looks like this : `Author: JaroslawLegierski <j*********@orange.com> ` (which is almost OK but maybe better with a space)

I let you know but I think this is clearly not critical.

JaroslawLegierski:
> and @JaroslawLegierski looks like this : `Author: JaroslawLegierski <j*********@orange.com> ` (which is almost OK but maybe better with a space)

I modified the Name field in the Github configuration -> now is in the form FirstName space LastName. I'll have to see what is the effect of these changes.

Warmek:
I've changed my user.name via command: 
git config --global user.name "Some Body" 

Will that help?

sbernard31:
did you commit your last commit (d0861f804693083e9503733b7366aa379569cd81) before or after running this ?
(because in your last commit there is still `Author: Warmek <b***********@orange.com>`

Warmek:
I've made my last commit before running this command

sbernard31:
OK when you will push a new commit, I will be able to say you if this is OK now.  :slightly_smiling_face:

Issue #534, opened 2018-07-13
Add support of write attributes at client side.


sbernard31:
This means : 
1. be able to handle write request attributes and store attributes
2. use this attributes to raise notification 
3. show attributes in Discover response

There is an ongoing work about this (#484).

rkimsb2:
+1

henris42:
Hi, any news on this one? I have an urgent need to somehow handle PMAX/PMIN attribute requests in the client. The client now answers with 500 Internal Error, which is not tolerated by our server - we cannot observe anything.

I'm happy also with ANY workaround, PMAX/PMIN do not need to actually work in this application, we just need to get the server the expected response (204 I believe)

sbernard31:
Clearly not a priority for us now.
This could need some modification at californium side, meaning this could be a not so easy task.
The only hope : an unintended contribution. So I won't bet on this at short term.

**Anyway a possible workaround**:
You need a custom `ObjectEnabler` which do what you want on `WriteAttributesRequest`
E.g : 
``` java
public class CustomObjectEnabler extends ObjectEnabler {

    public CustomObjectEnabler(int id, ObjectModel objectModel, Map<Integer, LwM2mInstanceEnabler> instances,
            LwM2mInstanceEnablerFactory instanceFactory, ContentFormat defaultContentFormat) {
        super(id, objectModel, instances, instanceFactory, defaultContentFormat);
    }

    @Override
    public synchronized WriteAttributesResponse writeAttributes(ServerIdentity identity,
            WriteAttributesRequest request) {
        return WriteAttributesResponse.success();
    }
}
```

Then if you are using `ObjectsInitializer` you can create a custom one too : 
```java
public class CustomObjectsInitializer extends ObjectsInitializer {

    @Override
    protected ObjectEnabler createNodeEnabler(ObjectModel objectModel) {
        Map<Integer, LwM2mInstanceEnabler> instances = new HashMap<>();
        LwM2mInstanceEnabler[] newInstances = createInstances(objectModel);
        for (LwM2mInstanceEnabler instance : newInstances) {
            // set id if not already set
            if (instance.getId() == null) {
                int id = BaseInstanceEnablerFactory.generateNewInstanceId(instances.keySet());
                instance.setId(id);
            }
            instance.setModel(objectModel);
            instances.put(instance.getId(), instance);
        }
        return new CustomObjectEnabler(objectModel.id, objectModel, instances, getFactoryFor(objectModel),
[...]

PR #1327, opened 2022-10-19
client:add writeAttributes support

dellzhui:
Add support for writeAttributes


sbernard31:
I don't know exactly what is your plan here. I looked at the code and I'm not sure to get the idea.

But maybe better to create an issue  (or reuse https://github.com/eclipse/leshan/issues/534) to discuss about that before.

Issue #1339, opened 2022-10-26
SLF4J 2.0.X?


jvermillard:
Is there a reason to keep master on SLF4J 1.7.X? For example eclipse jetty 11 uses slf4j 2.0 now

See: https://www.slf4j.org/faq.html#changesInVersion200

jvermillard:
I tried to upgrade :joy: 
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for leshan 2.0.0-SNAPSHOT:
[INFO] 
[INFO] leshan - core ...................................... SUCCESS [ 12.051 s]
[INFO] leshan - core californium .......................... FAILURE [  2.892 s]
[INFO] leshan - server core ............................... SKIPPED
[INFO] leshan - server californium ........................ SKIPPED
[INFO] leshan - server redis .............................. SKIPPED
[INFO] leshan - client core ............................... SKIPPED
[INFO] leshan - client californium ........................ SKIPPED
[INFO] leshan - integration tests ......................... SKIPPED
[INFO] leshan - core demo ................................. SKIPPED
[INFO] leshan - client demo ............................... SKIPPED
[INFO] leshan - server core demo .......................... SKIPPED
[INFO] leshan - server demo ............................... SKIPPED
[INFO] leshan - bootstrap server demo ..................... SKIPPED
[INFO] leshan ............................................. SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  15.914 s
[INFO] Finished at: 2022-10-26T16:29:43+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.revapi:revapi-maven-plugin:0.14.7:check (default) on project leshan-core-cf: The following API problems caused the build to fail:
[ERROR] java.class.externalClassExposedInAPI: enum org.slf4j.event.Level: A class from supplementary archives is used in a public capacity in the API. [org.slf4j:slf4j-api:jar:2.0.3]
[ERROR] java.class.externalClassExposedInAPI: interface org.slf4j.spi.LoggingEventBuilder: A class from supplementary archives is used in a public capacity in the API. [org.slf4j:slf4j-api:jar:2.0.3]
[ERROR] 
[ERROR] Consult the plugin output above for suggestions on how to ignore the found problems.

```

sbernard31:
No a good one :wink: 

[...]

PR #1340, opened 2022-10-27
upgrade to slf4j 2

jvermillard:
See GH-1339


sbernard31:
(If this is OK, I just wait a little bit to know if we will have a revapi fix soon)

sbernard31:
I rebased it on master.
+ add comment about why we exclude slf4j
+ rename commit to refer to https://github.com/eclipse/leshan/issues/1339
+ move jetty update in separated commit.

As the revapi will probably not be fixed soon. I will integrate this in `master`.

Issue #1352, opened 2022-11-16
leshan demo client incorrect initialization of object 3442, resources 120, 1120


davidahoward:
the current release tags M8, M9, and the master branch as of 16 November 2022 incorrectly initialize the Eclipse LwM2M v1.1 Test Object 3442 integer resources 120, 1120 to a value of 64, when it is supposed to be 1024.

120 | Integer Value | RW | Single | Optional | Integer |   |   | Initial value must be "1024".
1120 | Multiple Integer Value | RW | Multiple | Optional | Integer |   |   | Initial value must be 1 instance with ID 0 and value "1024".

sbernard31:
Good catch :+1: 

Same issue with Leshan v1.x : 
https://github.com/eclipse/leshan/blob/f62bfad308d31f2b5c5d63b905f574536daef429/leshan-client-demo/src/main/java/org/eclipse/leshan/client/demo/LwM2mTestObject.java#L34-L50

sbernard31:
Fixed in `master` for object 3442  by #1356 (commit 5877f57d87e81137b5dba40da4cdd03738c3651d)

Fixed in `1.x` for object 3441 by commit 938e43d911ce4dfdbc62b4a082e33869c4c9c4d0.

Thx again @davidahoward for reporting this. :pray:

PR #1353, opened 2022-11-16
Fix leshan demo client incorrect initialization of object 3442,

davidahoward:
resources 120, 1120 #1352

updated public class LwM2mTestObject,
public static final Long INITIAL_INTEGER_VALUE = 1024l;


davidahoward:
I've signed the ECA, not sure why the workflow still shows it...

PR #1354, opened 2022-11-17
Fix leshan demo client incorrect initialization of object 3442,

davidahoward:
resources 120, 1120 #1352

updated public class LwM2mTestObject,
public static final Long INITIAL_INTEGER_VALUE = 1024l;


davidahoward:
where is it getting this email address from?  It's not the one I am using
Anonymous (david.howard@****.com)

Issue #1374, opened 2023-01-02
calling LeshanServer::destroy() with queue mode disable generate an NPE


jvermillard:
### Version(s)

master @ 4527744910d2ecd67e576bdf7d0950e99d1d1497

### Which components

leshan-server-cf

### What happened

java.lang.NullPointerException: Cannot invoke "org.eclipse.leshan.server.queue.PresenceServiceImpl.destroy()" because "this.presenceService" is null

### How to reproduce

Disable queue mode using: `builder.disableQueueModeSupport();`
create the server and  destroy it using  using  LeshanServer::destroy()  method

If you look at: 
https://github.com/eclipse/leshan/blob/master/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServer.java#L219
w ith set presence server to null when queuemode is disabled then there is an obvious bug there:

https://github.com/eclipse/leshan/blob/master/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServer.java#433

The presenseServer must be tested for null

sbernard31:
You point to same URL twice but I guess you mean this is obvious that there is a bug at : 
https://github.com/eclipse/leshan/blob/master/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServer.java#L433

Good catch.

I will review your PR and see if we need same kind of fix for branch `1.x`.

sbernard31:
The fix is integrated in `master` (#1375)  and in `1.x` branch (c1123be7f8648769fa5f3171564f33602d44c0a3)

PR #1375, opened 2023-01-02
fix NPE destroying queue mode disable server

jvermillard:
Fix: https://github.com/eclipse/leshan/issues/1374


leshan-server-cf/src/test/java/org/eclipse/leshan/server/californium/LeshanServerTest.java
```diff
@@ -81,11 +81,34 @@ public void testStartStopDestroy() throws InterruptedException {
...
+        // HACK force creation thread creation.
+        forceThreadsCreation(server);
+        Thread.sleep(100);
+        server.stop();
+        Thread.sleep(100);
+        server.destroy();
+
+        // ensure all thread are destroyed
+        Thread.sleep(500);
+        assertEquals("All news created threads must be destroyed", numberOfThreadbefore, Thread.activeCount());
```
sbernard31, line 102: My first idea would have been to not check all the counting thread part (just a simple create/start/stop/destroy), but thinking a bit more about it I don't know what is the best way :shrug: 

I guess we can keep it like this :thinking:, any opinion ?

sbernard31 (approved):

sbernard31:
I renamed the commit header to refer to #1374.

leshan-server-cf/src/test/java/org/eclipse/leshan/server/californium/LeshanServerTest.java
jvermillard: I just copied the other tests to keep the same style, and all the other tests are checking threads, so I thought it was something to check in this specific config. Anyway, I don't have a strong opinion on that since I didn't write those tests :smiley:

sbernard31: >  all the other tests are checking threads

Almost all, the first one just checks just start/stop/start without thread counting.

I don't find any really good reason to remove or keep it. So I guess we can let it like this :sweat_smile:

Issue #1376, opened 2023-01-12
Rethrowing duplication errors as CodecException


Warmek:
### Question

When a SEND containing a duplicates (same path + timestamp) is received, TimestampedLwM2mNodes.Builder::build throws an IllegalArgumentException.
```java
if (noDuplicate && previous != null) {
    throw new IllegalArgumentException(String.format(
             "Unable to create TimestampedLwM2mNodes : duplicate value for path %s. (%s, %s)",
             internalNode.path, internalNode.node, previous));
}

```
That exception is just catched in SendResource::handlePOST, to be sent to the SendListener::onError callback and rethrown.
```java
} catch (RuntimeException e) {
    sendHandler.onError(registration, e);
    throw e;
}
```

Maybe it would be better to catch that exception (in LwM2mNodeSenMLDecoder::decodeTimestampedNodes) and rethrow as CodecException?
```java
} catch (SenMLException | IllegalArgumentException e) {
        String hexValue = content != null ? Hex.encodeHexString(content) : "";
        throw new CodecException(e, "Unable to decode nodes : %s", hexValue, e);
}
```

sbernard31:
Yep totally agree ! Good catch :+1:  !

Warmek:
I've made a commit on [opl/rethrow_illegalArgumentException](https://github.com/eclipse/leshan/compare/master...JaroslawLegierski:leshan:opl/rethrow_illegalArgumentException)

sbernard31:
Yep I think you can create a PR

sbernard31:
Fix in `master` by #1377.

Thx @Warmek for catching this :wink: !

PR #1377, opened 2023-01-13
SenMLDecoder should raise CodecException instead of IllegalArgumentException

Warmek:
Re-throw IllegalArgumentException as CodecException

It aims to implement #1376


sbernard31:
It sounds there is something wrong. 5 commits to just add 1 word :slightly_smiling_face:

sbernard31:
Integrated in `master` (commit 6a1ff2dad57d3513a4151e83e4d3d3f7032fc632)

PR #1378, opened 2023-01-16
Check Object Enabler Consistency at Client Creation

sbernard31:
The idea is to provide better feedback to user, if they try to create client without valid Object Enablers.

This idea was triggered by https://github.com/sbernard31/benchmark-clients/issues/8

The is implemented in this PR by reusing BootstrapConsistencyChecker which is mainly responsible to check the consistency state of the client after at the end of a bootstrap session. I guess it could also make sense to check it at creation ?

E.g. of error you can get at creation : 
```
Exception in thread "main" java.lang.IllegalArgumentException: Invalid 'ObjectEnabler' Setting : 
 - Client MUST have ObjectEnabler for 'Server' Object (ID:1)
 - Client MUST have ObjectEnabler for 'Device' Object (ID:3)
	at org.eclipse.leshan.client.LeshanClient.<init>(LeshanClient.java:97)
	at org.eclipse.leshan.client.LeshanClientBuilder.createLeshanClient(LeshanClientBuilder.java:329)
	at org.eclipse.leshan.client.LeshanClientBuilder.build(LeshanClientBuilder.java:294)
	at TestClient.main(TestClient.java:32)
```

### DIFF
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/bootstrap/BaseBootstrapConsistencyChecker.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/bootstrap/BaseBootstrapConsistencyChecker.java
@@ -15,8 +15,13 @@
  *******************************************************************************/
 package org.eclipse.leshan.client.bootstrap;
 
+import static org.eclipse.leshan.core.LwM2mId.DEVICE;
+import static org.eclipse.leshan.core.LwM2mId.DVC_SUPPORTED_BINDING;
+import static org.eclipse.leshan.core.LwM2mId.SECURITY;
+import static org.eclipse.leshan.core.LwM2mId.SERVER;
+
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


leshan-client-core/src/main/java/org/eclipse/leshan/client/bootstrap/BaseBootstrapConsistencyChecker.java
```diff
@@ -37,24 +43,79 @@ public abstract class BaseBootstrapConsistencyChecker implements BootstrapConsis
...
+            return errors;
+        } else {
+            return null;
+        }
+    }
+
+    /**
+     * Check mandatory object enabler are present
+     */
+    protected void checkMandatoryObjectEnabler(Map<Integer, LwM2mObjectEnabler> objectEnablers, List<String> errors) {
```
jvermillard, line 71: Why not check all objects tagged as mandatory in the model?

jvermillard: fine :+1:  then yeah a comment would be welcome since the term "mandatory" could create confusion with the constraint from the object spec

sbernard31: Done :)

sbernard31:
@jvermillard can I integrate this in `master`, or you want to review that a bit more ?

jvermillard (approved):

Issue #1373, opened 2022-12-19
Experiment Transport layer with java-coap (fork)


sbernard31:
To test new transport layer (#1025),  we decide to experiment it a little with : [**java-coap (fork)**](https://github.com/open-coap/java-coap). (see : https://github.com/eclipse/leshan/issues/1338)

This issue aims to centralize discussion about it.

sbernard31:
@JaroslawLegierski I know you started to work about it, do you have any code to share.

You are working on client side right ?

JaroslawLegierski:
> You are working on client side right ?

Yes I have first version. I am currently able to perform registration, reg update, send data to server and deregister client. 
The code doesn't look good (it's a terrible mess) - so I'd like to clean it a bit before sharing.

sbernard31:
@JaroslawLegierski do not hesitate to share even if code is not so good.

On my side, I will also begin to work on this as I'm a bit stuck : 
- integrating transport layer abstraction : waiting for feedback from orange.
- refactor transport Layer Module following #1295 : try to get some feedback ?
- exploring coap+tcp  :  waiting for feedback from californium to move forward.

As you are on client, I will try to look at server side :thinking:

JaroslawLegierski:
OK the code won't be fully cleaned but I'd like to at least remove all failed "test blocks" from it. Unfortunately, I spend a some time on organizational tasks lately (e.g. point 1 from the list above ;-) and generally I'm a little late in this week.

sbernard31:
(take your time, no pressure :wink:)

sbernard31:
I started to code something in [`java-coap`](https://github.com/eclipse/leshan/compare/java-coap) branch.
This is not clean at all (very inspired from what I did with californium which is maybe not a good idea ? :thinking: )
This is not tested (not ever run) but I share it just in case it could helps. => https://github.com/eclipse/leshan/commit/a45f3108c05792c2c5b781591182165b1fd3b407

JaroslawLegierski:
Please find my first tests results [here]( https://github.com/JaroslawLegierski/leshan/tree/opl/endpoint_client_javacoap) . I started by modifying the existing cf-handling classes. It seems to me that Your approach (starting from the scratch) is definitely better because I'm still using cf classes and I still have a mess in the code.

JaroslawLegierski:
[...]

Issue #1338, opened 2022-10-25
Which candidate to test transport layer ?


sbernard31:
To test abstraction Layer, it could make sense to try to implement a new transport based on something else than Californium.

One possible way is to create a new CoAP endpoint Provider based another CoAP java library.

I tried to look what is currently available :point_down: 

name | repository | last mod. | feature | maven central | alt. maven repo |
-- | -- | -- | -- | -- | --
**java-coap (fork)** | https://github.com/open-coap/java-coap | ~ some days ago | observe,  block transfer, coap+tcp, coaps+tcp, coaps (based on mbedtls)| :heavy_check_mark: | 
**java-coap** | https://github.com/PelionIoT/java-coap | ~ 1, 2 years ago | observe,  block transfer, coap+tcp, coaps+tcp | :x: | :heavy_check_mark: 
**ncoap** | https://github.com/okleine/nCoAP | ~ 5 years ago | observe,  block transfer | :x: | :heavy_check_mark: 
**coap-blaster** | https://github.com/google/coapblaster | ~ 3, 4 years ago | observe,  block2 | :heavy_check_mark:  | :x: 
[**jcoap**](http://ws4d.org/category/ws4d-jcoap/)|  https://gitlab.amd.e-technik.uni-rostock.de/ws4d/jcoap| ~ 3/4 years ago  | ? | :x:| :x:
[**mjCoAP**](http://netsec.unipr.it/project/mjcoap/)| :x: | ?  | observe,  block transfer  |:x:| :x: 

I also find there was a project to provide a **CoAP implementation in OpenJDK** known as [**Kona**](https://openjdk.org/projects/kona/) but it seems [the project is archived](https://bugs.openjdk.org/browse/JDK-8223962).   
Too bad, It could have been a simple default implementation without any dependencies and californium would have been the full feature one.

sbernard31:
Another solution is to directly target another protocol like experiment about mqtt ? #1236

sbernard31:
If we decide to test a new Java CoAP implementation, It's for now that better choice would be **[coap-java (fork)](https://github.com/open-coap/java-coap)**.

sbernard31:
Discussion about current experimentation with **coap-java(fork)** at #1373

sbernard31:
We provide now a CoAP implementation based on **java-coap(fork)**, See https://github.com/eclipse-leshan/leshan/issues/1373 for more details.

I think we can close this issue.

PR #1382, opened 2023-01-24
Add an EndpointProvider at Server side for CoAP based on java-coap(fork)

sbernard31:
This aims to experiment transport abstraction layer implementing another endpoint provider at Server side for CoAP based on **[java-coap(fork)](https://github.com/open-coap/java-coap)**.

See more details at  #1373.

### DIFF
--- a/leshan-tl-javacoap-server/pom.xml
+++ b/leshan-tl-javacoap-server/pom.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+Copyright (c) 2023 Sierra Wireless and others.
+
+All rights reserved. This program and the accompanying materials
+are made available under the terms of the Eclipse Public License v2.0
+and Eclipse Distribution License v1.0 which accompany this distribution.
+
+The Eclipse Public License is available at
+   http://www.eclipse.org/legal/epl-v20.html
+and the Eclipse Distribution License is available at
+   http://www.eclipse.org/org/documents/edl-v10.html.
+
+Contributors:
+    Sierra Wireless - initial API and implementation
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.eclipse.leshan</groupId>
+    <artifactId>lib-build-config</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+    <relativePath>../build-config/lib-build-config/pom.xml</relativePath>
+  </parent>
+  <artifactId>leshan-tl-javacoap-server</artifactId>
+  <packaging>bundle</packaging>
+  <name>leshan - server java-coap</name>
+  <description>A transport implementation for leshan server based on Java CoAP</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.eclipse.leshan</groupId>
+      <artifactId>leshan-server-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.leshan</groupId>
+      <artifactId>leshan-tl-javacoap-core</artifactId>
+    </dependency>
+  </dependencies>
+</project>
--- a/build-config/lib-build-config/pom.xml
+++ b/build-config/lib-build-config/pom.xml
@@ -97,6 +97,13 @@ Contributors:
                     <matcher>java-package</matcher>
                     <match>/org\.slf4j(\..*)?/</match>
                   </item>
+                  <item>
+                    <!-- HACK : we exclude com.mbed because of revapi bug. 
+                         See : https://github.com/revapi/revapi/issues/186
+                         This should be removed once it will be fixed in revapi. -->
+                    <matcher>java-package</matcher>
+                    <match>/com\.mbed\.coap(\..*)?/</match>
+                  </item>
                 </exclude>
               </elements>
             </revapi.filter>
--- a/leshan-integration-tests/pom.xml
+++ b/leshan-integration-tests/pom.xml
@@ -40,6 +40,10 @@ Contributors:
     </dependency>
     <dependency>
       <groupId>org.eclipse.leshan</groupId>
+      <artifactId>leshan-tl-javacoap-server</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.leshan</groupId>
       <artifactId>leshan-server-redis</artifactId>
     </dependency>
     <dependency>
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/DeleteTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/DeleteTest.java
@@ -62,7 +62,8 @@ public class DeleteTest {
     static Stream<org.junit.jupiter.params.provider.Arguments> transports() {
         return Stream.of(//
                 // ProtocolUsed - Client Endpoint Provider - Server Endpoint Provider
-                arguments(Protocol.COAP, "Californium", "Californium"));
+                arguments(Protocol.COAP, "Californium", "Californium"), //
+                arguments(Protocol.COAP, "Californium", "java-coap"));
     }
 
     /*---------------------------------/
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/DiscoverTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/DiscoverTest.java
@@ -56,7 +56,8 @@ public class DiscoverTest {
     static Stream<org.junit.jupiter.params.provider.Arguments> transports() {
         return Stream.of(//
                 // ProtocolUsed - Client Endpoint Provider - Server Endpoint Provider
-                arguments(Protocol.COAP, "Californium", "Californium"));
+                arguments(Protocol.COAP, "Californium", "Californium"), //
+                arguments(Protocol.COAP, "Californium", "java-coap"));
     }
 
     /*---------------------------------/
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/ExecuteTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/ExecuteTest.java
@@ -59,7 +59,8 @@ public class ExecuteTest {
     static Stream<org.junit.jupiter.params.provider.Arguments> transports() {
         return Stream.of(//
                 // ProtocolUsed - Client Endpoint Provider - Server Endpoint Provider
-                arguments(Protocol.COAP, "Californium", "Californium"));
+                arguments(Protocol.COAP, "Californium", "Californium"), //
+                arguments(Protocol.COAP, "Californium", "java-coap"));
     }
 
     /*---------------------------------/
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/QueueModeTest.java
@@ -58,7 +58,8 @@ public class QueueModeTest {
     static Stream<org.junit.jupiter.params.provider.Arguments> transports() {
         return Stream.of(//
                 // ProtocolUsed - Client Endpoint Provider - Server Endpoint Provider
-                arguments(Protocol.COAP, "Californium", "Californium"));
+                arguments(Protocol.COAP, "Californium", "Californium"), //
+                arguments(Protocol.COAP, "Californium", "java-coap"));
     }
 
     /*---------------------------------/
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/RegistrationTest.java
@@ -19,12 +19,9 @@
 package org.eclipse.leshan.integration.tests;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.eclipse.leshan.core.ResponseCode.CONTENT;
 import static org.eclipse.leshan.integration.tests.util.LeshanTestClientBuilder.givenClientUsing;
 import static org.eclipse.leshan.integration.tests.util.assertion.Assertions.assertArg;
 import static org.eclipse.leshan.integration.tests.util.assertion.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.assertThrowsExactly;
-import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
@@ -41,7 +38,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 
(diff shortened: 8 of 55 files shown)


leshan-tl-javacoap-server/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>org.eclipse.leshan</groupId>
+      <artifactId>leshan-server-core</artifactId>
+    </dependency>
+    
+    <dependency>
+    <groupId>com.github.open-coap.java-coap</groupId>
+    <artifactId>coap-core</artifactId>
+    <version>6.5.0</version>
```
szysas, line 49: You can update it, it is now in central, jitpack was not very reliable.
```suggestion
    <groupId>io.github.open-coap</groupId>
    <artifactId>coap-core</artifactId>
    <version>6.8.0</version>
```

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/resource/ResourcesService.java
```diff
@@ -0,0 +1,122 @@
...
+    private Service<CoapRequest, CoapResponse> findHandler(RequestMatcher requestMatcher) {
+        return prefixedHandlers.stream().filter(e -> e.getKey().matches(requestMatcher)).findFirst()
+                .map(Entry::getValue).orElse(NOT_FOUND_SERVICE);
+
+    }
+
+    public static class ResourcesBuilder {
+        private final Map<RequestMatcher, Service<CoapRequest, CoapResponse>> handlers = new HashMap<>();
+
+        public ResourcesBuilder add(String uriPath, Service<CoapRequest, CoapResponse> service) {
```
szysas, line 71: Was the only reason for this class, that you needed to add resource without specifying method? Or is there something else?

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
```diff
@@ -0,0 +1,233 @@
...
+
+        // Send LWM2M Request
+        CompletableFuture<T> lwm2mResponseFuture = sendLwM2mRequest(destination, request, lowerLayerConfig);
+
+        // Wait synchronously for LWM2M response
+        try {
+            return lwm2mResponseFuture.get(timeoutInMs, TimeUnit.MILLISECONDS);
+        } catch (CompletionException | ExecutionException exception) {
+            // TODO we probably need to enhance this (better translate java-coap exceptions to leshan ones)
+            throw new SendFailedException("Unable to send request", exception);
```
szysas, line 95: At least you could do:
```suggestion
            throw new SendFailedException("Unable to send request", exception.getCause());
```

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
```diff
@@ -0,0 +1,233 @@
...
+    @Override
+    public <T extends LwM2mResponse> T send(ClientProfile destination, DownlinkRequest<T> request,
+            LowerLayerConfig lowerLayerConfig, long timeoutInMs) throws InterruptedException {
+
+        // Send LWM2M Request
+        CompletableFuture<T> lwm2mResponseFuture = sendLwM2mRequest(destination, request, lowerLayerConfig);
+
+        // Wait synchronously for LWM2M response
+        try {
+            return lwm2mResponseFuture.get(timeoutInMs, TimeUnit.MILLISECONDS);
```
szysas, line 92: How common are blocking API like this? Java-coap was design to be fully asynchronous so blocking calls might cause some issues. Although here it is probably ok.

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
```diff
@@ -0,0 +1,233 @@
...
+                    // TODO we probably need to enhance this (better translate java-coap exceptions to leshan ones)
+                    errorCallback.onError(new SendFailedException("Unable to send request", exception));
+                    return null;
+                })
+                // Handle CoAP Response
+                .thenAccept((lwM2mResponse) -> {
+                    responseCallback.onResponse(lwM2mResponse);
+                });
+
+        // TODO handle timeout (cancel future)
```
szysas, line 123: In case of timeout, it will be `CoapTimeoutException`

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
```diff
@@ -0,0 +1,233 @@
...
+            DownlinkRequest<T> lwm2mRequest, LowerLayerConfig lowerLayerConfig) {
+
+        if (lwm2mRequest instanceof ObserveRequest) {
+            return sendObserveRequest(destination, lwm2mRequest, lowerLayerConfig);
+        } else {
+            // Create Coap Request to send from LWM2M Request
+            CoapRequest coapRequest = translator.createCoapRequest(destination, lwm2mRequest, toolbox);
+
+            // Create a Coap Client to send request
+            CoapClient coapClient = CoapClientBuilder.clientFor(destination.getIdentity().getPeerAddress(), coapServer);
```
szysas, line 136: You do not need to create coapclient just to send request. 
You can use `coapServerserver.clientService().apply(..)`

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpointsProvider.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public void stop() {
+        // TODO in Leshan stop means "we can restart after a stop", so we should check what means stop for java-coap
+        coapServer.stop();
+    }
+
+    @Override
+    public void destroy() {
+        // TODO there is no destroy, so we just stop ?
```
szysas, line 111: Yeap

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpointsProvider.java
```diff
@@ -0,0 +1,114 @@
...
+            coapServer.start();
+        } catch (IllegalStateException | IOException e) {
+            // TODO handle this correctly
+            e.printStackTrace();
+        }
+    }
+
+    @Override
+    public void stop() {
+        // TODO in Leshan stop means "we can restart after a stop", so we should check what means stop for java-coap
```
szysas, line 105: There is no restart after stop, need to create new instance to start again. 
Out of curiosity, what's the use case for restart?

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    @Override
+    public URI getURI() {
+        return endpointUri;
+    }
+
+    @Override
+    public <T extends LwM2mResponse> T send(ClientProfile destination, DownlinkRequest<T> request,
+            LowerLayerConfig lowerLayerConfig, long timeoutInMs) throws InterruptedException {
```
szysas, line 85: Is it really so, that each request can have different timeout?

leshan-tl-javacoap-server/pom.xml
sbernard31: :tada: **Available on maven central ! :tada:** 

I changed it. (0fbcfb3)

But updating I find a kind of API break between 6.5 and 6.8

`DataConvertingUtility.parseUriQuery(String)` was raising `ParseException` and now it doesn't.
Note that `DataConvertingUtility.parseUriQueryMult(String)` still raises a `ParseException`
I don't know if this is expected :shrug: 

I also updated comparison table at  https://github.com/eclipse/leshan/issues/1338.

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/resource/ResourcesService.java
sbernard31: Yep If I remember well this is the only reason. :thinking:

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
sbernard31: I consider that sync API is mainly useful to write demo, POC or tests.
A production server should rather use the async way.

But I'm curious : 
> Java-coap was design to be fully asynchronous so blocking calls might cause some issues.

which kind of issue are you in mind ?

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
sbernard31: OK I will change this.

But if I should not use it, what is the use case for `CoapClientBuilder.clientFor(destination, coapServer)`

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpointsProvider.java
sbernard31: To be honest I'm not so sure if this is really needed, I know that we use it in some tests (especially at LWM2M client side to simulate that device is sleeping).

Californium has a start / stop / destroy mechanism where stop means that's you can restart unlike destroy.
Stop is kind of "pause". 

As Leshan was strongly tied to californium, we follow this API.
Then when I did first transport abstraction layer, I keep it... 
Always easier to add feature than remove it, but maybe better to just keep start() and destroy(). :thinking:

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
sbernard31: I see at least one use case : you don't want same timeout when you send "classic" request and when you send request which involved large payload (in request or response) using blockwise transfer. 

See : https://github.com/eclipse/leshan/wiki/Request-Timeout

I guess this should be easily implemented as most transport layer library should propose a way to cancel a request. 
(says the maintainer of the library which does not provide API to cancel LWM2M request :grin:)

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
sbernard31: Thx I will do it when I will polish exception handling and timeout :pray:

leshan-tl-javacoap-server/pom.xml
szymonsasin-nordic: Yes, that was expected.

szysas: Yes, that was expected.

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
szysas: I meant that if while handling a request (for example in Filter) you would block a call, then it might get into deadlock.

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
szysas: It's just a helper wrapper. It will place destination address in every request so no need to define it. Mostly useful in tests.

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
szysas: I see, maybe I could figure out a way to pass timeout on each request, which will overwrite default timeout.

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
sbernard31: I get it but I imagine for Future about getting response it is ok, right ? I mean this is one of the future feature.

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
sbernard31: Ok done :)

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
sbernard31: I don't know if this will work but my plan was do something like : 
```
public <T> CompletableFuture<T> timeoutAfter(long timeout, TimeUnit unit) {
    CompletableFuture<T> result = new CompletableFuture<T>();
    delayer.schedule(() -> result.completeExceptionally(new TimeoutException()), timeout, unit);
    return result;
}
```
OR to look how CompletableFuture.orTimeout is implemented. 

Following wording at https://github.com/eclipse/leshan/wiki/Request-Timeout , java-coap timeout is which kind of timeout ?

szysas: I was talking about response timeout. (btw, very good explanation 👍)

leshan-tl-javacoap-server/src/main/java/org/eclipse/leshan/transport/javacoap/endpoint/JavaCoapServerEndpoint.java
szysas: Yes, in this case that's the only way.

sbernard31:
I finally succeed to pass all CoAP (only) integration tests with Californium used at LWM2M client side and java-coap at LWM2M server side.

(Probably not exhaustive) Missing feature : 
 - **composite operation** because Patch, IPatch and Fetch  is not implemented in java-coap  but should be available soon (318fe3d92dd61555ac260d8c86a4c03bdaf85960) and should not be so hard to implement. 
 - **observe** can not be shared / persisted : probably the biggest issue. https://github.com/open-coap/java-coap/issues/36
 - we can not **stop** then restart a endpoint based on java-coap.

sbernard31:
With java-coap v6.12.0, I was able to : 
- Remove some workaround using new API : ae98e567db81c1ee4ece4576fb72f5c1ca9a6e82
- Add support of **Composite operation** : b52be486ee203906f158f38c6664e7ada8339f11

Thx @szysas :pray:

sbernard31:
I rebased this in **master** (several conflict especially because of Identity Refactoring)

There is also some observe code available at [java_coap_server_observe](https://github.com/eclipse-leshan/leshan/tree/java_coap_server_observe) which depends on "not yet released version" of java-coap

sbernard31:
:heavy_check_mark: Observe relation can now be  shared / persisted, since we use **java-coap v6.18.0**.

All integration test are OK :heavy_check_mark:, there are still some TODO in the code but I think this is now in a testable state.
So, I plan to integrate in `master` soon and so it will be available in sandbox.

sbernard31:
This is now integrated in `master`.

java-coap endpoint is available in **leshan-server-demo**, see Server tab: 

Capture d’écran du 2023-10-19 12-08-55

The port and host can be configured with options: 
```
-jh, --java-coap-host=<jlocalAddress>
                  Set the local CoAP address of endpoint based on
                    java-coap library.
                  Default: any local address.
-jp, --java-coap-port=<jlocalPort>
                  Set the local CoAP port of endpoint based on java-coap
                    library.
                  Default: 5685
```

PR #1388, opened 2023-01-25
Avoid to call registrationStore.removeObservation with null registration ID

### LINKED ISSUE #1384: LwM2mObservationStore 'remove(Token token)' is not functionnal
### Version(s)

fa38111cf1d4787ead04bd04002c2df7cccc8dc1

### Which components

leshan server

### Tested With

_No response_

### What happened

LwM2mObservationStore.remove(Token token)  calls registrationStore.removeObservation(String registrationId, ObservationIdentifier observationId) : https://github.com/eclipse/leshan/blob/fa38111cf1d4787ead04bd04002c2df7cccc8dc1/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/observation/LwM2mObservationStore.java#L95-L96

But it does call it with 'null' registrationId.

And RedisRegistrationStore 'needs' a non null registrationId as it uses it as lock; and does nothing if it does not exist :
https://github.com/eclipse/leshan/blob/fa38111cf1d4787ead04bd04002c2df7cccc8dc1/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisRegistrationStore.java#L557-L564

May be the bug is more on the RedisRegistrationStore side who could allow a null registrationId and still remove the observationId.

### How to reproduce

Create a 'bad' Observation (e.g. unknown path) with a server using RedisRegistrationStore : the observation is not deleted from the store.

In this use case, the BaseMatcher.ObservationObserverAdapter.onResponse callback from californium library will call the LwM2mObservationStore.remove(Token token) method (due to the if (response.isError()...) l.332) wich will trigger no deletion as explained.

As a consequence, the Observation will still be registered in the store whereas it should have been deleted.

### Relevant Output

_No response_

sbernard31:
This aims to fix #1384

But working on this, I also polish some details which sounds not so good to me : 
- NPE in RedisRegistrationStoreTest
- lastEndpointUsed in Registration should be mandatory
- add validation on lastEndpointUsed 
- Remove redundant validation in Registration Constructor
- Avoid any Registration call with `null` value for registration ID.

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/endpoint/EndpointUriUtil.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/endpoint/EndpointUriUtil.java
@@ -19,6 +19,8 @@ import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 
+import org.eclipse.leshan.core.util.Validate;
+
 public class EndpointUriUtil {
 
     public static URI createUri(String scheme, String host, int port) {
@@ -47,5 +49,21 @@ public class EndpointUriUtil {
 
     public static InetSocketAddress getSocketAddr(URI uri) {
         return new InetSocketAddress(uri.getHost(), uri.getPort());
+    }
+
+    public static void validateURI(URI uri) throws IllegalArgumentException {
+        Validate.notNull(uri);
+
+        if (uri.getScheme() == null) {
+            throw new IllegalArgumentException("URI Scheme MUST NOT be null");
+        }
+
+        if (uri.getHost() == null) {
+            throw new IllegalArgumentException("URI Host MUST NOT be null");
+        }
+
+        if (uri.getPort() == -1) {
+            throw new IllegalArgumentException("URI Post MUST NOT be undefined");
+        }
     }
 }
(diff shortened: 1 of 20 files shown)


cyril2maq:
I did run my tests which led to create this issue with this PR, it is now OK👍 

Regarding the last commit, I think an additional test could be interesting. Maybe a unit test on LwM2mObservationStore.remove(Token token) method checking that the observation has been removed.

Suggestion for LwM2mObservationStoreTest : 
```
    @Test
    public void remove_observation() {
        // given
        givenASimpleRegistration(lifetime);
        store.addRegistration(registration);

        org.eclipse.californium.core.observe.Observation observationToStore = prepareCoapCompositeObservation();
        observationStore.put(exampleToken, observationToStore);

        // when
        observationStore.remove(exampleToken);

        // then
        Observation leshanObservation = store.getObservation(registrationId,
                new ObservationIdentifier(exampleToken.getBytes()));
        assertNull(leshanObservation);
    }
```

sbernard31:
I added your suggestions (commit dbe5aa2)

I guess I can merge the PR in `master` now ?

cyril2maq:
> I added your suggestions (commit [dbe5aa2](https://github.com/eclipse/leshan/commit/dbe5aa21840d6a1f41606612eec09736193e3c89))
> 
> I guess I can merge the PR in `master` now ?

Yes !

sbernard31:
Thx for reviewing this :wink:

Issue #1387, opened 2023-01-25
Leshan Client : BootstrapDeleteRequest triggers *all* objects instances to be deleted


cyril2maq:
### Version(s)

9859a330e469bab306f4be3ab2ada767ef85de9c

### Which components

leshan client

### What happened

When the client receive a BootstrapDeleteRequest, it leads to clear all its objects instances whatever the 'path' value of the request is. Indeed in the following code, the 'objects' field contains all objects declared in the client :

https://github.com/eclipse/leshan/blob/122a1a8556f659f2a9b43ef81d4250672bc72e22/leshan-client-core/src/main/java/org/eclipse/leshan/client/bootstrap/BootstrapHandler.java#L95-L97

A filter matching the objects and the BootstrapDeleteRequest path is missing.

(the BootstrapDeleteRequest path must still match one of the client object, else the request will be rejected beforehand in ServerMessageDeliverer due to the findResource check). 

I think the key point is that previously in M10, the instance/request filtering was made at ObjectResource level where it used a dedicated nodeEnabler (restricted to the targeted objectId) instead of the generic DownlinkRequestReceiver which will trigger the call to BootstrapHandler.delete and iterates on all the client's objects.

### How to reproduce

Setup a leshan bootstrap server.
Setup a leshan-client in bootstrap mode, with several objects in its model. 
Setup a Bootstrap sequence for this client with a BootstrapConfig having 'toDelete' with just '/0'.

All objects instances of the client are in fact cleared instead of just '/0' ones.

sbernard31:
**Thx for your very clear reports :pray: !**

It seems you catch another bugs :slightly_smiling_face:,  I will work on it !

sbernard31:
#1389 should fix this let me know, if you plan to review/test it :pray:

cyril2maq:
Yes I will test it soon

sbernard31:
PR #1389 is now integrated in master.

Thx again for your help :+1:

PR #1389, opened 2023-01-25
BootstrapDeleteRequest must not always delete all instances at client side

sbernard31:
This aims to fix : #1387


cyril2maq:
This did fix the behavior described in  #1387 in my tests.
The PR LGTM.

Issue #1249, opened 2022-04-28
Configurable prefix for Redis stores


sbernard31:
A user from mailing list ask for a way define prefix for Redis Store.

E.g. FIRST_APP:REG:EP, OTHER_APP:REG:EP. 

See : https://www.eclipse.org/lists/leshan-dev/msg01118.html

sbernard31:
See #1398 which allow to set redis key and prefix on `RedisSecurityStore`.

sbernard31:
This is now implemented in `master` (see #1409)

Issue #1025, opened 2021-06-09
Add a way to add more transport Layer 


sbernard31:
In v1.0, LWM2M was only based CoAP(s).
In v1.1, LWM2M can be used over  CoAP+TCP,  CIoT or LoRaWAN
In v1.2, LWM2M can be used over MQTT and HTTP (theorically ... https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/520)

So at some time we will need to refactor the code to allow this kind of addition.

This issue is create to discuss/ start thinking about this.

sbernard31:
For LWM2M v1.1, I created 2 issues : 
- #1047 
- #1046

sbernard31:
I started to explore about this at #1220.

sbernard31:
I created an issue about new Java Module architecture : #1295.

sbernard31:
A try for the new API : https://github.com/eclipse/leshan/pull/1220#issuecomment-1206271554

sbernard31:
A cleaner version of the abstraction at client side is available at #1323 .

sbernard31:
A  version of the abstraction at bootstrap server side is available at #1336

sbernard31:
First version of transport abstraction layer is now integrated in `master`.

### :dart:   Why a transport layer abstraction ?
This transport abstraction layer aims to allow to :
 - remove strong californium dependencies. (cf should only be used to implement an endpoints provider)
 - being able to add any number of endpoints. (not just 1 coap and 1 coaps)
 - being able to add more transport layer.
 - being able to have different implementation of same transport layer. (could be different underlying library or just same library with different configuration)

### :information_source:   What is current state  ?
See PRs below are now integrated in `master`.
See their description to have an overview of new API, what is remaining to do and what was removed or broken :** 
 - https://github.com/eclipse/leshan/pull/1318
 - https://github.com/eclipse/leshan/pull/1323
 - https://github.com/eclipse/leshan/pull/1336


### :pray:  How you could help ?

This new Transport Abstraction Layer comes with lot of changes. So we strongly encourage community to give us feedback about API, design and any regressions or missing features.

If you decide to tests `master` and you face any difficulties to migrate to the new API, do no hesitate to ask ! (by opening a new issue)

### :construction_worker: :  What next ?

The next big step will be to apply a new module design (this will be mainly about project/package renaming): 
 -   https://github.com/eclipse/leshan/issues/1295
 
[...]

PR #1398, opened 2023-02-09
Allow to set custom REDIS keys for security endpoint and PSK ID

aliakseiz:
Added setters for `SEC_EP` and `PSKID_SEC` to be able to override the default values.

### DIFF
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/RedisSecureIntegrationTestHelper.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/RedisSecureIntegrationTestHelper.java
@@ -34,7 +34,7 @@ public class RedisSecureIntegrationTestHelper extends SecureIntegrationTestHelpe
         } else {
             jedis = new JedisPool();
         }
-        securityStore = new RedisSecurityStore(jedis);
+        securityStore = new RedisSecurityStore.Builder(jedis).build();
         return securityStore;
     }
 }
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/RedisSecureIntegrationTestHelper.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/RedisSecureIntegrationTestHelper.java
@@ -34,7 +34,7 @@ public class RedisSecureIntegrationTestHelper extends SecureIntegrationTestHelpe
         } else {
             jedis = new JedisPool();
         }
-        securityStore = new RedisSecurityStore.Builder(jedis).build();
+        securityStore = new RedisSecurityStore(jedis);
         return securityStore;
     }
 }
--- a/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
+++ b/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
@@ -34,7 +34,7 @@ import org.eclipse.californium.elements.config.Configuration;
 import org.eclipse.californium.elements.util.CertPathUtil;
 import org.eclipse.californium.scandium.config.DtlsConfig;
 import org.eclipse.californium.scandium.config.DtlsConfig.DtlsRole;
-import org.eclipse.californium.scandium.config.DtlsConnectorConfig.Builder;
+import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.servlet.ServletHolder;
 import org.eclipse.jetty.webapp.WebAppContext;
@@ -162,7 +162,7 @@ public class LeshanServerDemo {
             securityStore = new FileSecurityStore();
         } else {
             // use Redis Store
-            securityStore = new RedisSecurityStore(cli.main.redis);
+            securityStore = new RedisSecurityStore.Builder(cli.main.redis).build();
             builder.setRegistrationStore(new RedisRegistrationStore(cli.main.redis));
         }
         builder.setSecurityStore(securityStore);
@@ -190,8 +190,10 @@ public class LeshanServerDemo {
                 return new CoapsServerEndpointFactory(uri) {
 
                     @Override
-                    protected Builder createDtlsConnectorConfigBuilder(Configuration endpointConfiguration) {
-                        Builder dtlsConfigBuilder = super.createDtlsConnectorConfigBuilder(endpointConfiguration);
+                    protected DtlsConnectorConfig.Builder createDtlsConnectorConfigBuilder(
+                            Configuration endpointConfiguration) {
+                        DtlsConnectorConfig.Builder dtlsConfigBuilder = super.createDtlsConnectorConfigBuilder(
+                                endpointConfiguration);
 
                         // Add MDC for connection logs
                         if (cli.helpsOptions.getVerboseLevel() > 0)
--- a/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
+++ b/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
@@ -34,7 +34,7 @@ import org.eclipse.californium.elements.config.Configuration;
 import org.eclipse.californium.elements.util.CertPathUtil;
 import org.eclipse.californium.scandium.config.DtlsConfig;
 import org.eclipse.californium.scandium.config.DtlsConfig.DtlsRole;
-import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
+import org.eclipse.californium.scandium.config.DtlsConnectorConfig.Builder;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.servlet.ServletHolder;
 import org.eclipse.jetty.webapp.WebAppContext;
@@ -162,7 +162,7 @@ public class LeshanServerDemo {
             securityStore = new FileSecurityStore();
         } else {
             // use Redis Store
-            securityStore = new RedisSecurityStore.Builder(cli.main.redis).build();
+            securityStore = new RedisSecurityStore(cli.main.redis);
             builder.setRegistrationStore(new RedisRegistrationStore(cli.main.redis));
         }
         builder.setSecurityStore(securityStore);
@@ -190,10 +190,8 @@ public class LeshanServerDemo {
                 return new CoapsServerEndpointFactory(uri) {
 
                     @Override
-                    protected DtlsConnectorConfig.Builder createDtlsConnectorConfigBuilder(
-                            Configuration endpointConfiguration) {
-                        DtlsConnectorConfig.Builder dtlsConfigBuilder = super.createDtlsConnectorConfigBuilder(
-                                endpointConfiguration);
+                    protected Builder createDtlsConnectorConfigBuilder(Configuration endpointConfiguration) {
+                        Builder dtlsConfigBuilder = super.createDtlsConnectorConfigBuilder(endpointConfiguration);
 
                         // Add MDC for connection logs
                         if (cli.helpsOptions.getVerboseLevel() > 0)
--- a/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
+++ b/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
@@ -40,22 +40,28 @@ import redis.clients.jedis.util.Pool;
  */
 public class RedisSecurityStore implements EditableSecurityStore {
 
-    private static final String SEC_EP = "SEC#EP#";
+    private final String securityInfoByEndpointPrefix;
 
-    private static final String PSKID_SEC = "PSKID#SEC";
+    private final String endpointByPskIdKey;
 
     private final Pool<Jedis> pool;
 
     private final List<SecurityStoreListener> listeners = new CopyOnWriteArrayList<>();
 
     public RedisSecurityStore(Pool<Jedis> pool) {
+        this(pool, "SEC#EP#", "PSKID#SEC");
+    }
+
+    public RedisSecurityStore(Pool<Jedis> pool, String securityInfoByEndpointPrefix, String endpointByPskIdKey) {
         this.pool = pool;
+        this.securityInfoByEndpointPrefix = securityInfoByEndpointPrefix;
+        this.endpointByPskIdKey = endpointByPskIdKey;
     }
 
     @Override
     public SecurityInfo getByEndpoint(String endpoint) {
         try (Jedis j = pool.getResource()) {
-            byte[] data = j.get((SEC_EP + endpoint).getBytes());
+            byte[] data = j.get((securityInfoByEndpointPrefix + endpoint).getBytes());
             if (data == null) {
                 return null;
             } else {
@@ -67,11 +73,11 @@ public class RedisSecurityStore implements EditableSecurityStore {
     @Override
     public SecurityInfo getByIdentity(String identity) {
         try (Jedis j = pool.getResource()) {
-            String ep = j.hget(PSKID_SEC, identity);
+            String ep = j.hget(endpointByPskIdKey, identity);
             if (ep == null) {
                 return null;
             } else {
-                byte[] data = j.get((SEC_EP + ep).getBytes());
+                byte[] data = j.get((securityInfoByEndpointPrefix + ep).getBytes());
                 if (data == null) {
                     return null;
                 } else {
@@ -90,7 +96,7 @@ public class RedisSecurityStore implements EditableSecurityStore {
     @Override
     public Collection<SecurityInfo> getAll() {
         try (Jedis j = pool.getResource()) {
-            ScanParams params = new ScanParams().match(SEC_EP + "*").count(100);
+            ScanParams params = new ScanParams().match(securityInfoByEndpointPrefix + "*").count(100);
             Collection<SecurityInfo> list = new LinkedList<>();
             String cursor = "0";
             do {
@@ -111,19 +117,19 @@ public class RedisSecurityStore implements EditableSecurityStore {
         try (Jedis j = pool.getResource()) {
             if (info.getPskIdentity() != null) {
                 // populate the secondary index (security info by PSK id)
-                String oldEndpoint = j.hget(PSKID_SEC, info.getPskIdentity());
+                String oldEndpoint = j.hget(endpointByPskIdKey, info.getPskIdentity());
                 if (oldEndpoint != null && !oldEndpoint.equals(info.getEndpoint())) {
                     throw new NonUniqueSecurityInfoException(
                             "PSK Identity " + info.getPskIdentity() + " is already used");
                 }
-                j.hset(PSKID_SEC.getBytes(), info.getPskIdentity().getBytes(), info.getEndpoint().getBytes());
+                j.hset(endpointByPskIdKey.getBytes(), info.getPskIdentity().getBytes(), info.getEndpoint().getBytes());
             }
 
-            byte[] previousData = j.getSet((SEC_EP + info.getEndpoint()).getBytes(), data);
+            byte[] previousData = j.getSet((securityInfoByEndpointPrefix + info.getEndpoint()).getBytes(), data);
             SecurityInfo previous = previousData == null ? null : deserialize(previousData);
             String previousIdentity = previous == null ? null : previous.getPskIdentity();
             if (previousIdentity != null && !previousIdentity.equals(info.getPskIdentity())) {
-                j.hdel(PSKID_SEC, previousIdentity);
+                j.hdel(endpointByPskIdKey, previousIdentity);
             }
 
             return previous;
@@ -133,14 +139,14 @@ public class RedisSecurityStore implements EditableSecurityStore {
     @Override
     public SecurityInfo remove(String endpoint, boolean infosAreCompromised) {
         try (Jedis j = pool.getResource()) {
-            byte[] data = j.get((SEC_EP + endpoint).getBytes());
+            byte[] data = j.get((securityInfoByEndpointPrefix + endpoint).getBytes());
 
             if (data != null) {
                 SecurityInfo info = deserialize(data);
                 if (info.getPskIdentity() != null) {
-                    j.hdel(PSKID_SEC.getBytes(), info.getPskIdentity().getBytes());
+                    j.hdel(endpointByPskIdKey.getBytes(), info.getPskIdentity().getBytes());
                 }
-                j.del((SEC_EP + endpoint).getBytes());
+                j.del((securityInfoByEndpointPrefix + endpoint).getBytes());
                 for (SecurityStoreListener listener : listeners) {
                     listener.securityInfoRemoved(infosAreCompromised, info);
                 }
--- a/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
+++ b/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
@@ -38,9 +38,6 @@ import redis.clients.jedis.util.Pool;
  * <p>
  * Security info are stored using the endpoint as primary key and a secondary index is created for endpoint lookup by
  * PSK identity.
- * <p>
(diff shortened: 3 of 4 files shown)


sbernard31:
Hi, thx for this PR. :pray: 

Before to talk about the legal agreement issue :point_up: and review the code, could you explain why you need this, maybe you can explain a little your use case and the context where you need it ? This should help to provide a solution for your need.

aliakseiz:
Hi,
we have a strict REDIS keys naming rules in our company and the default `SEC#EP#` and `PSKID#SEC` don't comply with it. Also I couldn't find a way to override them using current code base.

For the moment I use a custom `RedisSecurityStore`, which differs only by changes provided in the PR.

Wanted to avoid such a copy-pasting and to make life a bit easier when upgrading the gateway to the next milestone. Was not sure what is the correct way to do that, decided to create a PR.

By the way, ECA is settled now.

jvermillard:
Looks useful, I'm starting to mix Leshan data structure and some other applications in the same Redis cluster, and I have the same need: a way to change the prefix to organize the data better.

That said, I'm not sure this should still be static and set by a setter, but a final attribute you can set using a constructor.

aliakseiz:
@jvermillard thank you for suggestion, constructor is much better, indeed.
I updated PR accordingly.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
```diff
@@ -40,16 +40,24 @@
  */
 public class RedisSecurityStore implements EditableSecurityStore {
 
-    private static final String SEC_EP = "SEC#EP#";
+    private final String SEC_EP;
```
jvermillard, line 43: Capitalization is only for constants

aliakseiz: Sure thing, missed that somehow.

sbernard31:
> we have a strict REDIS keys naming rules in our company and the default SEC#EP# and PSKID#SEC don't comply with it. 

OK I understand.
This makes me think to this issue : https://github.com/eclipse/leshan/issues/1249.
Do you also need this kind of feature for `RedisRegistrationStore` ?
@aliakseiz, just by curiosity you are using Leshan **v2.0.0-Mx** ?

> That said, I'm not sure this should still be static and set by a setter, but a final attribute you can set using a constructor.

:+1: 

> By the way, ECA is settled now.

:+1:

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
```diff
@@ -40,22 +40,30 @@
...
 
     private final Pool<Jedis> pool;
 
     private final List<SecurityStoreListener> listeners = new CopyOnWriteArrayList<>();
 
     public RedisSecurityStore(Pool<Jedis> pool) {
         this.pool = pool;
+        this.secEp = "SEC#EP#";
+        this.pskIdSec = "PSKID#SEC";
+    }
```
sbernard31, line 55: Probably better to use : 

```suggestion
    public RedisSecurityStore(Pool<Jedis> pool) {
        this(pool, "SEC#EP#", "PSKID#SEC");
    }
```

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
```diff
@@ -40,22 +40,30 @@
  */
 public class RedisSecurityStore implements EditableSecurityStore {
 
-    private static final String SEC_EP = "SEC#EP#";
+    private final String secEp;
 
-    private static final String PSKID_SEC = "PSKID#SEC";
+    private final String pskIdSec;
```
sbernard31, line 45: I don't know if you agree with me but I feel this attribute names are not so clear.
Maybe good opportunity to change it to something clearer ?

`secEp` => `securityInfoByEndpointPrefix`
`pskIdSec` => `endpointByIdentityKey`

If we change this we need to also change it for constructor arguments at : 
`public RedisSecurityStore(Pool<Jedis> pool, String secEp, String pskIdSec)`

sbernard31:
Maybe out of scope of this PR, but I also feel that default redis key value are not so good...

**Being more consistent :** 

**1)  `SEC#EP#` key**
This is a prefix to create a key from "endpoint name" to retrieve a `SecurityInfo` (E.g. of key `SEC#EP#my_endpoint`)
(this is used with redis [GET](https://redis.io/commands/get/) command)
So first part "SEC" is what we search and second part "EP" is the field used as index and third part "my_endpoint" is the value of the field.

**2) `PSKID#SEC`** key
Here this is the key search a value in a HashStored by field (See redis [HGET](https://redis.io/commands/hget/) command)
This allow to find "endpoint name" by "PSK Identity".
So order seems different, first part is the field index name "PSKID" and second part "SEC" the value ?
But why "SEC" as we store the "endpoint name" ? 
If we want to be consistent this should be some thing like EP#PSKID ?

What do you think ? Should we at least change `PSKID#SEC` to `EP#PSKID` ?
Or any better ideas for those key name ?
Or better to not change it for backward compatibility ? (knowing that is OK to break compatibility between 1.x and 2.x and with this new PR it is easy to use old prefix)

**Should we add more prefix to all key   ?** 

Do we need a Leshan global prefix used for all store ? e.g. : `LSHN#PSKID#SEC` and` LSHN#SEC#EP#`
OR prefix by store ? e.g. : `SECSTORE#PSKID#SEC` and `SECSTORE#SEC#EP#`
OR both ? e.g. `LSHN#SECSTORE#PSKID#SEC` and `LSHN#SECSTORE#SEC#EP#`

This allow to have separate namespaces at cost of more bytes. Don't know if we should go in that way :shrug:

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
aliakseiz: Sure, thank you.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
aliakseiz: Totally agree, more descriptive names will be better.

aliakseiz:
> just by curiosity you are using Leshan v2.0.0-Mx ?

Yes, we have a gateway in production environment based on a v2.0.0-M10. Trying to keep it up to date with the latest Leshan since v1.0.0-RC2. Didn't face any critical issues so far.

> This makes me think to this issue : #1249.
> Do you also need this kind of feature for `RedisRegistrationStore` ?

Not really, but I realize that for sake of consistency and convenience it might be nice to have an ability to customize REDIS keys or at least a root prefix.
I could try extending current PR with a similar approach for `RedisRegistrationStore`. Or you would suggest to prepare a separate one?

sbernard31:
> Yes, we have a gateway in production environment based on a v2.0.0-M10. Trying to keep it up to date with the latest Leshan since v1.0.0-RC2. Didn't face any critical issues so far.

Good to know :slightly_smiling_face:  
Did you test `master` since we made lot of changes about Transport Layer Abstraction ? See https://github.com/eclipse/leshan/issues/1025#issuecomment-1373824399.

> I could try extending current PR with a similar approach for RedisRegistrationStore. Or you would suggest to prepare a separate one?

I think this is better to limit this PR to `SecurityStore`.
Then we can do a separated PR for RegistrationStore. (It will probably lead to more question because there is a lot of argument and so maybe we need a builder or another solution ?)

Maybe we should also decide about https://github.com/eclipse/leshan/pull/1398#issuecomment-1429465578 before to move forward on RegistrationStore ? :thinking:

sbernard31:
Maybe if we need to discuss about RedisRegistrationStore we can use https://github.com/eclipse/leshan/issues/1249.

jvermillard:
@sbernard31 if you write a migration script, I'm fine with the rename to EP#PSKID :joy:

aliakseiz:
> Did you test master since we made lot of changes about Transport Layer Abstraction ?

Yes, no issues so far. Migration was rather easy and straightforward.

Regarding https://github.com/eclipse/leshan/pull/1398#issuecomment-1429465578:
> What do you think ? Should we at least change PSKID#SEC to EP#PSKID ?

As for me, I like the `SEC#EP#key` and `EP#PSKID` names. The order is consistent and has some logic behind. Once you get it, you can navigate easily.

> Or better to not change it for backward compatibility?

Those who will experience  issues with the new keys could easily override them to current ones and perform the migration later if needed.

> Should we add more prefix to all key ? Do we need a Leshan global prefix used for all store?

Let's indeed continue the discussion in https://github.com/eclipse/leshan/issues/1249 .

sbernard31:
> Yes, no issues so far. Migration was rather easy and straightforward.

Thx a lot for this feedback. :pray:  
If you have any feedback or idea to improve about the new API, do not hesitate to share.  

About `EP#PSKID`, I agree with so let's change the key value.
Let's me know if you plan to do it in this PR ?
If yes, maybe in another commit, with javadoc explaining the break between v1.x and v2.x and old key to use for backward compatibility ?

> Let's indeed continue the discussion in https://github.com/eclipse/leshan/issues/1249 .

I'm waiting for your opinion about prefix. :slightly_smiling_face:

aliakseiz:
> Let's me know if you plan to do it in this PR ?
> If yes, maybe in another commit, with javadoc explaining the break between v1.x and v2.x and old key to use for backward compatibility ?

Updated the PR accordingly. Any suggestions on the wording in javadoc are welcome.

> Do we need a Leshan global prefix used for all store ? e.g. : LSHN#PSKID#SEC and LSHN#SEC#EP#
> OR prefix by store ? e.g. : SECSTORE#PSKID#SEC and SECSTORE#SEC#EP#
> OR both ? e.g. LSHN#SECSTORE#PSKID#SEC and LSHN#SECSTORE#SEC#EP#

From my point of view, if you have a Redis instance dedicated for Leshan, prefix by store makes the most sense. Because it provides a necessary grouping and a quick overview of what is stored under those keys.

Global prefix might be useful only when Redis is shared between services, but in this case the prefix `LSHN#SECSTORE#SEC#EP#` most probably will not comply with your internal naming rules and you would want to use a custom prefixes anyway.

sbernard31:
> Any suggestions on the wording in javadoc are welcome

I review the last commit and it looks good to me. :+1: 

Just a small detail about the changes on `CONTRIBUTING.md`, in the future do not hesitate to create dedicated commit or even dedicated PR  for this kind of "out ot topic" changes.

> From my point of view, if you have a Redis instance dedicated for Leshan, prefix by store makes the most sense. Because it provides a necessary grouping and a quick overview of what is stored under those keys.
> 
> Global prefix might be useful only when Redis is shared between services, but in this case the prefix LSHN#SECSTORE#SEC#EP# most probably will not comply with your internal naming rules and you would want to use a custom prefixes anyway.

It makes sense to me. So let's add only 1 store prefix.

Remaining questions are : 
1. What should we use as default prefix ? `SECSTORE#` or `LSHNSECSTORE` or something else ?
2. Do we want to add a way to set prefix ? (like adding new constructor or add parameter to existent one)  OR letting user able to only change key is enough.

If we want to provide a way to set prefix we need to decide how ?  none exclusive solution could be : 
2.1 add a constructor `RedisSecurityStore(Pool<Jedis> pool, String prefix)`
2.2 add a constructor `RedisSecurityStore(Pool<Jedis> pool, String prefix, String securityInfoByEndpointPrefix, String endpointByPskIdKey)`
2.3 add a builder.

For prefix do you prefer to continue with this PR or a new one ?
Just by curiosity do you also plan to work on `RedisRegistrationStore` ?

Again thx for you contribution, that's really appreciated :pray:

aliakseiz:
> in the future do not hesitate to create dedicated commit

Sorry about that, probably a bad habit of mine from other projects. Will keep it in mind next time.

I would go for `SECSTORE#`, because it's shorter and still clear enough. If somebody would need to have an indication that it belongs to Leshan, it can be easily overridden via constructor.
To me it looks like using a builder here might be a bit to excessive. Not sure if the builders are widely used across Leshan and are treated as a common practice.
Maybe options 2.1 and 2.2 would be enough? First one for those who don't care about the internal naming and just need to hide the Leshan security store under some key. And second one for complete control over the keys and prefixes.

I will update this PR with prefix setting.

>  Just by curiosity do you also plan to work on RedisRegistrationStore ?

Yes, right after finishing current PR.

sbernard31:
> Sorry about that,

Not a big deal :wink: 

> I would go for SECSTORE#

:+1: 

> Not sure if the builders are widely used across Leshan and are treated as a common practice.

I can confirm Leshan code base has several Builder. There are generally added when there is too many parameters in constructors or too many constructors.
I know `Builder` has many drawback but I didn't find better way to create instance of a class with many `final` parameters.

> To me it looks like using a builder here might be a bit to excessive.

I agree adding  builder for `RedisSecurityStore` seems a bit overkill.
But my guess is that for `RedisRegisrationStore` this will be hard to avoid `Builder` because there is so many parameters.
So one reason to use Builder in `RedisSecutiryStore` could be for API consistency with `RedisRegisrationStore` (of course this point makes sense only if we decide to use builder in `RedisRegisrationStore`)

I just said that to let you know the big picture.  
On my side, I have no strong opinion between `builder` vs `more constructor`  in `RedisSecurityStore`.
So, feel free to add `constructor` or builder. 

> Maybe options 2.1 and 2.2 would be enough? First one for those who don't care about the internal naming and just need to hide the Leshan security store under some key. And second one for complete control over the keys and prefixes.

If you chose constructor, this makes sense to me.

> Yes, right after finishing current PR.

Great :+1:

aliakseiz:
> ...for API consistency with RedisRegisrationStore

That was exactly what I thought about right after sending my previous comment :slightly_smiling_face: 
Had some doubts regarding usage mostly because I am not familiar with the Leshan codebase yet. 

> because there is so many parameters

I totally agree with you and will add a builder then.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
```diff
@@ -52,14 +49,10 @@ public class RedisSecurityStore implements EditableSecurityStore {
...
 
-    public RedisSecurityStore(Pool<Jedis> pool) {
-        this(pool, "SEC#EP#", "EP#PSKID");
-    }
-
-    public RedisSecurityStore(Pool<Jedis> pool, String securityInfoByEndpointPrefix, String endpointByPskIdKey) {
-        this.pool = pool;
-        this.securityInfoByEndpointPrefix = securityInfoByEndpointPrefix;
-        this.endpointByPskIdKey = endpointByPskIdKey;
+    private RedisSecurityStore(Builder builder) {
```
sbernard31, line 52: If we want to allow inheritance of `RedisSecurityStore` maybe we should at least make it protected ?

aliakseiz:
Thank you @sbernard31 , those are all valid points and suggestions.
I think I fixed all of them, please take a look again.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
```diff
@@ -49,7 +49,13 @@ public class RedisSecurityStore implements EditableSecurityStore {
 
     private final List<SecurityStoreListener> listeners = new CopyOnWriteArrayList<>();
 
-    private RedisSecurityStore(Builder builder) {
+    public RedisSecurityStore(Pool<Jedis> pool) {
+        this.pool = pool;
+        this.securityInfoByEndpointPrefix = "SECSTORE#SEC#EP#";
+        this.endpointByPskIdKey = "SECSTORE#EP#PSKID";
```
sbernard31, line 55: maybe you can use 
```java
this(new Builder(pool).build());
```

So there is only 1 place where we define default value.
I know we will create a builder instance  just to do that but this is something which is not called often

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
```diff
@@ -215,7 +221,7 @@ public void setEndpointByPskIdKey(String endpointByPskIdKey) {
          * Set the prefix for all keys and prefixes including {@link #securityInfoByEndpointPrefix} and
          * {@link #endpointByPskIdKey}.
          * <p>
-         * Default value is {@literal SECSTORE#}.
+         * Default value is {@literal SECSTORE#}. Should not be {@code null}, can be empty.
```
sbernard31, line 224: Here it is said : `Should not be {@code null}`

But looking at : 
```java
    if (this.prefix != null) {
                this.securityInfoByEndpointPrefix = this.prefix + this.securityInfoByEndpointPrefix;
                this.endpointByPskIdKey = this.prefix + this.endpointByPskIdKey;
            }
```
You manage `null`.

I think it's OK to support `null` prefix.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/RedisSecureIntegrationTestHelper.java
```diff
@@ -34,7 +34,7 @@ protected SecurityStore createSecurityStore() {
         } else {
             jedis = new JedisPool();
         }
-        securityStore = new RedisSecurityStore(jedis);
+        securityStore = new RedisSecurityStore.Builder(jedis).build();
```
sbernard31, line 37: maybe we can now avoid to do this changes as constructor is back.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
```diff
@@ -162,7 +162,7 @@ public static LeshanServer createLeshanServer(LeshanServerDemoCLI cli) throws Ex
             securityStore = new FileSecurityStore();
         } else {
             // use Redis Store
-            securityStore = new RedisSecurityStore(cli.main.redis);
+            securityStore = new RedisSecurityStore.Builder(cli.main.redis).build();
```
sbernard31, line 165: Same here : maybe we can now avoid to do this changes as constructor is back.

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
```diff
@@ -190,8 +190,10 @@ public CaliforniumServerEndpointFactory createDefaultEndpointFactory(URI uri) {
                 return new CoapsServerEndpointFactory(uri) {
 
                     @Override
-                    protected Builder createDtlsConnectorConfigBuilder(Configuration endpointConfiguration) {
-                        Builder dtlsConfigBuilder = super.createDtlsConnectorConfigBuilder(endpointConfiguration);
+                    protected DtlsConnectorConfig.Builder createDtlsConnectorConfigBuilder(
+                            Configuration endpointConfiguration) {
+                        DtlsConnectorConfig.Builder dtlsConfigBuilder = super.createDtlsConnectorConfigBuilder(
+                                endpointConfiguration);
```
sbernard31, line 196: I guess this part of code should not be changed ?

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
aliakseiz: Certainly makes sense. I added a `protected` constructor to be able to use `this(...)` here.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
aliakseiz: Forgot about it, wrote a JavaDoc first, then added a null-check.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/util/RedisSecureIntegrationTestHelper.java
aliakseiz: Yes, of course, you're right.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
sbernard31: No problem, this happens. An easy way to limit this kind of issue is to do a review of you code, like if you review someone else's code. :wink:

leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/LeshanServerDemo.java
aliakseiz: Perhaps it was corrected by my IDE. Reverted it back.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
aliakseiz: This is actually one of the "rules" in my team at work. Don't know, why I didn't follow it while preparing this PR.
Will pay more attention next time.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
sbernard31: Uups my mistake I wanted to say : 
``` java
this(new Builder(pool)); 
```

So no need new constructor;

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
```diff
@@ -167,4 +180,90 @@ public void addListener(SecurityStoreListener listener) {
...
+        private String endpointByPskIdKey;
+
+        private String prefix;
+
+        /**
+         * Set the Redis connection pool for the {@link RedisSecurityStore}.
+         */
+        public void setPool(Pool<Jedis> pool) {
+            this.pool = pool;
+        }
```
sbernard31, line 204: As we give pool as Builder constructor maybe we don't need this setter or do you have a use case in mind ?

sbernard31:
I didn't notice that you Builder is not fluent.
Are you OK if we change it in that way ? (each setter must return the builder so you can write)

```java
securityStore = new RedisSecurityStore.Builder(jedis).
       .setPrefix(null) //
       .setEndpointByPskIdKey("EPKEY") //
       .setSecurityInfoByEndpointPrefix("SECKey").build();
```

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
aliakseiz: Good, it's exactly what I wanted to do at first.

leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisSecurityStore.java
aliakseiz: Good point, I removed this setter.

aliakseiz:
> I didn't notice that you Builder is not fluent.

Don't know what I was thinking of, of course setters should return `Builder` instance. Fixed it.

sbernard31 (approved):
:heavy_check_mark:  I think we're good now.

I play a bit with it and it seems to work, I also tested it locally with `mvn clean install javadoc:javadoc -Predis` and all seems good :+1:  !

I will now integrated this in `master`.

Thank you very much for this contribution :pray:

sbernard31:
(as I said we will ignore the github action issue, I haven't time to go deeper in it and didn't find any obvious way to fix it)

aliakseiz:
> I think we're good now.

Thank you for all your advice and for patience :+1: 

I will start https://github.com/eclipse/leshan/issues/1249 now.

sbernard31:
> Thank you for all your advice and for patience +1

You're welcome.

> I will start https://github.com/eclipse/leshan/issues/1249 now.

:+1:

sbernard31:
This is now integrated in `master` (in squashed commit fd5c9c0b4c620e888fe8221c9142134c92292eed + typo commit b7dc3cf3f828864ff44d654891f15751f1a43668)

I notice that your email is `@any-ware.nl`. Fun fact, when I started to work on Eclipse project my company was named `anyware-technology`. It seems to be a common name in IoT :grin:

PR #1404, opened 2023-02-27
Changing CiperSuiteId from single to multiple resource

Warmek:
Changing CipherSuiteId field in BootstrapConfig.java from single resource to multiple resource as to match OMA specification.

It aims to implement https://github.com/eclipse/leshan/issues/1402

PR #1408, opened 2023-03-03
Fix Invalid datatype for DTLS/TLS Ciphersuite resource (/0/?/16) in BootstrapConfig

### LINKED ISSUE #1402: Invalid datatype for `DTLS/TLS Ciphersuite` resource (/0/?/16) in BootstrapConfig.
### Question

In BootstrapConfig.java line 335:
```java
        /**
         * When this resource is present it instructs the TLS/DTLS client to propose the indicated ciphersuite(s) in the
         * ClientHello of the handshake. A ciphersuite is indicated as a 32-bit integer value. The IANA TLS ciphersuite
         * registry is maintained at https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml. As an
         * example, the TLS_PSK_WITH_AES_128_CCM_8 ciphersuite is represented with the following string "0xC0,0xA8". To
         * form an integer value the two values are concatenated. In this example, the value is 0xc0a8 or 49320.
         * <p>
         * Since Security v1.1
         */
        public ULong cipherSuite = null;
```

In OMA documentation cipherSuite is defined as "Multiple" but from what I can read above it seems that the resource can be "concatenated" to integer value. My question is: Is ULong a correct type for this resource or should it be another type that could contain 2 Strings?

sbernard31:
This aims to fix #1402.
This is work is based on PR #1404

### DIFF
--- a/leshan-server-core/pom.xml
+++ b/leshan-server-core/pom.xml
@@ -43,6 +43,11 @@ Contributors:
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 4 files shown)


Warmek (approved):
👍

sbernard31:
@Warmek thx for review. 

This is now integrated in master (all commit squashed in 8f0db2449248bb4d20d4237d09ddb095a83f5969)

Issue #1441, opened 2023-04-24
Resources with no defined operations MUST NOT be considered as executable


Warmek:
### Question

Should resources with no defined operations be defined as executable in leshan server demo? For example, resource 13 in 1-1_1.xml don't have any operation defined but in leshan server demo it's displayed as executable.

sbernard31:
> Should resources with no defined operations be defined as executable in leshan server demo?

Nope, this is a bug.

Warmek:
I want to fix it. Where should I start looking for assigning label for this button?

sbernard31:
Just to be sure we are talking about `master` ? (not `1.x` branch ?)

sbernard31:
If this is about `master`.
I take a look at what is sent by the REST API by using my browser developer Tools. 
And I see for resource 13 : 
```json
{
	"13": {
		"id": 13,
		"name": "Registration Priority Order",
		"operations": "NONE",
		"instancetype": "single",
		"mandatory": false,
		"type": "unsigned_integer",
		"range": "",
		"units": "",
		"description": "The LwM2M Client sequences the LwM2M Server registrations in increasing order of this value. If this value is not defined, registration attempts to this server are not impacted by other server registrations."
	}
}
```

And looking at : 
https://github.com/eclipse/leshan/blob/d5052f181918af3680eb42a7428810a8fb65bcdd/leshan-server-demo/webapp/src/components/resources/ResourceControl.vue#L40-L45

and : 
https://github.com/eclipse/leshan/blob/d5052f181918af3680eb42a7428810a8fb65bcdd/leshan-server-demo/webapp/src/components/resources/ResourceControl.vue#L95-L97

It seems that returning `NONE` or checking String contains `E` is not a good idea ?
I don't know what is the best move... 
Maybe returning empty String instead of `NONE` ?
Or check that operation is strictly equals to E ? using includes make sense for `R` or `W` but not for `E`
See : https://github.com/eclipse/leshan/blob/d5052f181918af3680eb42a7428810a8fb65bcdd/leshan-core/src/main/java/org/eclipse/leshan/core/model/ResourceModel.java#L26-L40

Warmek:
I've created a fix on [opl/EXEButton](https://github.com/JaroslawLegierski/leshan/tree/opl/EXEButton).
I went with strictly comparing operation to "E" as I feel like switching from NONE to empty string could be confusing when debugging some issues.
[...]

PR #1442, opened 2023-04-28
Fixed wrongly assigned EXE buttons

Warmek:
Fixed wrong recognition of "NONE" operation as "E" (Execute) in web app, causing resources without operations to be assigned "EXE" buttons

Aims to implement [#1441](https://github.com/eclipse/leshan/issues/1441)

### DIFF
--- a/leshan-server-demo/webapp/src/components/resources/ResourceControl.vue
+++ b/leshan-server-demo/webapp/src/components/resources/ResourceControl.vue
@@ -93,7 +93,7 @@ export default {
       return this.resourcedef.operations.includes("W");
     },
     executable() {
-      return this.resourcedef.operations.includes("E");
+      return this.resourcedef.operations == ("E");
     },
   },
   methods: {


leshan-server-demo/webapp/src/components/resources/ResourceControl.vue
```diff
@@ -93,7 +93,7 @@ export default {
       return this.resourcedef.operations.includes("W");
     },
     executable() {
-      return this.resourcedef.operations.includes("E");
+      return this.resourcedef.operations == ("E");
```
sbernard31, line 96: 1. Any reason to use  `(` `)` ?
2. I think `==` should work but generally I prefer to use `===` by default and use `==` only when conversion is needed.  (See : https://stackoverflow.com/a/3586781/5088764)

Warmek:
I've fixed the issues

sbernard31:
:pray: thx for the fix.

I integrated this in `master` (commit be7963221c56153daf4ca977058e34d0a9ed4fdd)

I also : 
- squashed all commit in one 
- rename the commit message to refer to initial issue.

Issue #1421, opened 2023-03-15
Unable to handle notification after device ip change in M11


JaroslawLegierski:
### Version(s)

leshan-2.0.0-M11

### Which components

leshan-server-demo

### Tested With

leshan-clint-demo

### What happened

We have device with observed resource (e.g. /6/0/0) cid is on. After device ip address change (without reg update) when Leshan server is receiving Notification we are getting following exception: 

```sh
16:17:42.521 [CoapServer(main)#1] WARN org.eclipse.leshan.server.demo.servlet.EventServlet - Unable to handle notification of [AWaY4UQ8lx:/6/0/0]
java.lang.NullPointerException: null
	at org.eclipse.leshan.server.model.VersionedModelProvider$DynamicModel.getObjectModel(VersionedModelProvider.java:70)
	at org.eclipse.leshan.server.model.VersionedModelProvider$DynamicModel.getResourceModel(VersionedModelProvider.java:61)
	at org.eclipse.leshan.core.node.codec.senml.LwM2mNodeSenMLDecoder.getResourceType(LwM2mNodeSenMLDecoder.java:555)
	at org.eclipse.leshan.core.node.codec.senml.LwM2mNodeSenMLDecoder.extractLwM2mResources(LwM2mNodeSenMLDecoder.java:468)
	at org.eclipse.leshan.core.node.codec.senml.LwM2mNodeSenMLDecoder.parseRecords(LwM2mNodeSenMLDecoder.java:271)
	at org.eclipse.leshan.core.node.codec.senml.LwM2mNodeSenMLDecoder.decodeTimestampedData(LwM2mNodeSenMLDecoder.java:182)
	at org.eclipse.leshan.core.node.codec.DefaultLwM2mDecoder.decodeTimestampedData(DefaultLwM2mDecoder.java:201)
	at org.eclipse.leshan.server.californium.endpoint.ServerCoapMessageTranslator.createObservation(ServerCoapMessageTranslator.java:108)
	at org.eclipse.leshan.server.californium.endpoint.CaliforniumServerEndpointsProvider$2.onNotification(CaliforniumServerEndpointsProvider.java:168)
	at org.eclipse.californium.core.network.CoapEndpoint$NotificationDispatcher.onNotification(CoapEndpoint.java:810)
	at org.eclipse.californium.core.network.BaseMatcher$2.onResponse(BaseMatcher.java:250)
	at org.eclipse.californium.core.coap.Request.setResponse(Request.java:916)
[...]

PR #1464, opened 2023-07-03
Add configuration to enable to update registration on send operation. 

sbernard31:
See #1421 for more details.


Warmek (approved):
👍

sbernard31:
@JaroslawLegierski following discussion at https://github.com/eclipse-leshan/leshan/issues/1421

Do you think we should integrate this in `master` ?

JaroslawLegierski:
> @JaroslawLegierski following discussion at #1421
> 
> Do you think we should integrate this in `master` ?

I think that this feature could be useful in the future (e.g. used together also with updateRegistrationOnNotify).

sbernard31:
> I think that this feature could be useful in the future (e.g. used together also with updateRegistrationOnNotify).

Ok I will integrate it in `master`, I hope I will not regret it later as this is not really LWM2M compliant.

Issue #1447, opened 2023-05-12
Support of SNI


sbernard31:
SNI support is currently not planed at short / mid term but this issue aims to discuss about it.

sbernard31:
@jakubsobolewskisag, I see you request about it at https://github.com/eclipse/leshan/issues/1386#issuecomment-1543834576

Do you have a clear idea about how it should looks like ? 
Do not hesitate to share any input about it here.

jakubsobolewskisag:
Hi @sbernard31 !
I tried to enable SNI support on Leshan server using DTLS settings like below:
`coapConfig.set(DtlsConfig.DTLS_USE_SERVER_NAME_INDICATION, true);`

And it seemed to work fine when device connected in regular (no bootstrap) session. Connection worked, SNI name was correctly provided in "serverNames" variable in CertificateProvider. It looked really, really good!

However, when I tried to bootstrap with SNI, at some point the SNI information was lost. It started OK, "serverNames" variable was OK, cipher suite negotiation went fine, and at the point where bootstrap server was making "WriteBootstrapRequest", there was an exception because of not matching endpoints/contexts.

```
2023-05-10 17:31:12.547  INFO 1665197 --- [/0.0.0.0:5684#1] o.e.c.scandium.DTLSConnector.drops       : DTLSConnector (DTLS-0.0.0.0:5684) drops 539 bytes outgoing, IP([0:0:0:0:0:0:0:1]:42063) != DTLS([0:0:0:0:0:0:0:1]:42063,ID:F6C770916D)
[...]

PR #1466, opened 2023-07-10
Add support of SNI 

sbernard31:
Add support of SNI at client side.
Add some integrations tests about SNI.

At server and bootstrap server, there is no specific Leshan abstraction for now, we just use Californium API to support SNI for `coaps` with rpk and x509. 

Note there is also no support added to demo. (If you are using a bootstrap server which is able to put SNI resource (id:14) of Security object (id:0) to a **leshan-client-demo**, it should be able to use it)

More details at #1447.


sbernard31:
This is now integrated in `master` and should be available in **2.0.0-M15**.

PR #1469, opened 2023-07-11
Update dependencies (yarn, java-lib, maven-plugin...)

sbernard31:
There is some compatibility break in maven-surefire-plugin which could break the build


sbernard31:
Ok, that failed like expected but last commit fix the issue.

PR #1483, opened 2023-07-25
Update release.jenkins

fredg02:
Suggestions for https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3435

### DIFF
--- a/.jenkins/release.jenkins
+++ b/.jenkins/release.jenkins
@@ -23,24 +23,15 @@ pipeline {
     buildDiscarder(logRotator(numToKeepStr: '3'))
   }
   stages {
-    stage('Checkout Code') {
-      steps {
-        git url: 'git@github.com:eclipse-leshan/leshan.git',
-            credentialsId: 'github-bot-ssh',
-            branch: '${BRANCH}'
-      }
-    }
-    
     stage('Get Credentials') {
       steps {
-        withCredentials([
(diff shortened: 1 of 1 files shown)


.jenkins/release.jenkins
```diff
@@ -23,24 +23,15 @@ pipeline {
...
   }
   stages {
-    stage('Checkout Code') {
-      steps {
-        git url: 'git@github.com:eclipse-leshan/leshan.git',
-            credentialsId: 'github-bot-ssh',
-            branch: '${BRANCH}'
-      }
-    }
-
```
sbernard31, line 33: If we remove this, how jenkins will know which branch to checkout ?
(the branch to build is a parameter of the build  `${BRANCH})

fredg02: It's already defined in the job config here, when you specify where the Jenkinsfile is:
=>  https://ci.eclipse.org/leshan/job/leshan-bp/configure

sbernard31: it works if the branch name is a build parameter ?

sbernard31: I faced another issue when I didn't add the git checkout in jenkins file.
I'm unable to push my commits because the job is stuck forever. This looks like a bit to : https://stackoverflow.com/a/65436207/5088764

So I add it again but I also use `skipDefaultCheckout()`, so only 1 checkout should be done.

PR #1522, opened 2023-10-03
Add `transport` command to leshan-client-demo

sbernard31:
This is experimentation about adding a `transport` command to **leshan-client-demo** to be able to configure transport layer. 

This looks like : 
```
Usage: leshan-client-demo transport [-h] [COMMAND]
Configure Leshan transport layer. Can be used to add, remove or use different
implementation of supported LWM2M transport.

By default, Californium library is used for coap/coaps. Some Californium
parameters can be tweaked in Californium3.client.properties file.
Default behavior is like using :
transport californium coap coaps

More examples :
coaps support only based on Californium library :
transport californium coaps

coap support based on java-coap library :
transport java-coap coap

coaps support based on Californium library and coap support based java-coap
library:
transport californium coaps java-coap coap

Launch transport [californium|java-coap] -h for more details.

  -h, --help   Display help information.
Commands:
  californium  Configure transport based on Californium.
  java-coap    Configure transport based on java-coap.

```

This could sound a bit overkill but this will allow to have specific option for each transport.
E.g. connection id option should probably be moved have a `transport californium coaps` option

### DIFF
--- a/leshan-client-demo/pom.xml
+++ b/leshan-client-demo/pom.xml
@@ -37,6 +37,10 @@ Contributors:
       <groupId>org.eclipse.leshan</groupId>
       <artifactId>leshan-core-demo</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.leshan</groupId>
+      <artifactId>leshan-tl-javacoap-client</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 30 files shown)


sbernard31:
:x: Bad news, picocli behavior is not what I expected and so this way to go will not work... :grimacing: 

See : https://github.com/remkop/picocli/issues/2133 for more details. 

For now I didn't find any alternative... :disappointed:

Issue #1524, opened 2023-10-09
Error handling improvement in SendHandler


JaroslawLegierski:
### Question

Please find [here](https://github.com/eclipse-leshan/leshan/compare/master...JaroslawLegierski:leshan:opl/handling_error_improvement) small modification proposal in SendHandler. The main aim of this modification is based on new InvalidSendRequestException which could allow us to characterize the error better. What is your opinion about this change ?

JaroslawLegierski:
We can reuse InvalidRequestException instead of InvalidSendRequestException. This solution is also OK for us.

sbernard31:
Integrated in **master**, thx for contribution :pray:

PR #1526, opened 2023-10-10
Error handling improvement in SendHandler

JaroslawLegierski:
Small modification in SendHandler. The main aim of this modification is based on adding  InvalidRequestException in SendRequest processing.  #1524

### DIFF
--- a/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendHandler.java
+++ b/leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendHandler.java
@@ -24,6 +24,7 @@ import org.eclipse.leshan.core.node.LwM2mPath;
 import org.eclipse.leshan.core.node.TimestampedLwM2mNodes;
 import org.eclipse.leshan.core.peer.LwM2mPeer;
 import org.eclipse.leshan.core.request.SendRequest;
+import org.eclipse.leshan.core.request.exception.InvalidRequestException;
 import org.eclipse.leshan.core.response.SendResponse;
 import org.eclipse.leshan.core.response.SendableResponse;
 import org.eclipse.leshan.server.registration.Registration;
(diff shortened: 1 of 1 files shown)


leshan-server-core/src/main/java/org/eclipse/leshan/server/send/SendHandler.java
```diff
@@ -90,10 +91,11 @@ public void run() {
...
                     fireDataReceived(updatedRegistration, request.getTimestampedNodes(), request);
                 } else {
-                    onError(updatedRegistration, String.format("Invalid Send Request, server returns %s %s", //
+                    String errorMsg = String.format("Invalid Send Request, server returns %s %s", //
                             sendResponse.getCode().getName(), //
-                            sendResponse.getErrorMessage() != null ? "because" + sendResponse.getErrorMessage() : ""),
-                            null);
+                            sendResponse.getErrorMessage() != null ? "because" + sendResponse.getErrorMessage() : "");
+                    onError(updatedRegistration, errorMsg,
+                            new InvalidRequestException(sendResponse.getErrorMessage() != null ? errorMsg : ""));
```
sbernard31, line 98: I read it again and I'm not sure it make sense to have ""Invalid Send Request, server returns %s %s"" in exception description description :thinking: 

Maybe we should rather go with something like : 
```java
   onError(updatedRegistration, String.format("Invalid Send Request, server returns %s %s", //
                            sendResponse.getCode().getName(), //
                            sendResponse.getErrorMessage() != null ? "because" + sendResponse.getErrorMessage() : ""),
                            new InvalidRequestException(sendResponse.getErrorMessage() != null ? sendResponse.getErrorMessage() : "unknown reason"));
```

@JaroslawLegierski  WDYT ?

JaroslawLegierski: I think that is very good idea.

Issue #1566, opened 2023-12-19
🛡 Find a better alternative to dependabot !


sbernard31:
Currently we are using **dependabot** to check if we depends of libraries affected by known vulnerabilities. 

But there is some limitation, like not being able to check vulnerabilities on different branches.
(See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3949 for more details )

So this issue aims to centralize  research of using better open source alternative.

**Without talking about tooling itself there is 3 approach :** 

1. The obvious way ?  Checking dependencies only. 
2. The more secure way ? Checking whole build file systems.  (See : https://github.com/eclipse-leshan/leshan/pull/1546#issuecomment-1857418303)
3. The more clean way ?  Generate standard SBOM ([SPDX](https://spdx.dev/) and [CycloneDX](https://cyclonedx.org/)) then checking vulnerabilities from it and maybe later checking it with `dash-licences`. (See : https://github.com/eclipse/dash-licenses/issues/191#issuecomment-1858075528)

**Result :**

|  Alternative | Approach|  Advantage | Drawback | PR |
| - | - | - | - | - |
| [owasp](https://owasp.org/) - [dependency-check-maven](https://jeremylong.github.io/DependencyCheck/dependency-check-maven/index.html)  |  Dependencies Check |  :white_check_mark:  Maven integration  | :x: Maven integration not so good with [frontend-maven-plugin](https://github.com/eirslett/frontend-maven-plugin) see https://github.com/jeremylong/DependencyCheck/issues/6325  | 
| [trivy](https://aquasecurity.github.io/trivy) | FileSystem Check |  | :x: No maven Integration |  https://github.com/eclipse-leshan/leshan/pull/1546| 
| ? |  SBOM  |  |  |  |

sbernard31:
It seems that SBOM solution is better, so I integrated it in `master` and `1.x`, jenkins build is available at : https://ci.eclipse.org/leshan/job/leshan-weekly/

PR #1567, opened 2023-12-19
Use `dependency-check-maven` plugin to check dependencies vulnerabilities. 

sbernard31:
This is part of searching a way to check dependencies vulnerabilities, see https://github.com/eclipse-leshan/leshan/issues/1566

It works well with pure maven dependencies but not so good with [frontend-maven-plugin](https://github.com/eirslett/frontend-maven-plugin) see https://github.com/jeremylong/DependencyCheck/issues/6325.

Even if we can make it work with workaround, we can maybe found better ? :thinking:

### DIFF
--- a/.jenkins/test.jenkins
+++ b/.jenkins/test.jenkins
@@ -27,7 +27,7 @@ pipeline {
     jdk 'temurin-jdk11-latest'
   }
   options {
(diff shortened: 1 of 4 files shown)


sbernard31:
This was solution was not chosen,  see #1566 for more details.

PR #1569, opened 2024-01-04
Add Weekly Jenkins Build to check dependencies vulnerabilites with trivy

sbernard31:
This do the job (see https://github.com/eclipse-leshan/leshan/issues/1566) but it seems there is some maven dependencies which are not taking into account.  

(See possible limitation : https://github.com/aquasecurity/trivy/discussions/5874)

So maybe we should use more dedicated tool to generated SBOM (e.g. [cyclonedx-maven-plugin](https://github.com/CycloneDX/cyclonedx-maven-plugin)) and then use trivy (or something else) to check vulnerabilities.

### DIFF
--- a/.jenkins/weekly.jenkins
+++ b/.jenkins/weekly.jenkins
@@ -1,0 +1,52 @@
+/************************************************************************************
+  
+  Weekly Build : 
+  Checks for vulnerability
+  
+*************************************************************************************/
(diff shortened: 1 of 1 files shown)


sbernard31:
This was solution was not chosen,  see #1566 for more details.

PR #1570, opened 2024-01-05
Generate SBOM then use trivy to check vulnerabilities 

sbernard31:
This is still about 
-  #1566 

As : 
 - #1567 
 - #1569 
 
 was not a huge success :sweat_smile: .
 
So we first generate CycloneDX SBOM with : 
 - [cyclonedx-maven-plugin](https://github.com/CycloneDX/cyclonedx-maven-plugin) to generate SBOM about maven 
 - and we use trivy for frontend because we didn't find dedicated tools for yarn : see https://github.com/yarnpkg/berry/discussions/6063 (maybe we should consider to move to npm ? see : https://github.com/eclipse-leshan/leshan/issues/1550#issuecomment-1878802371)
 
So we have full control about what we put in SBOM.

Then we use trivy to scan vulnerabilities from this SBOMs.
Then one day we could also reuse that SBOM to check licenses with : https://github.com/eclipse/dash-licenses/issues/191


sbernard31:
(@jvermillard I will probably go with that way, let me know if you see blocking issue about that)

sbernard31:
@jvermillard , this is now integrated in `master` but feedback are still welcome. Nothing is set in stone. :slightly_smiling_face:

PR #1577, opened 2024-01-24
SENML JSON integration test

mgdlkundera:
SENML JSON test with reference to conversation:
https://github.com/eclipse-leshan/leshan/issues/1554

### DIFF
--- a/leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
+++ b/leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java
@@ -664,6 +664,42 @@ public class LwM2mNodeDecoderTest {
     }
 
     @Test
+    public void senml_json_timestamped_resources() throws CodecException {
+        // given
+        StringBuilder b = new StringBuilder();
+        b.append("[");
+        b.append("{\"bn\": \"/3303/0/5700\",");
+        b.append("\"bt\": 1699877805.766,");
+        b.append("\"v\": -128.8");
+        b.append("},");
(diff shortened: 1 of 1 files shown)


mgdlkundera:
In this structure we have 2 bn, so if we don't implement this test our leshan client will not work with leshan server in case of some future changes in code. We want to be sure that this data structure  will be supported in the future that is why we should add this dedicated test.

sbernard31:
> We want to be sure that this data structure will be supported in the future 

I understand.

> In this structure we have 2 bn, 

I'm not sure what you have in mind but there are tests in `LwM2mNodeDecoderTest` with 2 `bn`. 

So please, could you have another quick look at [existing tests](https://github.com/eclipse-leshan/leshan/blob/3a1adfea5c06c23f0a73e5339f71a76bfca0e58e/leshan-core/src/test/java/org/eclipse/leshan/core/node/codec/LwM2mNodeDecoderTest.java#L896-L1369)

Just to be sure we don't add useless tests.

Then If after looking at this, you still think that current tests are not enough, then we can add your tests. :slightly_smiling_face:

mgdlkundera:
I looked through the existing tests again and there is no test with this structure, so could we add my test? 🙂

sbernard31:
OK, so I integrated this in `master` (commit 777d2ed4d63d30ebfc483e45b61ae74ce466c61e)

I did some little changes : 
- I moved the test with other senml tests
- I renamed it
- I get node by timestamp (to ensure timestamp are well decoded)
- I rename commit title to refer to original issue : https://github.com/eclipse-leshan/leshan/issues/1554

mgdlkundera:
Tkank you 🙂

Issue #1574, opened 2024-01-14
How are Observations stored on client side?


Ozame:
### Question

We are using the leshan to simulate thousands of devices/clients. Since performance-wise one of the limiting factors is the use of threads, we use shared thread pools and stop clients when the devices are sleeping. This approach has some issues, one such seems to be that when the LeshanClient is stopped, it seems to lose all the observations sent by the server. When testing the client with Leshan Server Demo, the observed values are updated nicely on the server if the client is not stopped. When stopped and started again, no notifications are sent when the observed resources are changed, unless the server re-observes them.

My question is, how are those observations stored on the client side, is there some way we could save these so that the client restart would not affect them?

sbernard31:
Could I ask which version of Leshan you are using ? 

> Since performance-wise one of the limiting factors is the use of threads, we use shared thread pools 

Yep there was some improvement about that but still some limitations. (code was not initially thought with this use case in mind. Neither in Leshan nor Californium)

See  for more details  at : 
 - About Leshan : https://github.com/eclipse-leshan/leshan/issues/491
 - About californium : https://github.com/eclipse-californium/californium/issues/1203
 - About Leshan too but probably less useful so I added just in case : https://github.com/eclipse-leshan/leshan/issues/1406
 
> and stop clients when the devices are sleeping. 

Just by curiosity what kind of gain to you get with stop() ? 

 >  When testing the client with Leshan Server Demo, the observed values are updated nicely on the server if the client is not stopped. When stopped and started again, no notifications are sent when the observed resources are changed, unless the server re-observes them.
 
 I guess this is because when you start the client it send a new Register Request.
 And AS Observations are tied to 1 registration WHEN the server receive new register request THEN it create a new registration and drop the previous one with all its tied observation. 

> My question is, how are those observations stored on the client side, is there some way we could save these so that the client restart would not affect them?

If I'm right :point_up: this is not about saving the observations its about not sending a new registration on start.

sbernard31:
Are you using `coap` or `coaps` ?

sbernard31:
> This seems to have fixed the behavior as needed, the same registration is kept.

And it works for you now ? 
Maybe we could add this to Leshan client (maybe optionally) ? :thinking: 

> but I feel that if it is the client who sends the notifications on resource change, it should have knowledge of the observations that have been done to it's resources. 

[...]

PR #1579, opened 2024-02-02
Have DefaultRegistrationEngine try to update reg after start

Ozame:
As discussed in #1574 , some changes are needed in the Default implementation of RegistrationEngine to support the stopping and starting of Leshan client without re-registration happening every time. 

Basically the list of registered servers is checked when start-method is called, and registration update is tried. 

*Note* Will still need to at least update the tests if this is deemed merge-worthy.

### DIFF
--- a/leshan-client-core/src/main/java/org/eclipse/leshan/client/engine/DefaultRegistrationEngine.java
+++ b/leshan-client-core/src/main/java/org/eclipse/leshan/client/engine/DefaultRegistrationEngine.java
@@ -179,16 +179,28 @@ public class DefaultRegistrationEngine implements RegistrationEngine {
         stop(false); // Stop without de-register
         synchronized (this) {
             started = true;
-            // Try factory bootstrap
-            // TODO support multi server
-            LwM2mServer dmServer = factoryBootstrap();
-
+            // TODO support multiple servers
+            LwM2mServer dmServer;
+            if (registeredServers.isEmpty()) {
+                // Try factory bootstrap
+                dmServer = factoryBootstrap();
+            } else {
+                dmServer = registeredServers.entrySet().iterator().next().getValue();
+            }
             if (dmServer == null) {
                 // If it failed try client initiated bootstrap
                 if (!scheduleClientInitiatedBootstrap(NOW))
                     throw new IllegalStateException("Unable to start client : No valid server available!");
             } else {
-                registerFuture = schedExecutor.submit(new RegistrationTask(dmServer));
+                // If there exists a registered server already, we try to send a registration update
+                if (!registeredServers.isEmpty()) {
+                    String registrationId = registeredServers.entrySet().iterator().next().getKey();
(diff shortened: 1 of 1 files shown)


sbernard31:
(I will look at this next week)

sbernard31:
I looked at this and regarding my comments at https://github.com/eclipse-leshan/leshan/issues/1574#issuecomment-1898829914,

> I have no clear idea if this is something which should be integrated OR not.
I'm not sure there is obvious use case but maybe this could be a step in adding support to Queue/Offline mode ? 🤔

I still don't know if this should be added.

Pro : 
- the code modification is pretty simple (but we need to add a configuration set way + new tests)n
- it supports your use case,
- maybe (really not sure) it could be a first step for implementing Queue mode :thinking: 

Con :
- not sure this will be used too much without a real Queue mode feature ? :thinking: 

> The another point, I'm not sure if it can work with all transport layer java-coap implementation ? (because start stop destroy is mainly a californium concept, some libraries has just start/destroy concept)

I adapt some tests to check if it could works with `java-coap` and it seems OK. (See d0f5e2f2030538f07c8040ab92d65cf3156e5b1d). So this point is not really an issue. 

Note that if we wanted to support this feature we should rather add new tests instead of modifying existing one.

sbernard31:
An alternative to this PR : https://github.com/eclipse-leshan/leshan/issues/1574#issuecomment-1929276711

leshan-client-core/src/main/java/org/eclipse/leshan/client/engine/DefaultRegistrationEngine.java
```diff
@@ -179,16 +179,28 @@ public void start() {
...
+            } else {
+                dmServer = registeredServers.entrySet().iterator().next().getValue();
+            }
             if (dmServer == null) {
                 // If it failed try client initiated bootstrap
                 if (!scheduleClientInitiatedBootstrap(NOW))
                     throw new IllegalStateException("Unable to start client : No valid server available!");
             } else {
-                registerFuture = schedExecutor.submit(new RegistrationTask(dmServer));
+                // If there exists a registered server already, we try to send a registration update
```
sag-eweingaertner, line 195: I'd actually be sceptical here. If a client is stopped, it should not keep registrations. So this situation should not be allowed. 

If somebody truly needs a LWM2M client that supports this, they can either implement their own registration engine or subclass it.

I am not in favor of doing this.

sbernard31: Thx for taking time to look at this and also for sharing your opinion. :pray: 

For now, no plan to integrate this BUT if for some reason we decide to ingrate it  this will not be the default behavior.

sbernard31:
I closed this PR for now as I'm not sure we will go in that direction.
We have still some opened issue about that so we will not forget that topic.
Still possible to reopen later if needed.

PR #1586, opened 2024-02-20
Tests write attribute

mgdlkundera:
Following tests:
- test about epmin, epmax,
- test about pmin = pmax,
- test where we use several attribute together,
- test to be sure that attribute inheritance works.


sbernard31:
@mgdlkundera if there is something wrong with one of your PR,  most of the time there is no reason to close it and reopen a new one for same subject. You can just edit/fix your branch then push force it.

PR #1608, opened 2024-04-19
Implementation of read response with timestamped SenML

mgdlkundera:
Implementation of read response with timestamped SenML


sbernard31:
Please in the future, do not close a PR to create more or less the same one. Just edit your branch localy and then push --force.

Like explained : https://github.com/eclipse-leshan/leshan/pull/1608#issuecomment-2066375468

If you are not comfortable with git/github, you can eventually make "test"  on your own fork of the project and create PR on it.

PR #1621, opened 2024-06-07
Fix some behavior about Timestamped Node in SenML and Old JSON content Format

sbernard31:
This mainly fixes issue listed at : https://github.com/eclipse-leshan/leshan/issues/1612#issuecomment-2151914267

### DIFF
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/json/LwM2mNodeJsonDecoder.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/json/LwM2mNodeJsonDecoder.java
@@ -233,7 +233,9 @@ public class LwM2mNodeJsonDecoder implements TimestampedNodeDecoder {
     /**
      * Group all JsonArrayEntry by time-stamp
      *
-     * @return a map (relativeTimestamp => collection of JsonArrayEntry)
+     * @return a sorted map (relativeTimestamp => collection of JsonArrayEntry) order by descending time-stamp (most
+     *         recent one at first place). If null time-stamp (meaning no time information) exists it always at first
+     *         place.
      */
     private SortedMap<BigDecimal, Collection<JsonArrayEntry>> groupJsonEntryByTimestamp(JsonRootObject jsonObject) {
         SortedMap<BigDecimal, Collection<JsonArrayEntry>> result = new TreeMap<>(new Comparator<BigDecimal>() {
@@ -243,9 +245,9 @@ public class LwM2mNodeJsonDecoder implements TimestampedNodeDecoder {
                 // - supports null (time null means 0 if there is a base time)
                 // - reverses natural order (most recent value in first)
                 if (o1 == null) {
-                    return o2 == null ? 0 : 1;
+                    return o2 == null ? 0 : -1;
                 } else if (o2 == null) {
-                    return -1;
+                    return 1;
                 } else {
                     return o2.compareTo(o1);
                 }
--- a/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/json/LwM2mNodeJsonDecoder.java
+++ b/leshan-core/src/main/java/org/eclipse/leshan/core/node/codec/json/LwM2mNodeJsonDecoder.java
@@ -92,6 +92,9 @@ public class LwM2mNodeJsonDecoder implements TimestampedNodeDecoder {
             if (timestampedNodes.size() == 0) {
                 return null;
             } else {
+                TimestampedLwM2mNode timestampedNode = timestampedNodes.get(0);
+                validateNoTimestampedValue(timestampedNode, path);
+
                 // return the most recent value
                 return (T) timestampedNodes.get(0).getNode();
             }
@@ -109,6 +112,12 @@ public class LwM2mNodeJsonDecoder implements TimestampedNodeDecoder {
             return parseJSON(json, path, model, nodeClass);
         } catch (LwM2mJsonException | InvalidLwM2mPathException e) {
             throw new CodecException(e, "Unable to deserialize json [path:%s]", path);
+        }
+    }
+
+    protected void validateNoTimestampedValue(TimestampedLwM2mNode timestampedNode, LwM2mPath path) {
+        if (timestampedNode.getTimestamp() != null) {
+            throw new CodecException("Unable to decode node[path:%s] : value should not be timestamped", path);
         }
     }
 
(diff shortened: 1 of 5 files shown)


sbernard31:
@slaft let me know if that works for you. :slightly_smiling_face:

slaft:
Thank you @sbernard31

So doing:
```java
public class CustomLwM2mNodeSenMLDecoder extends LwM2mNodeSenMLDecoder {
    public CustomLwM2mNodeSenMLDecoder(SenMLDecoder decoder, boolean permissiveNumberConversion) {
        super(decoder, permissiveNumberConversion);
    }

    public CustomLwM2mNodeSenMLDecoder(SenMLDecoder decoder, LinkParser linkParser, boolean permissiveNumberConversion) {
        super(decoder, linkParser, permissiveNumberConversion);
    }

    @Override
    protected void validateNoTimestampedRecord(LwM2mResolvedSenMLRecord resolvedRecord) {
        // Do nothing
    }
}
```
```java
leshanServerBuilder.setDecoder(buildPatchedDecoder());

private LwM2mDecoder buildPatchedDecoder() {
        Map<ContentFormat, NodeDecoder> nodeDecoders = DefaultLwM2mDecoder.getDefaultNodeDecoders(false);
        nodeDecoders.put(ContentFormat.SENML_JSON, new CustomLwM2mNodeSenMLDecoder(new SenMLJsonJacksonEncoderDecoder(), true));
        nodeDecoders.put(ContentFormat.SENML_CBOR, new CustomLwM2mNodeSenMLDecoder(new SenMLCborUpokecenterEncoderDecoder(), false));
        return new DefaultLwM2mDecoder(nodeDecoders, DefaultLwM2mDecoder.getDefaultPathDecoder());
    }
```


Allows me to accept a Read-Composite/Observe-Composite response with timestamps (which was accepted by default before).

---
Just to be sure, this does not provide a way to accept this payload right? (discussed here https://github.com/eclipse-leshan/leshan/issues/1553#issuecomment-2150530588)

Read/Observe on `/3442/0`:
```json
[
  {"bn":"/3442/0/","n":"120","t":1717150530.123456789,"v":1},
  {"n":"110","vs":"a"}
]
```

sbernard31:
@slaft  You're right.

If you want to support : 
```json
[
  {"bn":"/3442/0/","n":"120","t":1717150530.123456789,"v":1},
  {"n":"110","vs":"a"}
]
```

This should work : 
```java
public class CustomLwM2mNodeSenMLDecoder extends LwM2mNodeSenMLDecoder {
    public CustomLwM2mNodeSenMLDecoder(SenMLDecoder decoder, boolean permissiveNumberConversion) {
        super(decoder, permissiveNumberConversion);
    }

    public CustomLwM2mNodeSenMLDecoder(SenMLDecoder decoder, LinkParser linkParser,
            boolean permissiveNumberConversion) {
        super(decoder, linkParser, permissiveNumberConversion);
    }

    @Override
    protected void validateNoTimestampedRecord(LwM2mResolvedSenMLRecord resolvedRecord) {
        // Do nothing
    }

    @Override
    public List<TimestampedLwM2mNode> decodeTimestampedData(byte[] content, LwM2mPath path, LwM2mModel model,
            Class<? extends LwM2mNode> nodeClass) throws CodecException {
        LwM2mNode node = decode(content, path, model, nodeClass);
        return Arrays.asList(new TimestampedLwM2mNode(null, node));
    }
}
```
(but you will lost timestamp data)

slaft:
Thank you @sbernard31

If my understanding is correct, this will also impact historical notifications for single Observations.
(No impact on timestamped data on composite Observations or Send however.)

sbernard31:
> If my understanding is correct, this will also impact historical notifications for single Observations.
(No impact on timestamped data on composite Observations or Send however.)

That's right.

It's still possible to write a more smarter version of :
```java
@Override
public List<TimestampedLwM2mNode> decodeTimestampedData(byte[] content, LwM2mPath path, LwM2mModel model,
        Class<? extends LwM2mNode> nodeClass) throws CodecException {
    LwM2mNode node = decode(content, path, model, nodeClass);
    return Arrays.asList(new TimestampedLwM2mNode(null, node));
}
```
in your `CustomLwM2mNodeSenMLDecoder`. 

But I can help because I'm sure what should mean : 
```
[
  {"bn":"/3442/0/","n":"120","t":1717150530.123456789,"v":1},
  {"n":"110","vs":"a"}
]
```
And how we can express it with current LWM2M node model in Leshan. 


The other solution is maybe to revert https://github.com/eclipse-leshan/leshan/pull/1610.
I think we should seriously thing about that because if that was an error better to rollback now. 
If we rollback that just doing : 
```java
public class CustomLwM2mNodeSenMLDecoder extends LwM2mNodeSenMLDecoder {
    public CustomLwM2mNodeSenMLDecoder(SenMLDecoder decoder, boolean permissiveNumberConversion) {
        super(decoder, permissiveNumberConversion);
    }

    public CustomLwM2mNodeSenMLDecoder(SenMLDecoder decoder, LinkParser linkParser, boolean permissiveNumberConversion) {
        super(decoder, linkParser, permissiveNumberConversion);
    }

    @Override
    protected void validateNoTimestampedRecord(LwM2mResolvedSenMLRecord resolvedRecord) {
        // Do nothing
    }
}
```
will ignore timestamp for composite/read/observe response and historical notification will be kept for notification. 
(No impact on timestamped data on composite Observations or Send however.)

mgdlkundera:
In our case best solution would be to take the latest timestamp value
for object/object instance with many timpestamped values 
in a read and observe and also for a composite operations.

Ignoring timestamp is not a good aproach for us,
so would it be possible to get only one latest timestamp value? What do you think about this idea?

sbernard31:
> Ignoring timestamp is not a good aproach for us, so would it be possible to get only one latest timestamp value?
What do you think about this idea?

I understand that maybe this solves your particular problem but I can not see how this could be a good API/behavior for Leshan.

I'm a bit frustrating because I feel (maybe I'm wrong, I hope I'm wrong) that you try to push your needs without taking into account what is good for Leshan project itself without trying to maintain or improve consistency in the project. 

So if your want this behavior, I think you can implement it more or less easily in a `CustomLwM2mNodeSenMLDecoder`

I will probably take time again to summarize issues about timestamp even if I'm not sure I'm understandable... :disappointed:

sbernard31:
I try to clarify my though here : https://github.com/eclipse-leshan/leshan/issues/1623

sbernard31:
If you read :point_up:, you will see that I think we take so much time to try to solve a model/client behavior issue.

FMPOV, the right way to express **Sensor Measurement Timestamp** would be to use/create a dedicated resource in your object model. I strongly encourage you to try to change  (or ask to change) your client behavior.

Waiting and even if I would prefer to revert https://github.com/eclipse-leshan/leshan/pull/1610,  I can understand that you could need a temporary workaround solution.

So finally maybe current state of the code is a good compromise : 
- we try to keep more or less a kind of API/behavior consistency in Leshan. 
- it allow you to implement your own behavior without rewritten too much code. 

So let me know if I should relaunch the 2.0.0-M15 process OR if you're still stuck about that.

sbernard31:
I integrate this is `master`. So it will be tested by sandbox user.

mgdlkundera:
Lets release M15 current implementation, without rollback.

sbernard31:
@mgdlkundera thank you for letting me know that  :pray:

Issue #1629, opened 2024-07-05
Plan for 2.0.0-M16 🎯


sbernard31:
This issue is about preparing the **2.0.0-M16** release.

This should contain only : 
 - [x] https://github.com/eclipse-leshan/leshan/issues/1295
 - [x] Integration of californium 3.12.1 (to fix https://github.com/eclipse-leshan/leshan/issues/1627)
 - [x] https://github.com/eclipse-leshan/leshan/pull/1630
 - maybe some minor bug fixed in we found some.

To make the migration more smooth.

**What is currently included for M16:**

New commits in master since M15 :

```
e673ac753: [ci] Add build artifacts to test jobs.
1377e5262: Integration of Californium 3.12.1
f9137bb00: Clean revapi configuration.
3f9b3f7ad: Change module <name> to avoid them to be too long (ugly mvn output)
8d434d42d: Refactoring add oscore module - edit file content
d5c81f1ea: Refactoring add oscore module - no file content changes [dont build]
bc9031447: Separate DeviceManagement and Bootstrap Request class hierarchy.
ab4f9a457: Refactoring add bsserver module - edit file content
286962589: Refactoring add bsserver module - no file content changes [dont build]
6d2ff69c7: Refactoring rename module/package - edit file content
34caf241c: Refactoring rename module/package - no file content changes [dont build]

```

**Redis Store Compatibility break :**
No break.

**When ?**

Californium 3.12.1 will be release the 9th July (https://github.com/eclipse-californium/californium/issues/2258)
I will be unavailable from 11th to 24th July.

So release will be at the earliest end of July.

sbernard31:
@cyril2maq, @slaft, @niklas-sparfeld-gcx, @jvermillard, @JaroslawLegierski, @mgdlkundera, @jakubsobolewskisag

`master` is probably near to the state of what will be released in 2.0.0-M16. (except we find new critical bug) 
So you can test `master` if wanted
This https://github.com/eclipse-leshan/leshan/issues/1295#issuecomment-2211032195 could help for migration. 

Do not hesitate to share any concerns.

(I'm back on 24th)

niklas-sparfeld-gcx:
The refactoring changed the sender interface to take a `DownlinkDeviceManagementRequest` instead of `DownlinkRequest`. This is of course fine, it's just that you didn't mention it here and I think it is an API change.

sbernard31:
[...]

PR #1630, opened 2024-07-10
Fix TlvEncoder Date encoding for dates in a distant future (exceeding MAX_INT)

jakubsobolewskisag:
If encoded date is in distant future (number of seconds since 1970 exceeds MAX_INT), TlvEncoder breaks because it always tries to encode dates as 4-byte integers. This fix checks whether number of seconds exceeds MAX_INT, and if it does the value is encoded in 8-bytes. Otherwise, previous behavior will be used (4-bytes).

### DIFF
--- a/leshan-lwm2m-core/src/main/java/org/eclipse/leshan/core/tlv/TlvEncoder.java
+++ b/leshan-lwm2m-core/src/main/java/org/eclipse/leshan/core/tlv/TlvEncoder.java
@@ -115,7 +115,13 @@ public class TlvEncoder {
     public static byte[] encodeDate(Date value) {
         ByteBuffer tBuf;
         long lValue = value.getTime() / 1000L;
-        if (lValue <= Integer.MAX_VALUE) {
+        if (lValue <= Byte.MAX_VALUE && lValue >= Byte.MIN_VALUE) {
+            tBuf = ByteBuffer.allocate(1);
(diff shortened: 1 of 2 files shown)


sbernard31:
Thx for the contribution :pray: 

Looking at [LWM2M-v1.1.1@core§Table: C.-2 Data Type Mapping](https://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#Table-C-2-Data-Type-Mapping) 
about Time for TLV : 

> Same representation as Integer.

Does it means that `Date` encoding should look more like : https://github.com/eclipse-leshan/leshan/blob/f9137bb00285ffc5a519362f4edac189c6a24349/leshan-lwm2m-core/src/main/java/org/eclipse/leshan/core/tlv/TlvEncoder.java#L64-L81 ?

jakubsobolewskisag:
@sbernard31 done - I think I covered all of the edge cases in the unit tests.

sbernard31:
This is now integrated in `master` (squashed in 1 commit 2a1a8af1f6a7393fdd66f0d7ca3a10edd8bdc1f3), it should be soon available leshan sandbox and will be part of 2.0.0-M16 (#1629)

Thx @jakubsobolewskisag :pray:

PR #1647, opened 2024-09-05
Use BigDecimal instead of double  for decimal LWM2M Attribute

sbernard31:
This aims to solve : https://github.com/eclipse-leshan/leshan/issues/1583


sbernard31:
(will be available in 2.0.0-M17)

PR #1655, opened 2024-10-11
Ensure that client use same endpoint for a given registration

sbernard31:
This was discuss at https://github.com/eclipse-leshan/leshan/issues/1415


sbernard31:
@JaroslawLegierski , @cyril2maq, It could be worth to look at this too.

Issue #1418, opened 2023-03-10
 Redis Store Migration from 2.0.0-M10 to 2.0.0-M11


sbernard31:
## Changes :

Between 2.0.0-M10 and 2.0.0-M11, there is lot of changes in Redis Store I tried to list it just below :point_down:  : 
| Store                                        | old key/prefix     | new key/prefix                                | Data compatibility break |  PR                     |
|-----------------------------------------|-----------------------|----------------------------------------------|----------------------------------|-----------------------|
| **RedisSecurityStore**            |`SEC#EP#`         |  `SECSTORE#SEC#EP#`             | compatible                       | #1398                 |
|                                                 |`PSKID#SEC`     |  `SECSTORE#EP#PSKID`            | compatible                       | #1398                 |
| **RedisRegistrationStore**      | `REG:EP:`         |  `REGSTORE#REG#EP#`             | not compatible                 | #1414 #1318     |
|                                                 | `EP:REGID:`      |  `REGSTORE#EP#REGID#`         | compatible                       | #1414                |
|                                                 | `EP:ADDR:`        |  `REGSTORE#EP#ADDR#`          | compatible                      | #1414                 |
|                                                 | `EP:IDENTITY:`  |  `REGSTORE#EP#IDENTITY#`    | compatible                       | #1414                |
|                                                 | `LOCK:EP:`         |  `REGSTORE#LOCK#EP#`          | compatible                       | #1414                |
|                                                 | `OBS:TKN:`         |  `REGSTORE#OBS#TKN#`          | not compatible                 | #1414 #1318     |
|                                                 | `TKNS:REGID:`   |  `REGSTORE#TKNS#REGID#`   | compatible                       | #1414                |
|                                                 | `EXP:EP`             |  `REGSTORE#EXP#EP`              | compatible                       | #1414                |

## Recommended Migration Way : 

For  **RedisSecurityStore** : 
1. use old keys : `new RedisSecurityStore.Builder(jedis).setPrefix(null).setEndpointByPskIdKey("PSKID#SEC").build();`
2. OR write a migration script which renames keys (let us know if you write something like this)

For **RedisRegistrationStore**, as registration is just kind of "session", probably better to just drop all data and so force client to re-register.
_Note that key name changed, so if you did nothing, registration will not be found and so client will re-register but old data will not be deleted._
Looking at  https://github.com/eclipse/leshan/issues/1401#issuecomment-1462328493 could help. 

## Will it change again soon ? 

[...]

Issue #1415, opened 2023-03-09
Should Registration be strictly tied to a server endpoint ?


sbernard31:
With new [Transport Layer Abstraction](https://github.com/eclipse/leshan/issues/1025#issuecomment-1373824399), a sever can have several endpoints (opened socket address) that client can use to connect to it. 
A new field  `lastEndpointUsed` was added to the `Registration` to know which server endpoint is used by the client.

When I created this field, I had in mind that maybe client could connect to another server endpoint and just do a registration update.
But finally, I think that this could lead to some issues (e.g. for observation) and I'm not even sure if this is allowed by LWM2M specification.

Maybe it would be better to strictly tied the registration to only 1 server endpoint. In that case, if client want to use another server endpoint, it must create a new registration.

Do you have any opinion on this ? (or knowledge about this part of the specification)

If we agree that registration must be tied to only one server endpoint, so we probably need a better name ? 
`endpointUsed`, `serverEndpoindUsed` any idea ?

jvermillard:
The memory usage:
```
    // Components of all URIs: [<scheme>:]<scheme-specific-part>[#<fragment>]
    private transient String scheme;            // null ==> relative URI
    private transient String fragment;

    // Hierarchical URI components: [//<authority>]<path>[?<query>]
    private transient String authority;         // Registry or server

    // Server-based authority: [<userInfo>@]<host>[:<port>]
    private transient String userInfo;
    private transient String host;              // null ==> registry-based
    private transient int port = -1;            // -1 ==> undefined

    // Remaining components of hierarchical URIs
    private transient String path;              // null ==> opaque
    private transient String query;

    // The remaining fields may be computed on demand, which is safe even in
    // the face of multiple threads racing to initialize them
    private transient String schemeSpecificPart;
    private transient int hash;        // Zero ==> undefined

    private transient String decodedUserInfo;
    private transient String decodedAuthority;
    private transient String decodedPath;
    private transient String decodedQuery;
    private transient String decodedFragment;
    private transient String decodedSchemeSpecificPart;
```

sbernard31:
>If we agree that registration must be tied to only one server endpoint, so we probably need a better name ?
endpointUsed, serverEndpoindUsed any idea ?

Any ideas ?

sbernard31:
Good morning from France :wink: 

> I would like to join this discussion because we are at a state right now, where we would like to run multiple instances of our Leshan server, but make this transparent for the client.

[...]

PR #1656, opened 2024-10-11
Strongly tie a observation to a server endpoint

sbernard31:
A registration is now tied to  server endpoint. (see https://github.com/eclipse-leshan/leshan/pull/1655)
Then observation should also be tied to 1 server endpoint. 

This will avoid Observation Token conflict between different server endpoint. (currently a token should be unique by Leshan Server with this modification a Observation Token should be unique by server endpoint)

So concretely that mean that Observation Identifier has changed.  
Before it was just a CoAP token, now this is the (CoAP token + server endpoint Uri used).
Unfortunately this will break RedisRegistrationStore data compatibility. (see https://github.com/eclipse-leshan/leshan/issues/1418#issuecomment-2419924978)


sbernard31:
@JaroslawLegierski , @cyril2maq, Oops I did it again. I break the RedisRegistrationStore Data compatibility.

I let you know that as soon as possible, so we can try to anticipate issue about it.

Maybe better to read this comment : https://github.com/eclipse-leshan/leshan/issues/1415#issue-1617480570 too. 

Let discuss about that here.

JaroslawLegierski:
> @JaroslawLegierski , @cyril2maq, Oops I did it again. I break the RedisRegistrationStore Data compatibility.
> 
> I let you know that as soon as possible, so we can try to anticipate issue about it.
> 
> Maybe better to read this comment : [#1415 (comment)](https://github.com/eclipse-leshan/leshan/issues/1415#issue-1617480570) too.
> 
> Let discuss about that here.

I tested this change today and as I understand it, the change in Redis is the addition the field e.g. **"epUri": "coaps://0.0.0.0:5684"**, in JSON presented below for the key **"OBS#TKN#\...."** responsible for observation for partitcular LwM2M client
```json
		{
			"id": "d53329b026af5dd2",
			"epUri": "coaps://0.0.0.0:5684",
			"regid": "mIqhc2qKmg",
			"userContext": {},
			"protocolData": {
				"leshan-cf-obs": {
					"request": "48018072d53329b026af5dd260543333303301300435373030622d17",
					"peer": {
						"address": "127.0.0.1",
						"port": 50926,
						"id": "leshan_1",
						"attributes": {
							"*DTLS_HANDSHAKE_MODE": "auto"
						}
					},
					"context": {
						"leshan-endpoint": "leshan1",
						"leshan-path": "/3303/0/5700n",
						"leshan-regId": "mIqhc2qKmg}}"
					},
					"kind": "single",
					"ct": 11543,
					"path": "/3303/0/5700"
				}
			}
		}
```
So the change only affects leshan clients with observation. Am I right?

sbernard31:
This is a bit more than that : see https://github.com/eclipse-leshan/leshan/issues/1418#issuecomment-2419924978

> So the change only affects leshan clients with observation. Am I right?

Those keys are only about observations.
But I don't know what happens if you are using an old registration using previous key with the new code in case where you don't use observations (I didn't test it)

JaroslawLegierski:
> This is a bit more than that : see [#1418 (comment)](https://github.com/eclipse-leshan/leshan/issues/1418#issuecomment-2419924978)
> 
> > So the change only affects leshan clients with observation. Am I right?
> 
> Those keys are only about observations. But I don't know what happens if you are using an old registration using previous key with the new code in case where you don't use observations (I didn't test it)

I tested the code change on the server side. After retransmissions (we lost InMemoryReadWriteLockConnectionStore on server site after server replace), the LwM2M client (Leshan client demo) initiated a new handshake and finally successfully completed the registration update.
leshan_server_code_change

sbernard31:
`OBSIDS#REGID#*`has a list value. I don't know if `MGET` command works on it. 
In Leshan code we use [`LRANGE`](https://redis.io/docs/latest/commands/lrange/).

JaroslawLegierski:
> `OBSIDS#REGID#*`has a list value. I don't know if `MGET` command works on it. In Leshan code we use [`LRANGE`](https://redis.io/docs/latest/commands/lrange/).
Sure I completely forgot that this is list of values. LRAGE reads it. Thx!

Issue #1257, opened 2022-05-25
OSCORE fallback use cases


sbernard31:
Using OSCORE, I found a pretty simple use case where the client will not be able to connect to the server anymore.

- I launch a `leshan-server-demo`
- I create oscore config for my device in the UI
- I launch a `leshan-client-demo` using the corresponding OSCORE setting
- simulate a crash (`kill -9 PID` on linux), this way the client doesn't send a deregister.
- then I try to relaunch the client (same OSCORE config)
- and I get a : `Registration failed: UNAUTHORIZED(401) Replay detected.`

@rikard-sics any idea about this ? and how we could handle this ?

rikard-sics:
Spontaneous reaction: This should be solved by making sure the Client uses the Appendix B.2 procedure defined in the OSCORE RFC here:
https://datatracker.ietf.org/doc/html/rfc8613#appendix-B.2
It will re-establish a new OSCORE Security Context, most importantly new Sender/Recipient Keys and a fresh replay window on the server. By doing that the Client restarting from Sender Sequencer Number 0 is not Key and Nonce reuse, and not a replay either. (Note that the Appendix B.2 procedure changes the ID Context but keeps the same Sender/Recipient IDs).

So basically if the client loses mutable state information in the context (like the Sender Sequence Number), it must use this procedure the first time it resumes communication with the server. I believe in the current code the Client will at least use this procedure when it registers to the server.

That being said I need to investigate this case to understand if the Client is trying to use Appendix B.2 in this case. If not, using it should solve this. If the client is trying to use it already, we need to investigate things further.

sbernard31:
> This should be solved by making sure the Client uses the Appendix B.2 procedure defined in the OSCORE RFC here:
https://datatracker.ietf.org/doc/html/rfc8613#appendix-B.2

I don't know what should be done in the code to be sure that appendex B.2 is used ? :thinking: 

> I believe in the current code the Client will at least use this procedure when it registers to the server.

I think it makes sense.
But as you said that _"Note that the Appendix B.2 procedure changes the ID Context "_
What the purpose to set the Context ID in OSCORE object (21) at bootstrap time :thinking:  ?

sbernard31:
@rikard-sics I would really appreciate if we can find a fix (or at least workaround) for this. :pray:

sbernard31:
I try to investigate on this.

> I don't know what should be done in the code to be sure that appendex B.2 is used ? 

[...]

PR #1694, opened 2025-02-11
Enable usage of OSCORE Appendix B.2 context re-derivation procedure

rikard-sics:
This pull request adds support for running the OSCORE context re-derivation procedure from Appendix B.2 of the OSCORE RFC. This procedure securely re-generates OSCORE Security Context information for both peers. The LWM2M client now uses this procedure towards the LWM2M server and LWM2M bootstrap server.
See https://tools.ietf.org/html/rfc8613#appendix-B.2

It is specified in the LWM2M 1.1 Transport Bindings document section 5.5.3 that Appendix B.2 of OSCORE should be used. Using the procedure in Appendix B.2 will derive a new OSCORE Security Context (with new Sender and Recipient keys). The benefit this has is that if a LWM2M client reboots and starts using the same Security Context that it was originally configured with, it will not be using the same Sender Key while starting over from sequence number 0 (thus performing nonce and key reuse). Rather it will first run Appendix B.2 to generate a new Context (Sender and Recipient keys) with the LWM2M Server or LWM2M Bootstrap server.

Every time the client connects the first time using OSCORE to a LWM2M Server or LWM2M Bootstrap server, Appendix B.2 will be run. Furthermore, currently if the client is restarted but the server is not, the server will complain about replayed messages. But since Appendix B.2 refreshes the security context (creating new keys and resetting the replay window) this problem will no longer exist.

### DIFF
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/oscore/OscoreTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/oscore/OscoreTest.java
@@ -231,10 +231,6 @@ public class OscoreTest {
 
         // Start it and wait for registration
         client.start();
-        // TODO to remove
-        // Uncomment to see the failure
-//        Failure failure = client.waitForRegistrationFailureTo(server);
-//        System.out.println(failure);
 
         server.waitForReRegistrationOf(registration);
 
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/oscore/OscoreTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/oscore/OscoreTest.java
@@ -197,4 +197,53 @@ public class OscoreTest {
         client.triggerRegistrationUpdate();
         client.waitForUpdateFailureTo(server, 3, TimeUnit.SECONDS);
     }
+
+    @Test
+    public void registered_device_with_oscore_to_server_with_oscore_then_stop_device_then_register_again()
+            throws NonUniqueSecurityInfoException, InterruptedException {
+
+        // This should trigger Appendix B.2 procedure from OSCORE RFC.
+
+        // Create OSCORE Client
+        client = givenClient.connectingTo(server)//
+                .using(getClientOscoreSetting()).build();
+
+        // Add client credentials to the server
+        server.getSecurityStore().add(SecurityInfo.newOscoreInfo(client.getEndpointName(), getServerOscoreSetting()));
+
+        // Check client is not registered
+        assertThat(client).isNotRegisteredAt(server);
+
+        // Start it and wait for registration
+        client.start();
+        server.waitForNewRegistrationOf(client);
+
+        // Check client is well registered
+        assertThat(client).isRegisteredAt(server);
+        Registration registration = server.getRegistrationFor(client);
+
+        // check we can send request to client.
+        ReadResponse response = server.send(registration, new ReadRequest(3, 0, 1), 500);
+        assertThat(response.isSuccess()).isTrue();
+
+        // Stop client without sending deregister request
+        client.stop(false);
+
+        // Start it and wait for registration
+        client.start();
+        // TODO to remove
+        // Uncomment to see the failure
+//        Failure failure = client.waitForRegistrationFailureTo(server);
+//        System.out.println(failure);
+
+        server.waitForNewRegistrationOf(client);
+
+        // Check client is well registered
+        assertThat(client).isRegisteredAt(server);
+        registration = server.getRegistrationFor(client);
+
+        // check we can send request to client.
+        response = server.send(registration, new ReadRequest(3, 0, 1), 500);
+        assertThat(response.isSuccess()).isTrue();
+    }
 }
--- a/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/oscore/OscoreTest.java
+++ b/leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/oscore/OscoreTest.java
@@ -236,7 +236,7 @@ public class OscoreTest {
 //        Failure failure = client.waitForRegistrationFailureTo(server);
 //        System.out.println(failure);
 
-        server.waitForNewRegistrationOf(client);
+        server.waitForReRegistrationOf(registration);
 
         // Check client is well registered
         assertThat(client).isRegisteredAt(server);
--- a/leshan-tl-cf-server-coap-oscore/src/main/java/org/eclipse/leshan/transport/californium/server/OscoreContextCleaner.java
+++ b/leshan-tl-cf-server-coap-oscore/src/main/java/org/eclipse/leshan/transport/californium/server/OscoreContextCleaner.java
@@ -58,7 +58,7 @@ public class OscoreContextCleaner implements RegistrationListener, SecurityStore
     @Override
     public void unregistered(Registration registration, Collection<Observation> observations, boolean expired,
             Registration newReg) {
-        if (registration.getClientTransportData().getIdentity() instanceof OscoreIdentity) {
+        if (registration.getClientTransportData().getIdentity() instanceof OscoreIdentity && newReg == null) {
             removeContext(((OscoreIdentity) registration.getClientTransportData().getIdentity()).getRecipientId());
         }
     }
@@ -67,11 +67,7 @@ public class OscoreContextCleaner implements RegistrationListener, SecurityStore
     public void securityInfoRemoved(boolean infosAreCompromised, SecurityInfo... infos) {
         for (SecurityInfo securityInfo : infos) {
             if (securityInfo.useOSCORE()) {
-                if (infosAreCompromised) {
-                    removeContextCompromised(securityInfo.getOscoreSetting().getRecipientId());
-                } else {
-                    removeContext(securityInfo.getOscoreSetting().getRecipientId());
-                }
+                removeContext(securityInfo.getOscoreSetting().getRecipientId());
             }
         }
     }
@@ -79,13 +75,6 @@ public class OscoreContextCleaner implements RegistrationListener, SecurityStore
     private void removeContext(byte[] rid) {
         OSCoreCtx context = oscoreCtxDB.getContext(rid);
         if (context != null)
-            if (!oscoreCtxDB.getContext(rid).getContextRederivationEnabled()) {
-                oscoreCtxDB.removeContext(context);
-            }
-    }
-
-    private void removeContextCompromised(byte[] rid) {
-        OSCoreCtx context = oscoreCtxDB.getContext(rid);
-        oscoreCtxDB.removeContext(context);
+            oscoreCtxDB.removeContext(context);
     }
 }
--- a/leshan-tl-cf-server-coap-oscore/src/main/java/org/eclipse/leshan/transport/californium/server/OscoreContextCleaner.java
+++ b/leshan-tl-cf-server-coap-oscore/src/main/java/org/eclipse/leshan/transport/californium/server/OscoreContextCleaner.java
@@ -67,7 +67,11 @@ public class OscoreContextCleaner implements RegistrationListener, SecurityStore
     public void securityInfoRemoved(boolean infosAreCompromised, SecurityInfo... infos) {
         for (SecurityInfo securityInfo : infos) {
             if (securityInfo.useOSCORE()) {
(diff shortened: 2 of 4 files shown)


rikard-sics:
As we discussed in issue #1257 I have now created a PR to enable the client using the OSCORE Appendix B.2 functionality.

I did discover that the JUnit test you prepared _registered_device_with_oscore_to_server_with_oscore_then_stop_device_then_register_again_ is unfortunately failing. However, looking at the textual output it seems the client successfully registers twice as expected.

rikard-sics:
Alright, I worked a bit more and found an approach that worked, also adding code to exclude OSCORE Security Contexts from removal in some situations (proposed by Jaroslaw in #1257).

Have a look at the last commit here:
https://github.com/eclipse-leshan/leshan/pull/1694/commits/7e533a7c548c5a6818079ff9583fb52cf44af810

rikard-sics:
Just to highlight an alternative. It's possible to keep using _waitForNewRegistrationOf()_ instead of _waitForReRegistrationOf()_ in the JUnit test if the following change is done in _LeshanTestServer_:
```
    public void waitForNewRegistrationOf(String clientEndpoint, int timeout, TimeUnit unit) {
        registrationEventInOrder.verify(registrationListener, timeout(unit.toMillis(timeout)).times(1)).registered(//
                assertArg(reg -> assertThat(reg.getEndpoint()).isEqualTo(clientEndpoint)), //
                any(), //
                any());
        registrationEventInOrder.verifyNoMoreInteractions();
    }
```
(Changing the isNull() to any() )

**What we want**

What I am thinking is the following:
- We want no changes to waitForNewRegistrationOf
- We want no changes to the context deletion logic. That is, a context should be deleted in _OscoreContextCleaner_ by _removeContext()_ regardless if the context is set to use Appendix B.2.
- We want it such that if the client is killed without de-registering and then attempting to register with the server again, upon reception of the first request the server should derive a new OSCORE Security context based on the content of the securityInfo for that client. Note that this first request is a request intended to start Appendix B.2 by the client. That execution of Appendix B.2 should continue and succeed. I think this bullet would be the key solution.

Does that align with your view?

rikard-sics:
Okay I pushed an improved version with only a small change in _OscoreContextCleaner_. The OSCORE Security Context cannot be removed when de-registration happens due to a new registration. If that happens the OSCORE context just derived from running OSCORE Appendix B.2 will be removed.

I am not sure why I can't use _waitForNewRegistrationOf()_ in the JUnit test but must rely on _waitForReRegistrationOf()_. I would think the de-registration happens even though the OSCORE context is not deleted, but I suppose it is related to that.

sbernard31:
About, **waitForNewRegistrationOf()** instead of **waitForReRegistrationOf()** :

waitForNewRegistrationOf : is for waiting new registration in case we haven't any registration for given client
waitForReRegistrationOf : is for waiting new registration in case we already have a registration for given client

In my test, I do a `stop` without de-registration (so we don't remove the context at server side)
So a registration already exist and so I should have used `waitForReRegistrationOf`

Sorry about that :pray:

leshan-tl-cf-shared-oscore/src/main/java/org/eclipse/leshan/transport/californium/oscore/cf/InMemoryOscoreContextDB.java
```diff
@@ -111,4 +118,24 @@ private static OSCoreCtx deriveContext(OscoreParameters oscoreParameters) {
...
+            return osCoreCtx;
+        } catch (OSException e) {
+            LOG.error("Unable to derive context from {}", oscoreParameters, e);
+            return null;
+        }
+    }
+
+    public void setClientRole(boolean clientRole) {
+        this.clientRole = clientRole;
+    }
```
sbernard31, line 140: Using a setter means that we can change the mode after `InMemoryOscoreContextDB` is created.

I'm not sure we want to be able to do that, so maybe better to add a constructor argument instead of a setter ?
If you want you can add a second constructor and use a default clientRole = false for the already existing one.

rikard-sics:
> About, **waitForNewRegistrationOf()** instead of **waitForReRegistrationOf()** :
> 
> waitForNewRegistrationOf : is for waiting new registration in case we haven't any registration for given client waitForReRegistrationOf : is for waiting new registration in case we already have a registration for given client
> 
> In my test, I do a `cllient.stop` without de-registration (so we don't remove the context at server side) So a registration already exist and so I should have used `waitForReRegistrationOf`
> 
> Sorry about that 🙏

No, worries, and thanks the clarification. Good, then we can stick to using **waitForReRegistrationOf()**

leshan-tl-cf-shared-oscore/src/main/java/org/eclipse/leshan/transport/californium/oscore/cf/InMemoryOscoreContextDB.java
rikard-sics: Yes, I see, that can make sense. I did it like that now in commit f480653b1f302be6ffbeab1bf4f27a5d0b026130

sbernard31:
>  We want no changes to waitForNewRegistrationOf() and preferably to use it over waitForReRegistrationOf()

Yep, (explanation just above at https://github.com/eclipse-leshan/leshan/pull/1694#issuecomment-2653184036)

>  We want no changes to the context deletion logic. That is, a context should be deleted in OscoreContextCleaner by removeContext() regardless if the context is set to use Appendix B.2.

For the scope of this PR, probably we still want to remove context on deregistration.
But I'm not sure if this is really want we want at long term. :shrug: 
The reason I do that is because I didn't find any other way to do OSCORE context housekeeping for now.
(see https://github.com/eclipse-leshan/leshan/issues/1230)

> We want it such that if the client de-registers and then attemps to register with the server again, upon reception of the first request the server should derive a new OSCORE Security context based on the content of the securityInfo for that client. Note that this first request is a request intended to start Appendix B.2 by the client. That execution of Appendix B.2 should continue and succeed.

>  If the client is killed (without de-registering) and then attempts to register with the server again, upon reception of the first request the server should retrieve the still available OSCORE Security context for that client.This first request is a request intended to start Appendix B.2 by the client. That execution of Appendix B.2 should continue and succeed.

For this PR probably we want to do that.
At long term, I don't know too. That questions are probably related to the point just above.
In other words : "Should `OscoreContext` lifetime be strictly tied to Registration lifetime ?"
(better to discuss at : https://github.com/eclipse-leshan/leshan/issues/1230)

leshan-tl-cf-server-coap-oscore/src/main/java/org/eclipse/leshan/transport/californium/server/OscoreContextCleaner.java
```diff
@@ -58,7 +58,7 @@ public void updated(RegistrationUpdate update, Registration updatedReg, Registra
     @Override
     public void unregistered(Registration registration, Collection<Observation> observations, boolean expired,
             Registration newReg) {
-        if (registration.getClientTransportData().getIdentity() instanceof OscoreIdentity) {
+        if (registration.getClientTransportData().getIdentity() instanceof OscoreIdentity && newReg == null) {
             removeContext(((OscoreIdentity) registration.getClientTransportData().getIdentity()).getRecipientId());
```
sbernard31, line 62: I see why it doesn't work without that.
but I'm not so sure about the solution.

I think there is case where we want to remove the context :thinking: 
Maybe we should not remove it only old registration and new registration use same `OscoreIdentity` ? 

(note that first argument of unregistered method is the old registration)

rikard-sics: Nice!

Not sure what your preferences are for nesting or verbosity, but what about something like:
```
    @Override
    public void unregistered(Registration registration, Collection<Observation> observations, boolean expired,
            Registration newReg) {
        if (registration.getClientTransportData().getIdentity() instanceof OscoreIdentity) {
            boolean matchingIdentity = registration.getClientTransportData().getIdentity()
                    .equals(newReg.getClientTransportData().getIdentity());
            if (!matchingIdentity) {
                removeContext(((OscoreIdentity) registration.getClientTransportData().getIdentity()).getRecipientId());
            }
        }
    }
```

sbernard31: I think there is potential NPE issue here. 

Generally I privilege readability. (but this is very subjective :sweat_smile:)
Also, i feel exist fast help for reading.


So maybe something like that ?
``` java
LwM2mIdentity unregisteredIdentity = registration.getClientTransportData().getIdentity();

// Not an OSCORE identity : nothing to clear
if (!(unregisteredIdentity instanceof OscoreIdentity))
    return;

// This is a Re-Registration and client use same OSCORE Identity : nothing to clear
if (newReg != null && unregisteredIdentity.equals(newReg.getClientTransportData().getIdentity())) {
    return;
}

// else
removeContext(((OscoreIdentity) unregisteredIdentity).getRecipientId());
```

rikard-sics: > I think there is potential NPE issue here. (if negReg is null)

Yeah, I noticed and updated the comment, but probably it updated after you saw it:
```
boolean matchingIdentity = newReg != null && registration.getClientTransportData().getIdentity()
                .equals(newReg.getClientTransportData().getIdentity());
```

> Generally I privilege readability. (but this is very subjective 😅) Also, i feel exist fast help for reading.
> 
> So maybe something like that ?

Looks good! I have updated the PR aligned with that now.

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/oscore/OscoreTest.java
```diff
@@ -197,4 +197,53 @@ public void registered_device_with_oscore_to_server_with_oscore_then_removed_sec
...
+
+        // Stop client without sending deregister request
+        client.stop(false);
+
+        // Start it and wait for registration
+        client.start();
+        // TODO to remove
+        // Uncomment to see the failure
+//        Failure failure = client.waitForRegistrationFailureTo(server);
+//        System.out.println(failure);
```
sbernard31, line 237: I think we can remove this comment

leshan-tl-cf-shared-oscore/src/main/java/org/eclipse/leshan/transport/californium/oscore/cf/InMemoryOscoreContextDB.java
```diff
@@ -35,25 +36,32 @@ public class InMemoryOscoreContextDB extends HashMapCtxDB {
 
     private final OscoreStore store;
 
+    public boolean clientRole = false;
```
sbernard31, line 39: this attribute should probably be final.

sbernard31 (changes requested):

leshan-tl-cf-shared-oscore/src/main/java/org/eclipse/leshan/transport/californium/oscore/cf/InMemoryOscoreContextDB.java
rikard-sics: Yep. Do you prefer to keep it public or shall I make it private?

leshan-integration-tests/src/test/java/org/eclipse/leshan/integration/tests/oscore/OscoreTest.java
rikard-sics: Done in commit 28278a507dae8ec0cc3a5e6a57800e6a01fe8632

leshan-tl-cf-shared-oscore/src/main/java/org/eclipse/leshan/transport/californium/oscore/cf/InMemoryOscoreContextDB.java
sbernard31: good catch it should be private :+1:

rikard-sics: Alright! Now done in commit 3aeba3189b4faddd4a741ef044b1422d2040c2be

sbernard31 (approved):
:+1:

PR #1708, opened 2025-04-02
Added support for device certificate chain in leshan-lwm2m-client module

jakubsobolewskisag:
## PR Description

LwM2m 1.2 specification allows devices to send/use X.509 certificate chains in the Security object and therefore in communication with LwM2m servers (and bootstrap servers). To prepare for this, Leshan client library can already use certificate chains in a backwards-compatible way. This change allows client implementations to use certificate chains.


sbernard31:
Thx for the PR.
As this is a bit out of scope, I created an issue to discuss about that : https://github.com/eclipse-leshan/leshan/issues/1710

sbernard31:
I pushed some more code to limit usage of DER encoding by default.
The code should allow to provide custom `ServersInfoExtractor` for Client and `ConfigurationChecker` for Bootstrap Server.

I will probably add some more commit to clean old code following sonar advice (I prefer to do that separately to keep commit readable).

Maybe out of scope of that PR : 
 - Maybe it would make sense to create a utility class for Tests to handle certificates/files in src/test/resouces 
 - `ServersInfoExtractor` it is now dynamic but I need to double check if the way we access to it is OK :thinking: 
 - I also notice that there is a problem in objectTree with objectEnabler initialization using `LwM2mObjectTree.addObjectEnabler(LwM2mObjectEnabler)`. It seems that `enabler.init(client, linkFormatHelper, serversInfoExtractor);` is not called and I think it should (I need to double check that)
 - some constructor have too much arguments we should find a way to make this cleaner.
 
 I will be out of office next week back Tuesday 21th.

leshan-lwm2m-bsserver/src/main/java/org/eclipse/leshan/bsserver/ConfigurationChecker.java
```diff
@@ -16,33 +16,11 @@
...
-import org.eclipse.leshan.core.util.StringUtils;
-
 /**
  * Check a BootstrapConfig is correct. This is a complex process, we need to check if the different objects are in
  * coherence with each other.
  */
-public class ConfigurationChecker {
-
-    private OscoreValidator oscoreValidator = new OscoreValidator();
+public interface ConfigurationChecker {
```
sbernard31, line 23: Just to let you know I change the `ConfigurationChecker` class as an interface and add `DefaultConfigurationChecker` class.

leshan-lwm2m-bsserver/src/main/java/org/eclipse/leshan/bsserver/LeshanBootstrapServerBuilder.java
```diff
@@ -313,6 +315,16 @@ public LeshanBootstrapServerBuilder setEndpointsProviders(
...
+    /**
+     * Sets custom {@link ConfigurationChecker} implementation.
+     *
+     * @param configurationChecker custom configuration checker
+     */
+    public LeshanBootstrapServerBuilder setConfigurationChecker(ConfigurationChecker configurationChecker) {
+        this.configurationChecker = configurationChecker;
+        return this;
+    }
+
```
sbernard31, line 327: I ask myself if we should add that ? 

I mean as the checker is only used in constructor of some `BootstrapConfigStore` implementation.

Maybe being able to do : 
```
builder.setConfigStore(new InMemoryBootstrapConfigStore(myConfiguraitonChecker));
 ```
 Is enough ? 
 
WIth current way, I have some concern if users try to do something like : 
```
builder.setConfigurationChecker(myConfiguraitonChecker));
builder.setConfigStore(new JSONFileBootstrapStore());
```

sbernard31:
I did : 
> -  Maybe it would make sense to create a utility class for Tests to handle certificates/files in src/test/resouces
  
 For that ones, I will do that in another PR : 
> -  ServersInfoExtractor it is now dynamic but I need to double check if the way we access to it is OK 🤔
> -  I also notice that there is a problem in objectTree with objectEnabler initialization using LwM2mObjectTree.addObjectEnabler(LwM2mObjectEnabler). It seems that enabler.init(client, linkFormatHelper, serversInfoExtractor); is not called and I think it should (I need to double check that)
> -  some constructor have too much arguments we should find a way to make this cleaner.

@jakubsobolewskisag, if you want you can review this PR and test if that solves your issue.

sbernard31:
(ignore the legal agreements issue, I suspect that eclipse infra is partially down)

sbernard31:
@jakubsobolewskisag finally I also did : 
> -  ServersInfoExtractor it is now dynamic but I need to double check if the way we access to it is OK 🤔
> - I also notice that there is a problem in objectTree with objectEnabler initialization using LwM2mObjectTree.addObjectEnabler(LwM2mObjectEnabler). It seems that enabler.init(client, linkFormatHelper, serversInfoExtractor); is not called and I think it should (I need to double check that)

in this PR. 

I don't touch anything more until I get you feeback :slightly_smiling_face:

leshan-lwm2m-bsserver/src/main/java/org/eclipse/leshan/bsserver/LeshanBootstrapServerBuilder.java
jakubsobolewskisag: True - the way you're proposing is much better.

jakubsobolewskisag:
@sbernard31 Thank you! I will check that next week, since I'm on PTO now ;)

leshan-lwm2m-bsserver/src/main/java/org/eclipse/leshan/bsserver/LeshanBootstrapServerBuilder.java
sbernard31: Ok so I will do that change :)

sbernard31:
(Please note that I will be unavailable for most of the month, so replies may be delayed 😅 )

sbernard31:
@jakubsobolewskisag do you still plan to review / test if this PR fit your needs ?

sbernard31:
@jakubsobolewskisag let me know if you still want to review/test it.

(So, I know if I need to wait for it :slightly_smiling_face:, no problem if you need more time but just let me know)

sbernard31:
@jakubsobolewskisag, just to let you know without any news from you, I plan to integrate this maybe next week.

Of course if you ask for more time, it's ok for me.

(As I haven't heard from you in a while, I hope everything is going well for you. :crossed_fingers:  )

jakubsobolewskisag:
@sbernard31 thank you! As I mentioned in the SNI topic, I was overloaded and got back to these topics only last week... Againg sorry for not responding earlier. Thank you again for merging this feature :)

PR #1755, opened 2026-05-21
(minor)(sonar) refactor(redis-server): improve code maintainability and security

boly38:
## PR Diff

- Extract validation logic in RedisRegistrationStore.Builder: (sonar: cognitive complexity)
  - Introduce validatePrefixesNotEmpty() for prefix validation
  - Introduce validatePrefixesUniqueness() for prefix uniqueness check
  - Introduce applyGlobalPrefix() for prefix application
  - Improve code readability and reusability
- Enhance security in SingleInstanceJedisLock: (sonar: SecureRandom)
  - Replace Random with SecureRandom for cryptographically secure randomness
  - Update generateLockValue() signature to use SecureRandom
  - Improve lock value generation security

## Context - Issue

Hi, 👋 

I had to work with leshan, and override leshan redis glue to work with different redis version. So I noticed some small sonar  improvements : secureRandom, cognitive complexity

- If you think this usefull, feel free to manage directly this PR/branch content without forcibely waiting for me,
- If you think this useless, feel free to reject and close this PR.

As this is minor self-describing PR, there is no dedicated issue.

Regards

### DIFF
--- a/leshan-lwm2m-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisRegistrationStore.java
+++ b/leshan-lwm2m-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisRegistrationStore.java
@@ -31,6 +31,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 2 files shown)


sbernard31:
Hi :wave:,

Thx for this contribution. :slightly_smiling_face: 

About,
> Extract validation logic in RedisRegistrationStore.Builder

I think changes are a good idea.
This will makes the code more readable and remove some duplicates.
(I would do some minor naming changes, see review for more details)

About,
> Enhance security in SingleInstanceJedisLock: (sonar: SecureRandom) 

I'm not so sure about it. `SecureRandom` is mainly for cryptography/security concern.
Here we don't use it for that kind of use cases... and as SecureRandom is a bit slower not sure this is a right move.

By curiosity,  I asked about that to "claude IA" (without sharing my first idea :sweat_smile: ), and he agree with that.
But he advices to use `ThreadLocalRandom` (`ThreadLocalRandom.current().nextBytes(...)`) instead to limit bottleneck... 

Javadoc of `Random` seems to confirm that : 
```java
 /**
 * ....
 * Instances of {@code java.util.Random} are threadsafe.
  * However, the concurrent use of the same {@code java.util.Random}
  * instance across threads may encounter contention and consequent
  * poor performance. Consider instead using
  * {@link java.util.concurrent.ThreadLocalRandom} in multithreaded
  * designs.
  * ....
 */
 ```
 
By the way [our sonar instance](https://sonarcloud.io/code?id=eclipse-leshan_leshan&selected=eclipse-leshan_leshan%3Aleshan-lwm2m-server-redis%2Fsrc%2Fmain%2Fjava%2Forg%2Feclipse%2Fleshan%2Fserver%2Fredis%2FSingleInstanceJedisLock.java) does even raise an issue about that.. :thinking:

**Comment by sbernard31** [comment]
> I had to work with leshan, and override leshan redis glue to work with different redis version.

Do you plan to contribute that ? 
Or to propose code to make it easier ?

File: leshan-lwm2m-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisRegistrationStore.java [hunk 1]
```diff
@@ -1049,39 +1049,32 @@ protected Builder generateDefaultValue() {
...
+            validatePrefix("endpointByIdentityPrefix", this.endpointByIdentityPrefix);
+            validatePrefix("endpointLockPrefix", this.endpointLockPrefix);
+            validatePrefix("observationByIdPrefix", this.observationByIdPrefix);
+            validatePrefix("observationIdsByRegistrationIdPrefix", this.observationIdsByRegistrationIdPrefix);
+            validatePrefix("endpointExpirationKey", this.endpointExpirationKey);
+        }
 
-            if (this.endpointExpirationKey == null || this.endpointExpirationKey.isEmpty()) {
-                throw new IllegalArgumentException("endpointExpirationKey should not be empty");
+        private void validatePrefix(String prefixName, String prefixValue) {
```
sbernard31: probably more consistent to name this method : `validatePrefixNotEmpty` instead of  `validatePrefix`

(Especially as caller is named `validatePrefixesNotEmpty`)

leshan-lwm2m-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisRegistrationStore.java
```diff
@@ -1049,39 +1049,32 @@ protected Builder generateDefaultValue() {
...
+            if (prefixValue == null || prefixValue.isEmpty()) {
```
sbernard31, line 1074: Should we also check that prefixName is not null ?

boly38:
> ThreadLocalRandom.current().nextBytes(...)

Hi Sbernard31 
- thanks, I'll remove SecureRandom to switch to `ThreadLocalRandom.current().nextBytes(...)` 👍 

(NB: reverting to original code version, my sonarlint sync with my sonar didn't re-throw issue on that, strange.. anyway switch to TLR)

leshan-lwm2m-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisRegistrationStore.java
boly38: U right ✔️ 👍

leshan-lwm2m-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisRegistrationStore.java
boly38: Good point, but since prefixName is only used internally (always called with hardcoded string literals from validatePrefixesNotEmpty()), it can never be null in practice. Also, `@NotNull` annotations are not used in this project, so adding a null check here would be inconsistent. I think leaving it as-is is fine.

boly38:
> work with different redis version

in fact the change is very light : retrieve 3 classes only.
- SingleInstanceJedisLock -> SetParams chaining replaced by separate calls
- JedisLock -> No functional change
- RedisRegistrationStore -> No functional change (only wires the lock impl)

 I didnt manage to fully enter in leshan server/project internal logic to do this (for eg. to auto switch/select between implem 4/5.x + 6.x) but I could provide some info a below:

<details><summary>Jedis 6.x compatibility fix — `SetParams` chaining break</summary>


### Problem

In Jedis 6.x, `SetParams.px(long)` return type changed from `SetParams` to `BaseSetExParams`.  
This breaks the fluent chaining pattern used in `SingleInstanceJedisLock`:

patch

```java
// Jedis 5.x — works
// Jedis 6.x — compile error: nx() not found on BaseSetExParams
-while (!"OK".equals(j.set(lockKey, randomLockValue, SetParams.setParams().nx().px(expiration)))) {
+// Jedis 6.x: SetParams.px(long) returns BaseSetExParams (not SetParams) — cannot chain fluently.
+ // Call nx() and px() separately: both mutate the instance in-place.
+SetParams setParams = SetParams.setParams();
+setParams.nx();
+setParams.px(expiration);
+
+while (!"OK".equals(j.set(lockKey, randomLockValue, setParams))) {
```

</details>

leshan-lwm2m-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisRegistrationStore.java
sbernard31: The idea was to detect quickly something wrong in Leshan code.
With actuel code if leshan dev call `validatePrefix(null,myPrefix).
This will lead to a visible error only if a user set  myPrefix to null or empty and so this will raise an NPE.

But I understand your point.

sbernard31:
Thx for modification. 

I should be able to integrate that in `master` on Wednesday (my only day of work next week, then in June I should be more available)

leshan-lwm2m-server-redis/src/main/java/org/eclipse/leshan/server/redis/RedisRegistrationStore.java
boly38: amend with
>Objects.requireNonNull(prefixName, "prefixName must not be null");

👍 was a great pleasure to exchange with you 🙏

sbernard31:
@boly38  this is now integrated in `master`. 
Thx for your contribution :pray: 

You probably know that but just in case, if you want to execute Leshan Junit tests for Redis, you should use `redis` profile like this :  `mvn clean install -Predis` (The default redis server URL is localhost:6379 but you can change it using REDIS_URI env var)

Also, your PR raise an interesting question : [Which version of Jedis should we use for Leshan v2.0.0 ?](https://github.com/eclipse-leshan/leshan/issues/1757)  
Please, do not hesitate to share your thought on that topic :pray:    
(you OR any people at Orange involved as Leshan user)

